PR #6, opened 2016-12-29
Adding Default Language switch to generate config script.

Bitti09:
As an answer to this request: https://forum.mailcow.email/t/mailcow-german-ui/274

---
This change is [Reviewable](https://reviewable.io/reviews/andryyy/mailcow-dockerized/6)


Bitti09:
rebased

andryyy:
I'm not sure about this change. I think about moving the htdocs into the container, we can do it with an env var then. But this env var should also set SOGo default language.

PR #7, opened 2016-12-31
Create LICENSE

PhoenixPeca:
Hi André,

I've added the same license to the mailcow dockerized version as to the non-deckerized version.

-----------
By the way, I apologize for I was not able to create new pushes lately, I've been very busy this holiday season.  However, I have managed to find a little time last week, so I created a pull request to the roundcube website's source code that [links to the mailcow project because of using roundcube](http://roundcube.net/download).
Roundcube
I'm sorry for the lousy description, but I guess that should do its job.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,339 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 2, June 1991
+
+ Copyright (C) 1989, 1991 Free Software Foundation, Inc., <http://fsf.org/>
+ 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ Everyone is permitted to copy and distribute verbatim copies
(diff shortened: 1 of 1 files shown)


andryyy:
Not sure about the license. It should be mentioned that the license of each software needs to be checked. 

I will remove most hardware infos from the panel too. And also the public folders etc. I don't want to use shell exec etc. anymore. 

Can you by the way change the add domain buttons to links in mailcow? :-) these in the tables. 

No problem about being busy. :-)

PhoenixPeca:
Hi,

I see..... Working with shell exec makes implementation of the dockerized version difficult, so yeah No problem!!! ;-) I'll change the add domain buttons to links...

Happy new year to you!!!

Braintelligence:
So... what's the license of mailcow-dockerized, then? 😛

PR #11, opened 2017-01-04
Nginx ciphers + http(s) redirect

williamkray:
forces redirect from :80 to :443 (typically expected behavior)

also adds more thorough list of SSL ciphers to nginx config. i was previously unable to sync cal/carddav data with davdroid, this resolved. ciphers were pulled from a recently initiated install of iRedMail.


andryyy (approved):

andryyy (changes requested):
Can you remove the nginx port 80 stuff? I will add a dynamic port config soon. :-) But the ciphers are very useful!

andryyy:
I added it right now. :)

PR #21, opened 2017-01-15
postfix-logwatch and mail.log in admin.php

tehXor:
Just as one suggestion to add a postfix-logwatch and mail.log view to admin.php

What could be improved:
- Add postfix-logwatch.conf to the `conf` directory and use it from there
- Maybe there would be a better way to grant `www-data` from the `php-fpm` container read access on the mounted log files than adding it to the `adm` group

Notes:
- I used postfix-logwatch over pflogsumm as it seemed to me the latter isn't counting postscreen entries
- Having `mailq` would be nice as well but I haven't found a lightweight way to do it yet, which wouldn't involve exposing the docker socket or running an SSH daemon in the postfix container


snc:
Adding a cron service or jobber (http://dshearer.github.io/jobber/) to the postfix image could be a solution. The job could write the output of `mailq` to Memcached or Redis or a shared volume.

broedli:
While your approach certainly works, I'm not in favor of your use of `shell_exec`.
I think an implementation like Rspamd has with it's history and redis  would be a much better idea.

tehXor:
This pretty much is a very rudimentary port of the `sys_info()` function of the original mailcow, which uses `file_get_contents` (https://github.com/mailcow/mailcow/blob/master/webserver/htdocs/mail/inc/functions.inc.php#L359) to bring back some of its features asap. Unfortunately `mailq` is missing though.

A more sophisticated approach than `file_get_contents` or `shell_exec` with `tail` clearly would be better also I'm not sure if it's worth it considering that probably only very few people per installation (Admins only) will use it. Performance and resource efficiency shouldn't really matter.

Plus personally I have become a fan of `postfix-logwatch` and I'm not sure if you would gain much or anything at all if you don't print its output directly (besides some resource savings).

broedli:
@andryyy had a little fun last night and implemented a log pipe from dovecot and postfix to redis with syslog-ng and implemented a rough example table into the mailcow UI (refer: 8c8bfc0, d614aaf, ecda4fb).  So this feature could be implemented. But how should it look? Should we pursue the approach taken in current `mailbox.php` with tabs at the top to avoid 'scroll fests' and combine the 'configuration' / and 'mailbox' section? How should such a dashboard look? Only logs? 

Here are some visual references:

logtable

tabs

Braintelligence:
@broedli Awesome! I think it looks good :)

tehXor:
I'm closing this as the new implementation of "Logs" is very neat 👍 

I want to preserve the postfix-logwatch report for me somehow but with the current Mailcow version I ran it manually from time to time and would create a new PR if I find the time to attach it to the web GUI in a fitting way.

PR #32, opened 2017-01-30
Spanish language

nthings:
This is the spanish translation


andryyy:
Wow, that was fast! Thanks, I will add it in a few minutes. :-)

PR #37, opened 2017-01-31
admin user is not being created

sfinke0:
Moin,

I just tried a fresh setup and it did not create the admin user therefore I was not able to login.

```sql
MariaDB [mailcow]> select * from admin;
Empty set (0.01 sec)
```

Please have a look, this might fix the problem.

cheers,
carazzim0


mis4s:
I confirm. Yesterday I installed and could not log in (admin/moohoo). Today I wanted to report a bug and now we have a pull request.

Thanks!

andryyy:
Thanks, did not test this... sorry.

PR #49, opened 2017-02-14
Update autodiscover.php

PhoenixPeca:
I guess we should leave the 'inc/vars.local.inc.php' below the config array so that the user can override the values of the array without loosing modifications everytime he/she updates.

However, I think things were different in the dockerized version, so maybe, ignore this PR if needed.


andryyy:
Oh, did not see this, sorry. But thanks!

PR #57, opened 2017-02-18
Added temporary alias count

PhoenixPeca:
Hi,

This feature should tell the domain admin how many **temporary aliases** did a mailbox user created. <strike>Since mailcow now supports dual-login, I don't think there's a need to limit the creation of temporary aliases.</strike>
capture


PhoenixPeca:
Discovered an issue. Im going to temporarily hold this

andryyy:
Hi, sorry for my late reply!
I cleaned the code and moved everything that queries the db to the functions (finally).

PhoenixPeca:
Hi Andre, don't be sorry. I've seen your pushes in the dev branch, and I think it's better 😄 . It's great to have this feature. hahah

PR #67, opened 2017-02-25
Fixed some PHP Errors and PHP Warnings

PhoenixPeca:
Hi @andryyy,
We should stop the PHP code execution if we can't connect to the database, otherwise we'll be greeted with PHP warning and fatal erros.
capture

See the fixed version below:
capture2

I have tested this in a 512 MB RAM machine, which is enough to render the database server out of memory and unable to start.


andryyy:
Can you push it to dev, too?

PR #70, opened 2017-02-26
HTML Improvement

PhoenixPeca:
Changelog:
---------------------------
* **Set the colspan to 999** - Setting a low colspan value could affect the layout of the pages in case new column is added.

> untitled
The red-bordered area has been excluded from the span of the "Add mailbox" column. Setting it to 999 will fix this and will prevent this from happening again.
-------------------------
* **Added a message that tells the user to add a domain (if there isn't any domain) in all 4 panels**
untitled
Previously, there isn't any message that tells the admin that further domain configuration is required to add a new mailbox/alias/etc
------------------------------
* **Added a new language entry in English** - Needs translation to other languages.
-------------------------------
* HTML `max-width` of a column is too small that it destroys the design of the page.
> capture
I've widened the max-width so that it fits perfectly.
------------------------------
_I'm sorry for the code indention. My editor doesn't always cooperate. hahahha_

### DIFF
--- a/data/web/lang/lang.en.php
+++ b/data/web/lang/lang.en.php
@@ -246,6 +246,7 @@ $lang['mailbox']['add_domain_alias'] = 'Add domain alias';
 $lang['mailbox']['add_mailbox'] = 'Add mailbox';
 $lang['mailbox']['add_resource'] = 'Add resource';
 $lang['mailbox']['add_alias'] = 'Add alias';
+$lang['mailbox']['add_domain_record_first'] = 'Please add a domain first';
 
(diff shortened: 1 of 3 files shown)


andryyy:
Thanks. :-) Is it fine setting colspan to 999? Shouldn't we fit it?

PhoenixPeca:
I'm sorry to tell you that I honestly don't know the consequences 😢, I just did it because it fixes the problem (and will theoretically fix the same problem in the future in case new columns are added) in modern browsers (i.e: Google Chrome).

PR #72, opened 2017-02-26
Incorrect mysql statement

PhoenixPeca:
Hi Andre,
I have corrected your MySQL query statement so that we can retrieve the proper spamalias count from the database. 😄 Leaving it as it is will result in a count 0 spamaliases no matter how many spamaliases a mailbox user makes.

I have tested this alreadly..


andryyy:
Oh my... what have I done. :-D

PR #99, opened 2017-03-06
Added docker-compose installation

fabreg:
Please mention the installation of docker-compose


andryyy:
Thanks, but this must be updated each time it is updated.

PR #126, opened 2017-03-17
GUI improvements: collapsible tables & password manager compatible login form

tehXor:
Two little improvements which could be generally interesting:

- Changed the type of the user field on the login form from "name" to "text" as afaik "name" is no valid HTML and some password managers do not recognize it correctly.
- Added some JavaScript for the mailbox view which makes the tables collapsible, which is very handy when they get very long. They also are collapsed per default when there is at least one table which is higher than 550 pixel (which means in total the page height already will be very big).


andryyy:
Thanks!
I'm about to move to FooTable and load data via JSON api, we will have pagination and a better filter then. :-)

PR #161, opened 2017-03-25
Prevent accidental text-highlighting

PhoenixPeca:
Double-clicking on the column header of the table often induces text-highlighting. Which I think is not good looking.. But this PR fixes it. 😉 

untitled


andryyy:
Thanks, but please push to dev. :-)

PhoenixPeca:
.okies, no problem! 😄 I completely forgot about it.  I'll do it next time. Sorry for that

PR #162, opened 2017-03-25
HTML Improve

PhoenixPeca:
As stated in this thread: [Engineering - StackExchange - html](http://softwareengineering.stackexchange.com/questions/224422/will-it-be-a-wrong-idea-to-have-style-in-body)

> A style element inside the body element violates HTML syntax rules....

Let's try to comply to these rules as much as possible. 😉


data/web/inc/header.inc.php
```diff
@@ -23,6 +23,29 @@
 <?=(preg_match("/mailbox.php/i", $_SERVER['REQUEST_URI'])) ? '<link rel="stylesheet" href="/css/mailbox.css">' : null;?>
```
andryyy: Why would you add belows changes when this line already includes a file "mailbox.css"?

PhoenixPeca: Oh, sorry. My bad.

PhoenixPeca: Fixed it. 😄

PR #163, opened 2017-03-25
Show "No Record" if Domain admin table is empty

PhoenixPeca:
Added "No record" message when there is no data to show.
- I have also removed an unused code.

untitled

### DIFF
--- a/data/web/admin.php
+++ b/data/web/admin.php
@@ -93,9 +93,10 @@ $tfa_data = get_tfa();
               </thead>
               <tbody>
                 <?php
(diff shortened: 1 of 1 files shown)


andryyy:
It already does.

andryyy:
No

andryyy:
Sorry, thought it was another table. :-)

I will change this to footable soon, so no need for it. But thanks!

Edit: By the way, footable does use a Json Api I am working on, but this section is not yet in the json_api.php file.

PR #179, opened 2017-04-06
Warn the user about installing mailcow on Debian 8

unixfox:
I don't know if it's the right place to warn the user about running mailcow with Docker on Debian 8.
I do this as a prevention because this happened to me and I've to wait for Debian 9 or reinstall my VPS to Ubuntu 16.04, this is really annoying.


andryyy:
Can you create upcoming PRs in dev? :-) I will change it now and merge it, thanks!

PR #180, opened 2017-04-08
Second read in the if loop break automation as both cases in the loop…

Klazomenai:
… require user input, removing else to allow for automation


andryyy:
You assume the read timezone is always correct. If you want to use automation, you should generate the conf file in the script you are writing.

Some people install virtual machines from templates and don't check the timezone afterwards. To name an example. :)



Beste Grüße 
André

Issue #189, opened 2017-04-12
Add Kubernetes helm support


titilambert:
Hello,
I'm currently trying to adapt mailcow-dockerized as kubernetes helm chart.
I have already a good result !
I would like to know if you have interest on that stuff ?
If yes, how you want I commit that ? (a new folder ?)

Thanks for your answer

kristiandrucker:
@titilambert where can I find the helm charts? I want to deploy mailcow on my k8s cluster. Also can it work behind traefik and with ingress?

titilambert:
Hello! I got something working, but I need to rework it to make it more flexible (without my password).
BTW, in my own cluster, it works with nginx ingress.
But, we need to get https://github.com/andryyy/mailcow-dockerized/pull/190 merged before anything

titilambert:
@kristiandrucker Where do you have your k8s cluster ? Are you using any cloud provider ? AWS, GCE?

kristiandrucker:
@titilambert I have a baremetal install with pvc working.
Could you give me the helm chart so I may look at it and test it on my cluster?

kwiesmueller:
Hey there,
I would be happy to assist with both reviews and the chart itself. Otherwise I would build one in the next few days.

Oh and @titilambert if you need a quick solution for secrets in helm, we built some small management tool for our own clusters with helm (https://github.com/kwiesmueller/captain-template). It is WIP but i already use it a lot for managing all clusters with helm (aside from plain manifests).

TBK:
@kwiesmueller Did you implement your own chart?

I am migrating my setup to Kubernetes, so an already made/tested helm chart for mailcow would save me some time.

kwiesmueller:
@TBK0000 not yet, had no time yet. If you would like, we could try to continue the work that went into this here or take this over together if @titilambert stays inactive on this? Is there a PR maybe, or  have yo got a repo?

TBK:
Sure, I got exams the next couple of weeks after that it is my intention to develop a chart for mailcow.

I have setup a repo for the work https://github.com/JJTC-Helm-Charts/MailCow - PRs are welcome.

kwiesmueller:
@TBK0000 great, will take a look the some day this week

Promaethius:
https://github.com/kubernetes/kompose/blob/master/docs/conversion.md
As far as I can tell, kompose will work with everything except the volume and dns parameters.

[...]

PR #190, opened 2017-04-12
Add DBhost parameter

titilambert:
Modifications for implementing #189


LeoColman:
This PR was closed. Does it work, or is it a planned feature?

andryyy:
SQL is not SQL. People will hook their SQL version 5.5 etc. and complain why stuff fails. 

You should be very sure your Sql is compatible. I have not tested various versions.

LeoColman:
Interesting. The main objective is to lower the resources the server is using to the minimum possible, and running MYSQL with everything else is kinda costy. Do you think this would provide an alternative to this problem?

andryyy:
You can change the sql configuration file to fit your resources. SOGo EAS cache is quiet aggressive.

PR #198, opened 2017-04-16
Allow adding aliases to alias domains

mkuron:
Currently, one always gets "Max aliases exceeded" when attempting to add an alias that redirects alias@alias-domain to user@domain. This is because alias domains are listed in a separate SQL table that doesn't even have a column that specifies the maximum number of aliases.

The [documentation](https://andryyy.github.io/mailcow-dockerized/first_steps/#sender-and-receiver-model) even says that this should work:
> It is important to know, that you are not able to receive mail for my-alias@my-alias-domain.tld. You would need to create this particular alias.

However, what the documentation suggests cannot currently be done due to the error message above.

My pull request simply prevents the number of aliases from being checked on alias domains.

### DIFF
--- a/data/web/inc/functions.inc.php
+++ b/data/web/inc/functions.inc.php
@@ -2678,7 +2678,7 @@ function mailbox_add_alias($postarray) {
 		$address      = $local_part.'@'.$domain;
 
     $domaindata = mailbox_get_domain_details($domain);
-    if ($domaindata['aliases_left'] == 0) {
+    if (is_array($domaindata) && $domaindata['aliases_left'] == 0) {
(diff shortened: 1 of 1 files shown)


Shad86:
i also think that only happen on fresh installations, seems it broke on some commit because i updated my own older version to the latest git and still can add aliases but today i had also a guy with a fresh installation and it just wont work to add e-mail aliases because it says as you said @mkuron  "Max aliases exceeded".

andryyy:
Thanks for reporting it! I changed some other parts, too, so we don't need that PR anymore.

PR #199, opened 2017-04-16
ClamAV virus scanning

### LINKED ISSUE #58: AntiVirus Integration?
Hi, are there any plans for ClamAV integration or some sort of diffrent antivirus?

mkuron:
This pull request integrates ClamAV virus scanning.
Please set up the clamav Dockerfile for an auto-build on Docker Hub after merging.

For those who don't want to have virus scanning on the server-side, simply remove the service from _docker-compose.yml_. If the three seconds for the connection attempt to ClamAV are too long, just delete _data/conf/rspamd/local.d/antivirus.conf_.

Closes #58

### DIFF
--- a/data/Dockerfiles/clamav/Dockerfile
+++ b/data/Dockerfiles/clamav/Dockerfile
@@ -33,9 +33,6 @@ RUN sed -i 's/^Foreground .*$/Foreground true/g' /etc/clamav/clamd.conf && \
     echo "TCPSocket 3310" >> /etc/clamav/clamd.conf && \
     sed -i 's/^Foreground .*$/Foreground true/g' /etc/clamav/freshclam.conf
 
-# volume provision
-VOLUME ["/var/lib/clamav"]
-
 # port provision
 EXPOSE 3310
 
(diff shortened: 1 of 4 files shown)


andryyy:
Hi, thanks, but I already rewrote the scripts and changed the configuration files. It will not be default. Most of the time the download/refresh of the databases fails with this clamd Dockerfile. I defined an array of working and fast mirrors to work this around.

mkuron:
Sounds good. When will you make your version available?

andryyy:
Maybe today or tomorrow.

mkuron:
@andryyy, I saw your commits on https://github.com/andryyy/mailcow-dockerized/commits/dev that implement the same as my pull request. I have two comments:
- You use `max_size = 20000000` in data/conf/rspamd/local.d/antivirus.conf. Viruses in larger messages will not be scanned, making it possible to circumvent the virus scanner. Postfix's size limit is already set to 26214400 bytes, which is only slightly more, so you might as well remove the `max_size` setting from antivirus.conf.
- Could you please add the relevant lines for a ClamAV container to docker-compose.yml and just comment them out? That will make it easier for someone to enable ClamAV if they want it.

mkuron:
@andryyy, I see you added the Dockerfile section last night. Now just remove the `max_size` loophole and add the build to Docker Hub, then it's done!

Issue #202, opened 2017-04-17
Incomplete IPv6 support with security implications


mkuron:
Docker translates all incoming IPv6 connections to IPv4. Due to this, all incoming IPv6 connections appear to be IPv4 connections coming from 172.22.1.1 from the perspective of the services (Dovecot, Postfix).

This has a few security-relevant consequences:
- [dovecot-anvil](https://wiki.dovecot.org/Authentication/Penalty) cannot slow down brute-force attempts coming in via IPv6.
- [fail2ban](https://andryyy.github.io/mailcow-dockerized/first_steps/#use-fail2ban) cannot block brute-forcing IPv6 hosts/networks trying to access Postfix or Dovecot.
- I'm not sure whether the web interface or SOGo have any authentication failure penalty, but if they do, the same thing applies unless you are using a [reverse proxy](https://andryyy.github.io/mailcow-dockerized/first_steps/#optional-reverse-proxy).
- Postfix postscreen and greylisting do not work for IPv6. Another consequence is that `Received:` headers added to emails received via IPv6 contain `172.22.1.1` instead of the remote host's IPv6 address.

Furthermore, Postfix cannot deliver outgoing emails via IPv6 because Docker's protocol translation only works for incoming connections.

This is not directly a problem of Mailcow (in fact, IPv6 worked on the standalone non-dockerized Mailcow), but rather a consequence of that its Docker containers are not set up in an IPv6-aware fashion. It is not clear to me how this can be fixed (as far as I am aware, Docker doesn't do NAT for IPv6 like it does for IPv4, so that whole port-exposing mechanism doesn't exist), but as it is security-relevant, it should be fixed soon. Perhaps [docker-ipv6nat](https://github.com/robbertkl/docker-ipv6nat) could be integrated into Mailcow-dockerized.

mkuron:
#203 was merged

PR #203, opened 2017-04-17
IPv6 support

mkuron:
Partial fix for #202 

To actually get IPv6 support, you additionally need https://github.com/robbertkl/docker-ipv6nat .

While ipv6nat can be run from a Docker container, there appears to be a bug (at least in my Debian 8 setup) that removes all IPv6 addresses from all interfaces on the host: https://github.com/robbertkl/docker-ipv6nat/issues/10 . Once that bug is fixed, I will update _docker-compose.yml_ to include the container. It should be enabled by default because we are in the year 2017 and IPv6 is no longer optional. Until then, I recommend the following:
```
curl -Lo docker-ipv6nat https://github.com/robbertkl/docker-ipv6nat/releases/download/v0.2.3/docker-ipv6nat.amd64
chmod +x docker-ipv6nat
./docker-ipv6nat
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -265,6 +265,14 @@ services:
           aliases:
             - nginx
 
+    ipv6nat:
+      image: robbertkl/ipv6nat
+      restart: always
+      privileged: true
+      network_mode: "host"
+      volumes:
+      - /var/run/docker.sock:/var/run/docker.sock:ro
+
 networks:
   mailcow-network:
     driver: bridge
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -268,10 +268,12 @@ services:
 networks:
   mailcow-network:
     driver: bridge
+    enable_ipv6: true
     ipam:
       driver: default
       config:
         - subnet: 172.22.1.0/24
+        - subnet: fd4d:6169:6c63:6f77::/64
 
 volumes:
   vmail-vol-1:


andryyy:
You need to start the Docker daemon with additional parameters. Also nat with IPv6 is neither default nor always wanted and a good idea. IPv6 works much different from v4. 

See https://docs.docker.com/engine/userguide/networking/default_network/ipv6/

This is much more complex and needs more discussion. Adding a container for nat with v6 makes me shake. :/



Beste Grüße 
André

mkuron:
You are absolutely right, but IPv6 is essentially useless by default in Docker. Have a look at the readme on https://github.com/robbertkl/docker-ipv6nat for an explanation why this solution is good. I was surprised that it actually works without further configuration -- it basically perfectly mimicks Docker's IPv4-NAT behavior for IPv6.

Docker without IPv6-NAT (as ugly a solution as it may be) cannot be used for a proper dual-stack mailserver as, for example, the Postfix and Dovecot containers would have different global IPv6 addresses. This is incompatible with the usual desire (and implicit assumption of Mailcow) to have a single host name for both IMAP and SMTP services. So the only alternative to using NAT is using a different host name for the web interface, the IMAP server and the SMTP server.

I wanted to start the discussion of how proper IPv6 support can be added and NAT seems to be a minimally-invasive solution without any adverse effects. If you look through the Docker issue tracker, you'll find that they are considering adding IPv6-NAT to Docker itself.

mkuron:
Also, this pull request only adds link-local IPv6 addresses to all containers without actually enabling NAT. So it can safely be merged without affecting any current setups. NAT would only be used if somebody feels ok about using it and fires up docker-ipv6nat himself.

By the way, having IPv6 in docker-compose is equivalent to starting the Docker daemon with the additional parameters.

mkuron:
The Docker bug has been fixed with 17.05.0-ce, which was just released a few hours ago. Therefore I have added the ipv6nat container to docker-compose.yml in this PR. I have already been running it for three weeks now and it works flawlessly. Note that it does not require any modifications to the Docker daemon config.

As Docker's IPv6 support is quite useless by default (it results in one global IPv6 address being used per container, which prohibits its use for things like Mailcow, where you need to share one host name for everything), there is no way around such a third-party solution that does IPv6 NAT. IPv6 NAT is built into the Linux kernel, so there is nothing exotic going on here. All that this extra container does is run some `ip6tables` commands that mimic the `iptables` commands Docker uses to expose IPv4 ports.

Robbert Klarenbeek has a good explanation [why Docker's own IPv6 support is insufficient](https://github.com/robbertkl/docker-ipv6nat#why-would-i-need-this) and [why IPv6 NAT isn't as crazy as it sounds](https://github.com/robbertkl/docker-ipv6nat#nat-on-ipv6-are-you-insane) for those with strong opinions about NAT. I'm also no friend of NAT, but as it is taking place within a single host and for a good reason, I think it's not overly objectionable.

andryyy:
Can you rebase the push to dev? Let's give it a try. :)

mkuron:
Here you go. Let me know if you run into any problems.

andristeiner:
Tested in a real live deployment, and compared to our current, "classic" dualstack mailcluster. Everything looks good so far :+1: 

After updating, `docker-compose up` will not work out of the box because the network needs to be recreated:
```
ERROR: Network "mailcowdockerized_mailcow-network" needs to be recreated - enable_ipv6 has changed
```
This might be a hurdle for unexperienced users and need to be communicated very well.

Braintelligence:
@andristeiner Yeah, the necessary steps to overcome this would be good to know! 😸

andristeiner:
@Braintelligence I've just removed not required networks with `docker network prune` before starting the new containers. For sure, you can also remove this particuler network with `docker network rm mailcowdockerized_mailcow-network`.

mkuron:
No need to manually remove networks. Just run `docker-compose down` before upgrading.

andryyy:
This "just" needs to be mentioned somewhere. It is not part of the docs.

K2rool:
A minor edit need to be done to docker-compose.yml  to add volume /lib/modules:/lib/modules:ro on my server ipv6nat is just restarting all the time.

The log's showed 
<pre>
ipv6nat_1            | 2017/05/14 15:38:19 exit status 3: modprobe: can't change directory to '/lib/modules': No such file or directory
ipv6nat_1            | ip6tables v1.6.0: can't initialize ip6tables table `filter': Table does not exist (do you need to insmod?)
ipv6nat_1            | Perhaps ip6tables or your kernel needs to be upgraded.
</pre>

The issue is caused by the ipv6 based kernel module's not been loaded on my server by default Centos 7.

Quote from robbertkl/docker-ipv6nat


> On some systems, IPv6 filter related kernel modules will not be loaded by default, and you'll see error messages in the log. Luckily, ip6tables will automatically load all necessary kernel modules for us, and it will even do so from within the container, since we're a privileged container anyway! To accommodate this, we need to mount the modules so ip6tables can load them: just add -v /lib/modules:/lib/modules:ro to the above docker run command.


Adding /lib/modules:/lib/modules:ro volume will ensure that issue is not going happen on anyone else system as iptables in the container will just load the needed kernel module if needed.

K2rool:
Hi, dose any config change's need to be done for dual stack to work on Mailcow, i got native ipv6 on the host node but ipv6 dose not seem to work on any of Mailcow container.

mkuron:
IPv6 works out of the box. Note that, if you have any firewall rules configured on the host (`ip6tables`), you may get conflicts that need to be resolved manually.

K2rool:
Work's out the box that what i thought was the case, it's probably not a iptable issue since i disabled firewalld on Centos by default the iptable rule allow everything anyway.

Thank's anyway i will just have to try to debug the issue.

andryyy:
Had a lot of problems on IPv4 only environments. Some stuff still needs to be fixed.

K2rool:
What kind of problems, I run Mailcow since that commit on a server with only IPV4 not noticed any issue.

mkuron:
> Had a lot of problems on IPv4 only environments. Some stuff still needs to be fixed.

I'm running a dual-stack environment and therefore haven't seen issues. Please provide additional details and report an issue at https://github.com/robbertkl/docker-ipv6nat

andryyy:
Most problems happened with the services and timeouts. Need to test it a bit more.

PR #204, opened 2017-04-17
Add license

mkuron:
License was missing after moving from mailcow to mailcow-dockerized


andryyy:
See other PR and comments, please.

mkuron:
Sorry, should have searched the closed ones too (#7). The issue isn't actually resolved though... It would only apply to the code anyway, to which you and the contributors have copyright; the Dockerfiles and config files which may have been collected from other places typically don't meet the minimum requirements for copyright.

PR #205, opened 2017-04-17
Add a "Forwarding Hosts" feature so messages from certain servers are never rejected

mkuron:
Some mail providers, e.g. [FastMail](https://www.fastmail.com/help/technical/spamchecks.html), offer a feature called "Forwarding Hosts". This allows one to add mail servers to a whitelist, ensuring that messages received from them are never rejected. They are still subjected to spam filtering, but any spam detected is moved to the Junk folder and not rejected.

The idea behind this feature is as follows: If you have an account on an external mail server (e.g. a Gmail account, a university account, etc.) and have configured it to forward all incoming mails to your "new" email address hosted on Mailcow, you want to ensure that all messages -- including forwarded spam -- are accepted. If Mailcow rejected any messages, the external mail server would generate a delivery status notification which would then get forwarded to you. This means you would still get one DSN for each rejected email, but instead of looking at sender and subject in your inbox, you have to sift through the headers to determine whether the message it refers to is actually spam or legitimate. In the latter case, you would then have to log into the external mail server and manually copy over the legitimate message.

This is my last pull request for the day...


mkuron:
I've made a little update so that instead of providing an IP address, you can now just enter a domain name, whose SPF record will then be fetched to determine the IP addresses of mail servers that might be forwarding emails. I also added proper input validation and provided a German translation.

Please have a look and let me know what you think.

andryyy:
Hi, looks awesome. :) Can you push this to dev and use the new SQL init function? Just use any other table as example.
Can you confirm black and whitelists still work when no forwarding host is defined? Sometimes strange things happen with these maps.

mkuron:
>Can you push this to dev and use the new SQL init function?

Here you go. By the way, I think you probably wanted to delete data/web/inc/init.sql when you rewrote the SQL init, but forgot to do so. The file is still there, but is no longer referenced anywhere (except for the `deny all` rule in the Nginx config).

> Can you confirm black and whitelists still work when no forwarding host is defined?

It was a bit unclear to me what rspamd does when it has a map without criteria, so I now changed the script to not include the rule when no hosts are defined. This should address your concern.

andryyy:
I really wonder what id does when I disable HELO checks and HELO with something like "x;echo 200 permit;". Or worse.

mkuron:
It doesn't do anything. Postfix's [`tcp_table` protocol](http://www.postfix.org/tcp_table.5.html) is very limited and furthermore there is no way a HELO would get into the script's stdin. postscreen just sends `get ` and the client's IP address to it (not the HELO!), and there is no way a client can inject a string into its IP address. Additionally, the script is executed while the TCP connection is being established, before a HELO is even received.

andryyy:
Ahh, okay, just saw the script while being in a rush and was curious. :)
Thanks!

andryyy:
Aaaand the panel collapse is broken. Will check this in a bit....

mkuron:
Woops, sorry abou that. Most likely it's the duplicated `id="accordion_access"`.

andryyy:
I removed the collapse stuff now, was about to change this anyway. This is a good opportunity. :)
I like this feature!

mkuron:
Glad to hear that. I was actually kind of surprised that none of the mailserver solutions you find on GitHub offer this -- its usefulness is obvious and it was a low-hanging fruit to implement.

PR #209, opened 2017-04-18
Fix CalDAV/CardDAV URLs displayed in SOGo web interface when used behind reverse proxy

### LINKED ISSUE #207: Incorrect CardDAV/CalDAV URLs in SOGo web interface behind reverse proxy

mkuron:
The `X-Forwarded-Proto` and `X-Forwarded-Port` headers are now used to pass the correct parameters to SOGo. Of course, they needed to be limited to trusted hosts, and since `if`s are not recommended in Nginx's configuration for performance reasons, I had to resort to `map`s. These can only exist in the `http` scope of the configuration and cannot contain boolean logic on the condition, so it's 20 lines, but easy to understand.

I have updated the documentation accordingly (Nginx now doesn't need `proxy_redirect` anymore; technically Apache also wouldn't need `ProxyPassReverse`, but I couldn't test that) and I have added a section about how to do it with HAProxy.

Fixes #207

### DIFF
--- a/data/conf/nginx/site.conf
+++ b/data/conf/nginx/site.conf
@@ -1,4 +1,23 @@
 proxy_cache_path /tmp levels=1:2 keys_zone=sogo:10m inactive=24h  max_size=1g;
+
+# use the non-standard X-Forwarded-* headers for WebObjects
+map $http_x_forwarded_proto $maybe_real_scheme {
+  default $http_x_forwarded_proto;
+  ''      $scheme;
+}
+map $http_x_forwarded_port $maybe_real_port {
+  default $http_x_forwarded_port;
+  ''      $server_port;
+}
+map $realip_remote_addr $real_scheme {
+  default $scheme;
+  172.22.1.1 $maybe_real_scheme;
+}
+map $realip_remote_addr $real_port {
+  default $server_port;
+  172.22.1.1 $maybe_real_port;
+}
+
 server {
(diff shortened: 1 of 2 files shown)


andryyy:
Thanks!

Have you tested it with ports other than 443? And without reverse proxy?

mkuron:
I technically haven't done either, but I have performed equivalent tests: I verified that whatever gets sent in `X-Forwarded-Port` gets used in the URL and I verified that if the request isn't coming from the whitelisted IP (172.22.1.1), the URL is generated as before.

andryyy:
Tested with Apache 2.4 as reverse proxy on port 80 to 8080 and it fails (redirects to 127.0.0.1 after login to SOGo).
I also cannot confirm the problem with the port you were having.

mkuron:
Please reopen the PR, this is definitely an issue.

Did you get a redirect to 127.0.0.1:80 or to 127.0.0.1:8080? I suspect it's the former. Such an issue could occur if the `ServerName` setting (Apache) or `server_name` (Nginx) on the reverse proxy doesn't match the `server_name` in Mailcow's Nginx. You can obviously work around it using `ProxyPassReverse`/`proxy_redirect` if you can't change the names to match. I can add these to the PR if you like. Alternatively, I could add an `X-Forwarded-Host` header, but I'd need to set up all three supported reverse proxies (Apache/Nginx/HAProxy) to ensure that the documentation is correct.

Issue #175, opened 2017-04-03
mailcow-setup-relayhost equivalent for mailcow-dockerized?


chaosbunker:
Hi André,

how would you recommend to configure mailcow-dockerized to do smtp-relay? in the regular mailcow version you have the script mailcow-setup-relayhost in the misc folder ..

what's best practice to do that in the dockerized version?

thanks and best regards:-)

andryyy:
Hi,

Insert this to `data/conf/postfix/main.cf`. "relayhost" does already exist (empty), just change its value.
```
relayhost = [your-relayhost]:587
smtp_sasl_password_maps = hash:/opt/postfix/conf/smarthost_passwd
smtp_sasl_auth_enable = yes
```

Create the credentials file:
```
echo "your-relayhost username:password" > data/conf/postfix/smarthost_passwd
```

Run:
```
docker-compose exec postfix-mailcow postmap /opt/postfix/conf/smarthost_passwd
docker-compose exec postfix-mailcow chown postfix: /opt/postfix/conf/smarthost_passwd /opt/postfix/conf/smarthost_passwd.db
docker-compose exec postfix-mailcow chmod 600 /opt/postfix/conf/smarthost_passwd /opt/postfix/conf/smarthost_passwd.db
docker-compose exec postfix-mailcow postfix reload
```

chaosbunker:
Thanks André! I ported the script you provided in mailcow 0.14 to mailcow-dockerized, based on your instructions here .. let me know if you are interested in adding it to mailcow-dockerized:-)

andryyy:
Sure, can you upload it somewhere or create a PR for review?

chaosbunker:
Created the PR ... git is still a bit foreign to me, so I hope the way I did everything is alright..

PR #210, opened 2017-04-18
port of mailcow-setup-relayhost from mailcow 0.14 to work with mailcow-dockerized

chaosbunker:
As referenced in #175 here is the script, which is working for me. Thanks for showing what needs to be done. I found the script quite helpful in mailcow 0.14 so I used the porting as a learning exercise.

in the conditional for restore-string I had some weird issues that I still don't understand.

to get the value for $RELAYHOSTCFG I first did 

`RELAYHOSTCFG=$(docker-compose exec postfix-mailcow postconf -h smtp_sasl_password_maps)`

but it kept messing up the output/echo of the restore-string .. the relayhostcfg array was always overwritten with the values of the usrpwd array .. they seemed to be displayed on top of each other .. very strange. if i created the array manually it worked fine. so I ended up pulling the relayhostcfg from data/conf/postfix/main.cf as that also worked.

so I just wanted to check on that again, when I noticed that 
`docker-compose exec postfix-mailcow postconf -h smtp_sasl_password_maps` 
now does not output anything .. it seems to refer to what is set in /etc/postfix/main.cf, because when I do 
`docker-compose exec postfix-mailcow postconf -c /opt/postfix/conf -h relayhost`
 it outputs the correct value. same for the other settings set .. i'm quite confused as to why it worked fine yesterday and today it doesn't. I tried with a clean install from your master, but get the same behavior O_o

also I had to set `smtp_sasl_security_options =` to get the relay to work properly .. without it the logs say 

> warning: SASL authentication failure: No worthy mechs found

in any case...take a look at the script and let me know if anything needs to change. i just tested it
again and emails are being relayed as expected


andryyy:
Thanks, let me check it. :) 

You may need to use docker-compose exec -T in scripts.

chaosbunker:
Thanks for the hint to use `docker-compose exec -T`, i just checked and it echoes the array properly now:-)

Still I don't understand why I have to pass `postconf -c /opt/postfix/conf` to get to the active config file .. shouldn't it automatically go there?

andryyy:
I just removed some old code and moved it to the root directory. :) Let's keep the helper scripts there for easier handling.

PR #212, opened 2017-04-20
Fix CalDAV/CardDAV URLs displayed in SOGo web interface when used behind reverse proxy

### LINKED ISSUE #207: Incorrect CardDAV/CalDAV URLs in SOGo web interface behind reverse proxy

mkuron:
Resubmitting an improved version of #209. Please give it another try.

I found out what caused the different behavior you observed with Apache. HAProxy, which I'm using, defaults to the equivalent of `ProxyPreserveHost on`. After I added `http-request set-header Host 127.0.0.1:8080` to replicate `ProxyPreserveHost off`, I got the same behavior as you did. Nginx should have also exhibted that behavior as it defaults to the equivalent of `ProxyPreserveHost off`.

I made one further modification so that the value of the `Host` header (which includes the port number unless it's 443) is passed through to `x-webobjects-server-url` and this ensures that SOGo's URL generation code (both for the CalDAV/CardDAV URLs and the login redirect) uses the exact same base URL as the client's web browser.

I have now tested different port numbers and going straight to Mailcow's Nginx without a reverse proxy and always saw correct behavior. 

Fixes #207.


andryyy:
It is important to test all cases on Nginx and Apache, too. I cannot assume it just works, though it looks good. :) You need to test every case, even if you don't use Apache2 or Nginx. And they need to be tested the same as you do with HAProxy. "Seems to work" or "should work" is not enough.

After all you are setting the hostname, port and scheme in the Reverse Proxy configuration, which is what I tried to avoid. :-/ But I see that there is/may be no way around. 

There is no way around the maps?

mkuron:
> It is important to test all cases on Nginx and Apache, too. I cannot assume it just works, though it looks good. :) You need to test every case, even if you don't use Apache2 or Nginx.

I have now tested Nginx, Apache and HAProxy with different port numbers, HTTPS enabled and disabled and also tried Mailcow without the reverse proxy and made one more change so that I can now confirm that it works in every situation. The change needed is related to Apache automatically adding an `X-Forwarded-Host` header. If I attempt to use `RequestHeader` to override it, it always appends to it instead. To fix that, `ProxyAddHeaders Off` was needed (which then necessitated setting `X-Forwarded-For` myself). We really don't want Apache adding `X-Forwarded-Host` automatically because it unconditionally trusts `X-Forwarded-Host` headers sent by the client by prepending them to the one it generates.

> After all you are setting the hostname, port and scheme in the Reverse Proxy configuration, which is what I tried to avoid. :-/

I now pull all the information (host, scheme, port) from server variables (even on Apache, where that requires mod_rewrite), so that should address some of your concern.

>  But I see that there is/may be no way around.

It's the only way that the web server can be made aware of the address that was used in the client's browser, so indeed there is no way around it.

> There is no way around the maps?

Unfortunately not. While Nginx has an [`if` directive](http://nginx.org/en/docs/http/ngx_http_rewrite_module.html#if), its use is strongly [discouraged](https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/) because it is inherently incompatible with the declarative philosophy of Nginx's configuration file.

andryyy:
Thanks! :)

andryyy:
Without RP after a login:

https://https/SOGo/so/mail@domain.tld/Mail

I will revert the changes and not accept further PRs about this issue as they seem to be not tested.

mkuron:
Oh dear, this is slightly embarassing. After testing it all, I seem to have made a copy-and-paste error while transferring the changes from my test system to the system that has my Git keys... @andryyy, please accept my apologies and let me resubmit a fixed version. Below is the trivial fix to the error that neither of us caught when we reviewed the patch.

```diff
 map $realip_remote_addr $real_host {
-  default $scheme;
+  default $host:$real_port;
   172.22.1.1 $maybe_real_host;
 }
```

I am still convinced that getting the fix in reverse proxy support merged is worthwhile. I almost certainly am not the only person who needs it. I would hate to let the work and extensive testing (which included setting up three different reverse proxies!) I put into this go to waste. One day, somebody else will run into the issue again, file an issue, submit a PR and most likely also miss one of the cases like I did here. Together, we can get this fixed now, once and for all.

chaosbunker:
Thanks for supplying the fix. As far as I understand, this would be a useful feature.

PR #221, opened 2017-04-26
Fix SOGo URLs behind reverse proxy again

### LINKED ISSUE #207: Incorrect CardDAV/CalDAV URLs in SOGo web interface behind reverse proxy

mkuron:
Fixes the URL that SOGo considers itself to be running under in reverse proxy setups and leaves them unaffected in direct connections. This is a resubmission of #212, which broke the latter and was therefore reverted.

@andryyy, I'm re-publishing this here because I'm sure some other people will find it useful. If you [don't want to merge it anymore](https://github.com/andryyy/mailcow-dockerized/pull/212#issuecomment-296961413), that's fine, but then please leave the PR open so people can find it and please update the documentation to indicate that reverse proxy setups report incorrect URLs without my patch.

I did extensive testing of all three documented reverse proxy servers in #212 and have now fixed the trivial config error that broke direct connections. Please test it yourself and let me know if there are any remaining problems. This is unlikely, but I can't guarantee that I covered all edge cases.

Fixes #207

### DIFF
--- a/data/conf/nginx/site.conf
+++ b/data/conf/nginx/site.conf
@@ -1,4 +1,31 @@
 proxy_cache_path /tmp levels=1:2 keys_zone=sogo:10m inactive=24h  max_size=1g;
+
+# use the non-standard X-Forwarded-* headers for WebObjects
+map $http_x_forwarded_proto $maybe_real_scheme {
+  default $http_x_forwarded_proto;
+  ''      $scheme;
+}
+map $http_x_forwarded_port $maybe_real_port {
+  default $http_x_forwarded_port;
+  ''      $server_port;
+}
+map $http_x_forwarded_host $maybe_real_host {
+  default $http_x_forwarded_host;
+  ''      $host:$real_port;
+}
+map $realip_remote_addr $real_scheme {
+  default $scheme;
+  172.22.1.1 $maybe_real_scheme;
+}
+map $realip_remote_addr $real_port {
+  default $server_port;
+  172.22.1.1 $maybe_real_port;
+}
+map $realip_remote_addr $real_host {
+  default $host:$real_port;
+  172.22.1.1 $maybe_real_host;
+}
+
 server {
   include /etc/nginx/conf.d/listen_ssl.active;
   include /etc/nginx/mime.types;
@@ -34,7 +61,7 @@ server {
   real_ip_recursive on;
 
   location = /principals/ {
-    rewrite ^ $scheme://$host:$server_port/SOGo/dav;
+    rewrite ^ $real_scheme://$real_host/SOGo/dav;
     allow all;
   }
 
@@ -96,8 +123,8 @@ server {
     proxy_set_header x-webobjects-server-protocol HTTP/1.0;
     proxy_set_header x-webobjects-remote-host $remote_addr;
     proxy_set_header x-webobjects-server-name $server_name;
-    proxy_set_header x-webobjects-server-url $scheme://$host:$server_port;
-    proxy_set_header x-webobjects-server-port $server_port;
+    proxy_set_header x-webobjects-server-url $real_scheme://$real_host;
+    proxy_set_header x-webobjects-server-port $real_port;
     client_body_buffer_size 128k;
     client_max_body_size 100m;
   }
@@ -110,8 +137,8 @@ server {
     proxy_set_header x-webobjects-server-protocol HTTP/1.0;
     proxy_set_header x-webobjects-remote-host $remote_addr;
     proxy_set_header x-webobjects-server-name $server_name;
-    proxy_set_header x-webobjects-server-url $scheme://$host:$server_port;
-    proxy_set_header x-webobjects-server-port $server_port;
+    proxy_set_header x-webobjects-server-url $real_scheme://$real_host;
+    proxy_set_header x-webobjects-server-port $real_port;
     client_body_buffer_size 128k;
     client_max_body_size 100m;
     break;
@@ -183,7 +210,7 @@ server {
   real_ip_recursive on;
 
   location = /principals/ {
-    rewrite ^ $scheme://$host:$server_port/SOGo/dav;
+    rewrite ^ $real_scheme://$real_host/SOGo/dav;
     allow all;
   }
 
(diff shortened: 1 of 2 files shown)


andristeiner:
Verified with nginx in front :+1: 

@andryyy As we always try to put everything behind nginx reverse proxies controlled by our automation, i would really appreciate if you could consider taking this one in. Im still a bit afraid of exposing containers to the outside and im pretty sure this is also what a lot of others do.

andryyy:
It needs to work without a reverse proxy, too. The only downside of the current implementation is the missing port in the DAV URLs, which I still cannot reproduce with Apache2 and Nginx in front of mailcow!

Maybe it is all about HAProxy and some missing headers?

@andristeiner https://andryyy.github.io/mailcow-dockerized/first_steps/#optional-reverse-proxy These docs still work fine. :/

mkuron:
> It needs to work without a reverse proxy, too.

It does now.

> The only downside of the current implementation is the missing port in the DAV URLs, which I still cannot reproduce with Apache2 and Nginx in front of mailcow!

Somehow SOGo needs to have learned about the port numbers in your setup. Apache/Nginx don't magically rewrite URLs in HTML code. Or are both your reverse proxy and the Nginx container running on port 443 and you have set the former to not listen on 127.0.0.1 and the latter to only listen on 127.0.0.1?

> Maybe it is all about HAProxy and some missing headers?

Which ones should that be? Nginx doesn't automatically add any headers.

chaosbunker:
I am curious what the difference is between this PR and the way to implement a RP that is shown in the docs?

i just set up mailcow-dockerized with a reverse proxy, as instructed in the docs and it works fine. was it correct to install nginx on the host and run that as the RP or would it make sense in any way or be of benefit to run a reverse proxy container like https://github.com/jwilder/nginx-proxy ?

what is the risk with exposing a container directly to the outside - or in other words, why is it maybe better to run nginx on the host as a RP?

mkuron:
> I am curious what the difference is between this PR and the way to implement a RP that is shown in the docs?

No idea. Somehow the RP's port number seems to be passed to Mailcow. HAProxy doesn't automatically add any headers which is why it doesn't work without this PR, but Nginx might -- though even then, I can't think of any header that would be trusted and evaluated by Mailcow.

> i just set up mailcow-dockerized with a reverse proxy, as instructed in the docs and it works fine. was it correct to install nginx on the host and run that as the RP or would it make sense in any way or be of benefit to run a reverse proxy container like https://github.com/jwilder/nginx-proxy ?

You can put it in a container if you like, it's only a matter of taste. Just make sure to NOT put it on the Mailcow network created by docker-compose. The traffic needs to pass through the host as that's the only IP trusted by Mailcow for _X-Forwarded-For_ headers.

> what is the risk with exposing a container directly to the outside - or in other words, why is it maybe better to run nginx on the host as a RP?

I'm not aware of any security benefits. The reason why one might want to run a reverse proxy in front of Mailcow is for the sharing of IPv4 addresses. If you have a reverse proxy, it can direct requests to various servers based on hostname.

chaosbunker:
> > I am curious what the difference is between this PR and the way to implement a RP that is shown in the docs?
> 
> No idea. Somehow the RP's port number seems to be passed to Mailcow. HAProxy doesn't automatically add any headers which is why it doesn't work without this PR, but Nginx might -- though even then, I can't think of any header that would be trusted and evaluated by Mailcow.

Not sure if I understand. Are you saying, that the instructions to set up a reverse proxy in the docs don't work with haproxy and this PR would make it work with apache, nginx and also haproxy? Or what does this PR do differently?

And thanks for answering my other questions:-)

mkuron:
Yes, this PR makes it work with all three because it explicitly passes through the protocol, hostname and port used for the frontend connection to the backend server.

andristeiner:
@andryyy Sorry, saw your answer just now, somehow i wasn't notified. However, you're absolutely right, it works with nginx even without this patch, it seems i mixed up something or maybe did not cleared my caches thoroughly.

However, Caldav/Carddav links within SOGo still end up as 127.0.0.1:8080 without this patch applied.

andristeiner:
> what is the risk with exposing a container directly to the outside - or in other words, why is it maybe better to run nginx on the host as a RP?

Its not so much about Docker, but about other measures already in place. For example, we have things like monitoring, SSL cipher settings and webserver updates already in place and automated over the whole infrastructure. If we can add services behind such a proxy, we don't have to bother about all those things manually for each software we're running.

mkuron:
> you're absolutely right, it works with nginx even without this patch
>
>  However, Caldav/Carddav links within SOGo still end up as 127.0.0.1:8080 without this patch applied.

Thanks for checking, @andristeiner. So this means Nginx doesn't magically learn the frontend port number. My patch is therefore also needed on Nginx when configured as documented and should thus be merged.

andryyy:
When we add server names and ports to the reverse proxy configuration (which I tried to avoid), I don't see the need of the maps.

And: I'm sorry, I was using Apache2 as reverse proxy while testing. Everything is fine here. So we should only have to adjust the Nginx configuration, which I will do now....

mkuron:
> When we add server names and ports to the reverse proxy configuration (which I tried to avoid), I don't see the need of the maps.

Unless you want to hard-code host name, scheme and port in Mailcow's Nginx config, you need the maps to conditionally pull that information from headers (i.e. only when they are available and when they come from the trusted 172.22.1.1). Nginx can't read environment variables, so I'm not sure how you could dynamically pass that information to Nginx.

mkuron:
@andryyy, abusing the `Host` header is a good idea. The only problem that leaves us with is _http_ vs. _https_. Any idea how to solve that? Right now any request going through the reverse proxy will have _http_ in the SOGo CalDav/CardDav URL. Even worse, the login redirect is also to _http_ (even though it is then typically redirected back to HTTPS), which means that cookies might be leaked to unencrypted connections.

andryyy:
Sigh, I just used a map. :P I cannot tell you how angry I was about this crap. I hope it works fine now. Tested it with Apache2 and Nginx as RP (SSL and plain) + no RP (SSL and plain).

I hope I did not miss anything...

mkuron:
@andryyy, thanks. Looks good. I didn't say it was easy...

As an alternative to the map, you could use https between the reverse proxy and Mailcow's nginx (i.e. make the protocol between client and RP match the one between RP and backend server). I actually already did that a few minutes ago (https://github.com/mkuron/mailcow-dockerized-docs/commit/6f1eda2af18bb7303b5be0f887e768431b4ec275), but you were quicker in adding the maps. At least with the maps we don't waste CPU cycles on SSL-encrypting the backend connection.

PR #236, opened 2017-05-01
Add documentation on how someone could integrate Gogs in his organization.

ecdsa521:
This is fairly trivial but I wanted to describe it in detail on how one could accomplish this. 

We're reusing already available docker image `gogs/gogs`.

Not sure if the IP I picked for Gogs could be standardized in any way, I just picked one that was free.

edit: Sorry for being an idiot with pull request. I think I picked wrong source branch and it now wants to merge 11 commits. Not sure what to do about it.


chaosbunker:
Woah cool, i was thinking about something like this earlier, when i mentioned my gogs installation on another server - will try! Thanks :-)

mkuron:
A lot of software can use IMAP or SMTP as an authentication backend (Owncloud/Nextcloud comes to mind, but there are many others). Do you think it makes sense to add each and every one of these to the Mailcow documentation? Gogs is not related to Mailcow's mailserver theme and it doesn't even need to be run from within Mailcow's docker-compose setup. Just create a new docker-compose.yml in a separate directory (or even a separate server, if you like), spin up MySQL and Gogs containers and configure them to use Mailcow's SMTP server via its public address.

ecdsa521:
It's just something I use now and wanted to document the process. I think many organizations could benefit from such integration, making mailcow a driver for everything an organization may need, but if it's worth including in official docs? No idea. I just wanted to leave this here.

chaosbunker:
@mkuron That's a good point - I personally run on my server mailcow 0.14 with seafile and gogs. when i think of using mailcow-dockerized soon instead of mailcow 0.14, i automatically think i want it "all in the same place" again. But it's a good point to keep it separated from other services that are kind of unrelated. I actually prefer that - but my problem is, that I am just beginning to understand docker-compose and have no clear understanding of how to run other docker containers beside mailcow-dockerized, on the same server (e.g. a wordpress install, seafile or in this case gogs) :-/
Does it only work with a reverse proxy, if I want to keep other stuff outside of the mailcow-network?

Anyway, I am grateful for hints or examples on how to integrate other stuff into mailcow, because it helps learn docker. But ultimately, i prefer running unrelated services separated from mailcow

mkuron:
You need a reverse proxy if you want multiple services running on the same port (80/443). The reverse proxy looks at the host name (gogs.example.com, mailcow.example.com) or URL (example.com/gogs, example.com/mailcow) and distributes requests to the appropriate backend webserver. That's not specific to Mailcow or Docker, you'd need it even if you installed Gogs and Seafile natively. In fact, Mailcow's Nginx container is a reverse proxy itself, proxying certain requests to the servers built into SOGo or rspamd.

chaosbunker:
thanks @mkuron for clarifying!

@ecdsa521, just to let you know - following your instructions i cannot get it to work. when navigating to the install page, it redirects from `https://${MAILCOW_HOST}/gogs` to `https://${MAILCOW_HOST}/install` and gives a 404, instead of `https://${MAILCOW_HOST}/gogs/install`

if i go straight to `https://${MAILCOW_HOST}/gogs/install` the page loads, but css etc is not found, because the links are all missing the `/gogs/` part

ecdsa521:
@chaosbunker I should mention that initial install should be done through port open on the host, since we only let Gogs know where it's running after install (in step 8, `ROOT_URL`).

chaosbunker:
Maybe you can update your instructions so that how to do that is clear - I for example have no idea right now how to "do the initial install through port open on the host" :-)

ecdsa521:
Go to `http://yourhost:3000/` if your `GOGS_WWW_PORT` is 3000. Also while the ports can be arbitrary I suggest changing `GOGS_WWW_PORT` to 3000 and `GOGS_SSH_PORT` to 4000 - my initial instruction had them swapped, and that made little sense since Gogs listens on 3000 by default.

chaosbunker:
still no luck - getting bad gateway 503

i have this in `data/gogs/gogs/conf/app.ini`

[server]¬
DOMAIN           = localhost
SSH_DOMAIN▸ ▸    = ${MAILCOW_HOST}
SSH_LISTEN_PORT▸ = 22
HTTP_PORT        = 30
ROOT_URL         = https://${MAILCOW_HOST}/gogs/
DISABLE_SSH      = false
SSH_PORT         = 22
START_SSH_SERVER = false
OFFLINE_MODE     = false

and i did switch GOGS_WWW_PORT and GOGS_SSH_PORT

ecdsa521:
Are you getting bad gateway by going to port :3000? Do you literally have variables, eg. `${MAILCOW_HOST}` in the Gogs' config? It cannot have that I'm afraid.

Can you do `curl http://172.22.1.123:3000/`? Can you `ssh -v 172.22.1.123` (as in: does it connect at all?)

chaosbunker:
No of course not - i am using a FQDN - i am just using ${MAILCOW_HOST} to make clear it equals the mailcow hostname :-)

weird, in the first go (when the port numbers where switched to 3000/ssh 4000/www it said "found", now it says `curl: (7) Failed to connect to 172.22.1.123 port 3000: Connection refused`

ecdsa521:
Perhaps nuke the data/gogs directory and start again from the install step? 
Is there anything in the logs? (`docker-compose logs gogs-mailcow`)

Also note that usually gogs' web interface comes few seconds after the container.

chaosbunker:
forgot to include:

`ssh -v 172.22.1.123`
```
debug1: rekey after 134217728 blocks
debug1: SSH2_MSG_NEWKEYS sent
debug1: expecting SSH2_MSG_NEWKEYS
debug1: rekey after 134217728 blocks
debug1: SSH2_MSG_NEWKEYS received
debug1: SSH2_MSG_EXT_INFO received
debug1: kex_input_ext_info: server-sig-algs=<ssh-ed25519,ssh-rsa,ssh-dss,ecdsa-sha2-nistp256,ecdsa-sha2-nistp384,ecdsa-sha2-nistp521>
debug1: SSH2_MSG_SERVICE_ACCEPT received
debug1: Authentications that can continue: publickey,keyboard-interactive
debug1: Next authentication method: publickey
debug1: Trying private key: /home/xxx/.ssh/id_rsa
debug1: Trying private key: /home/xxx/.ssh/id_dsa
debug1: Trying private key: /home/xxx/.ssh/id_ecdsa
debug1: Trying private key: /home/xxx/.ssh/id_ed25519
debug1: Next authentication method: keyboard-interactive
debug1: Authentications that can continue: publickey,keyboard-interactive
debug1: No more authentication methods to try.
Permission denied (publickey,keyboard-interactive).
```

edit:
docker-compose logs gogs-mailcow seemed fine until i nuked data/gogs :-P

chaosbunker:
Working now. I only changed

```
[server]
ROOT_URL         = https://${MAILCOW_HOST}/gogs/

```
do i need this?
```
SSH_LISTEN_PORT = 22
SSH_DOMAIN = [domain where ssh is available]
```

ecdsa521:
SSH_DOMAIN is important as that's what Gogs will use to create ssh clone urls. Might want to change it or keep as is, depends if you're satisfied with default.

chaosbunker:
Ah, I see. I added it. And i guess i don't need to add /gogs/ after SSH_DOMAIN, correct? like ${MAILCOW_HOST}/gogs ?

one thing does not work: if i add an ssh key in gogs UI and then try to clone a test repo via ssh, it asks for a password. so it seems that the private key i am using from my laptop does not get passed through to the container?

ecdsa521:
No, SSH_DOMAIN should be just the domain, nothing else. 

The second problem is that you're likely hitting your host's ssh, not the one running in gogs' docker. Is your clone url including ssh port (in this example, 4000)?

chaosbunker:
no, the clone URL looks like this: `git@mx.domain.com:username/test.git`

and `data/gogs/gogs/conf/app.ini` looks like this
screen shot 2017-05-02 at 01 30 07

ecdsa521:
Yeah, what's your `GOGS_SSH_PORT` in `mailcow.conf`? Gogs' app.ini and ports there relate to what's happening inside the container.

chaosbunker:
GOGS_SSH_PORT=4000

ecdsa521:
Yeah, so you told Gogs that it can use port 22 and it makes clone urls with that. 

But when the outside world tries to connect to your server at port 22 it likely hits the host's, real, main ssh server. You can probably see those attempts in your /var/log/auth.log. 

Instead you could
a) change real ssh port to something else and set `GOGS_SSH_PORT=22` in `mailcow.conf`
b) if you have multiple ip's, make it so real ssh is bound to one, and `GOGS_SSH_PORT=other.ip:22`
c) tell gogs that SSH_PORT=4000 and it will create clone urls that reflect it.

chaosbunker:
Ok, cool - thanks for taking the time and explaining! I successfully cloned the repo via ssh using method c)

but none of the options is ideal, i find

regarding a) i am not a fan of changing the default ssh port of the server
if i use c) the clone url looks horrible :-p `ssh://git@mx.domain.com:4000/username/test.git` - not the end of the world, but i like it the way i am used to it git@domain:user/repo.git
and i don't have an extra ip and getting one is an extra 2€/month, so not really an option for this either:-)

ecdsa521:
Probably not a good idea but I found incredibly hacky way to accomplish this... but it works. 

1) create git user on host
2) edit out `allow-agent-forwarding,no-pty` from `data/gogs/git/.ssh/authorized_keys` (probably need to make a script to do do that, and run every minute from cron. This file is managed by gogs and we're rewriting it to accomplish something stupid)
3) hardlink `data/gogs/git/.ssh/authorized_keys` to `/home/git/.ssh/authorized_keys`
4) add to host's `/etc/ssh/sshd_config`:
```
Match User git
    ForceCommand ssh -A -oProxyCommand="nc 172.22.1.123 22" git@localhost
```
5) Make sure agent forwarding is allowed on your computer. Eg. add `AllowAgentForwarding yes` to your ~/.ssh/config

What it does is letting you connect to host's git user,  only to be forced into ssh into gogs in such way that your ssh agent is preserved.
 
But tbh I'd probably go with option b) and that way cut on stupid people trying to bruteforce real server's SSH. Instead they'd be knocking on this docker instance. Or just pay 2 euro. Or maybe use ipv6. Or a vpn (connect to real host only via vpn, otherwise to gogs)

chaosbunker:
Haha, thanks - yeah that's pretty hacky and not really ideal, even though it would be close to what i wanted :-P

I think i would just live with the ugly url for now:-)

PR #237, opened 2017-05-01
Switch debian:testing to debian:stretch

mkuron:
This is in preparation of the Debian 9 release, coming up in a few weeks. At that point, stretch will graduate from testing to stable and testing will start pointing at buster. Since we wouldn't want to automatically be switched to a highly unstable version, this change explicitly uses stretch in all Dockerfiles.


andryyy:
Thanks, but please push to dev. I added it now and will merge it. :)

PR #250, opened 2017-05-06
Disable greylisting for forwarding hosts

mkuron:
The help text already stated that greylisting is disabled for the trusted forwarding hosts, but I seem to have forgotten to add the appropriate rule in #205.

### DIFF
--- a/data/conf/rspamd/dynmaps/settings.php
+++ b/data/conf/rspamd/dynmaps/settings.php
@@ -58,6 +58,7 @@ endforeach;
 		apply "default" {
 			actions {
 				reject = 999.9;
+				greylist = 999.8;
 			}
 		}
 		symbols [


andryyy:
Can you try it with

andryyy:
Can you try it with `want_spam = yes;`? I think that's the way to go, but I am unsure this disables all prefilter checks. I asked in #rspamd :)

mkuron:
It's unclear to me what `want_spam = yes` really does. The only occurence of that setting in the code is https://github.com/vstakhov/rspamd/blob/a1dc1fd0591d0880c1537c4032d404cc748537f6/src/plugins/lua/settings.lua#L489. It looks like `want_spam = yes` behaves like a wildcard whitelist rule, but I can't tell where in the processing chain that is applied.

`want_spam` is also pretty much undocumented, so I'd stay away from it for now and just stick to the hack of setting the action scores to high values.

andryyy:
Or we just apply a custom symbol with a high priority pre-filter. 

I will check want_spam and report.

mkuron:
Looking at the code more closely, it seems like `want_spam` also disables adding a header to spam mails. So this is definitely not what we want.

I suspect the pre-filter approach you suggested will have the same problem. We still want messages from the forwarding hosts to pass through rspamd for spam and virus checking, we just don't want to ever reject them.

andryyy:
Why would you want to scan them when you never reject or greylist? That's pretty much exactly what we need. 

Anyway you can run a forced action when a symbol is present. There is a module for it. Keep in mind that when you omit required_action or honor_action, it runs as a prefilter and detecting the symbol you added with settings will not work (symbol must be known to Rspamd). You will find the error in configtest.
So "honor_action = no action" in forced actions should work for your symbol added within settings. 

But I really don't know why you want to scan these mails. I will use the want_spam parameter for this. 


Beste Grüße 
André

mkuron:
> Why would you want to scan them when you never reject or greylist?

There is a third possible action you can apply to spam, the "add header" action.

> But I really don't know why you want to scan these mails

Let me reiterate the purpose of the forwarding hosts feature as per #205: Let's assume you have an old Gmail account. You configure it to forward all incoming emails (which includes spam) to your new Mailcow-hosted email account. There, you want spam filtering to be applied and spam to be sorted into the "Junk" folder. You do not ever want to have any messages rejected as Gmail's forwarding server would then send a bounce message to your Gmail account, which is again forwarded to your Mailcow account. There you see the bounce message in your inbox and you need to carefully read it to find out sender and subject of the bounced message to determine whether it was legitimate or spam. Instead of having to manually read every such bounce message, the forwarding hosts feature ensures that such bounces never happen and the offending messages are instead filed into the "Junk" folder, where you can easily browse through them.

So emails coming from forwarding hosts should either have no action applied (if their score is below 5) or have the "add header" action applied (if their score is above 5). "greylist" or "reject" must never be applied to them.

mkuron:
ecda4fb1d1f1df71fc7da2c2754182e43b2abf40 disables spam filtering entirely for forwarding hosts. I will submit another PR where one can selectively enable spam filtering for forwarding hosts -- this should make both me and @andryyy happy.

PR #255, opened 2017-05-06
Move to Alpine Nginx image

K2rool:
Hi, i recently migrate from older non docker version of Mailcow few days back quite liking it compered to the normal version.

One of the downsides of this version is the amount of disk space it is using for the 12 containers the image's are taking up 3,421.5MB which could be a issue for user running it on vps with small disk's.

If you switch Nginx  nginx:mainline-alpine (109MB to 15.5MB),Memcached  memcached:alpine (83.7MB to 6.9MB) and Redis redis:alpine (184MB to 19.8MB) to the Alpine version of the image you can save 334MB.

I've not tested if the Alpine version of Memcached or Redis work's properly as i don't have any experience of using Memcached or Redis , but the Alpine based image for nginx work's perfectly.

Am thinking the Mailcow image's could be rebased to Alpine as Alpine based image's are generally smaller but might cause more issue that it worth to just save some disk space.

Am not 100% sure if this changed should be merge but i will leave that to you andryyy to decide.

Ryan

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -233,13 +233,13 @@ services:
       depends_on:
         - sogo-mailcow
         - php-fpm-mailcow
-      image: nginx:mainline
+      image: nginx:mainline-alpine
       healthcheck:
         test: ["CMD", "ping", "php-fpm-mailcow", "-c", "10"]
         interval: 10s
         timeout: 30s
         retries: 5
-      command: /bin/bash -c "envsubst < /etc/nginx/conf.d/templates/listen_plain.template > /etc/nginx/conf.d/listen_plain.active &&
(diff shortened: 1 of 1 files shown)


psi-4ward:
I use `nginx:stable-alpine`, `redis:3.2-alpine`.

andryyy:
Hi,
I switched them over to alpine in dev and didn't run into problems.
Redis and Memcached should never be a problem. I was concerned about Nginx, but it runs just fine and as we don't use any OS dependent stuff there.
Thanks!

K2rool:
Since you did the commit's you self i will close this pull.

Ryan

PR #256, opened 2017-05-07
Optionally enable spam filter for forwarding hosts

mkuron:
This PR adds a boolean option to the forwarding hosts feature so that one can optionally turn on the spam filter for certain forwarding hosts. #205 introduced that feature default-on, ecda4fb in the context of #250 switched it to default-off.

See [#250](https://github.com/mailcow/mailcow-dockerized/pull/250#issuecomment-299633973) for why there are situations where you want to apply the "add header" action to spam received from forwarding hosts.


mkuron:
Updated for @andryyy's changes to the admin UI from last night. It would be nice if we could get this into dev before you merge your recent changes to master.

data/conf/rspamd/dynmaps/settings.php
```diff
@@ -31,19 +31,52 @@ function parse_email($email) {
 
 ?>
 settings {
-<?php
 
 /*
 // Start whitelist for forwarding hosts
 */
```
andryyy, line 37: /* */ Outside PHP will destroy the settings map.

data/conf/rspamd/dynmaps/settings.php
```diff
@@ -31,19 +31,52 @@ function parse_email($email) {
...
+?>
+	whitelist_forwarding_hosts_with_spam_filter {
+		priority = high;
+<?php
+foreach ($rows as $host):
+?>
+		ip = "<?=$host;?>";
+<?php
+endforeach;
+?>
```
andryyy, line 58: I will merge it, but remove the "apply" directives. We should use multimaps with matching symbols. This will work by using honor_actions.

andryyy:
I will merge it and change the multimap stuff now. Thanks for the new implementation!

mkuron:
You were probably quicker as I wasn't able to have a look at this until just now. However, I just did https://github.com/mkuron/mailcow-dockerized/commit/a9e8418eab6b9a8d8278e880a270d11ddd493fef that fixes the broken comment and replaces the `apply`s with `require_action`s (instead of the `honor_action`s you suggested). It isn't clear to me though what you wanted to do with `multimap`s.

If https://github.com/mkuron/mailcow-dockerized/commit/a9e8418eab6b9a8d8278e880a270d11ddd493fef is what you meant, I can create a PR from it.

andryyy:
I moved it to Redis like all the other maps. The checkbox and table layout was broken. You cannot push if you didn't even bother to check the UI. :( It wasn't very important, the table will be drawn by the API soon anyway. :)

Is all of this (esp. spf.inc.php + forwardinghosts.php) your own code? Just want to make sure. :) They kind of differ from each other.

mkuron:
> I moved it to Redis like all the other maps.

I think I found one error in your code and fixed it in #259. It looks like the actions weren't applied to the right symbol, but I couldn't test that. So please have a look yourself and merge with caution.

> The checkbox and table layout was broken. You cannot push if you didn't even bother to check the UI. :( It wasn't very important, the table will be drawn by the API soon anyway. :)

I tested everything on current master and then rebased to dev. The only conflict was in the admin UI and that was trivial to resolve. The HTML code in this PR still looks sensible to me, so I'm not sure what breakage your referring to. Maybe it only occurred in combination with the more extensive UI changes you made on dev?

> Is all of this (esp. spf.inc.php + forwardinghosts.php) your own code? Just want to make sure. :) They kind of differ from each other.

Yes, they are my own. With "different" you probably mean that one is nice and has comments and the other one doesn't? For the SPF parser, I just programmed along the specification, while the other one I wrote pretty ad-hoc.

andryyy:
Yes, they just differ in coding style etc. Or at least I thought. They are great! Thanks for contributing them. Feel free to add any header to the files.

PR #259, opened 2017-05-09
Consistent symbol names for forwarding hosts

mkuron:
After reading d64ed65575e3a287a78ba7f976b96dd1006f7a9a, it seemed like multimap.conf and force_actions.conf need to refer to forwarding hosts by the same symbol name but didn't. Please check whether this is really the case or due to my misunderstanding. I could not tested this change yet.


andryyy:
Oh, damn, thanks!

PR #260, opened 2017-05-09
Remove un-named docker volume for sogo

pschiffe:
This is the only un-named docker volume used in the compose file.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -133,7 +133,7 @@ services:
       dns_search: mailcow-network
       volumes:
         - ./data/conf/sogo/:/etc/sogo/
-        - /usr/lib/GNUstep/SOGo/WebServerResources/
+        - sogo-vol-1:/usr/lib/GNUstep/SOGo/WebServerResources/
       restart: always
       networks:
         mailcow-network:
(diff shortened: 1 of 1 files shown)


mkuron:
I don’t think this should even be a volume. The WebServerResources come from the sogo-common apt package, which is installed when the Dockerfile is built.

Mounting a volume just copies all the stuff that’s in the path upon first launch of the image. And if we use a named volume, I think the stuff is only copied upon first launch of the container, which might cause trouble with subsequent updates.

I don’t think we’re gaining anything by having the container, so just remove those two lines entirely. @andryyy, do you recall why you added that unnamed volume in the first place? Was that so you could mount it into the Nginx container, which you didn’t end up doing (instead you put a `python -m SimpleHTTPServer` into the Sogo container)?

andryyy:
I think I tried mounting it, then did this, that, this etc. and forgot it. I think it can be removed. Almost 95% sure. :P

pschiffe:
I've updated the PR, but feel free to close it and create a new one.

mkuron:
Was resolved in 4cb8596ff001ada7d0adab56343fc1c7bddf794a. This PR can be closed.

PR #267, opened 2017-05-13
Follow best practices in Dockerfiles

michael-k:
I tried to create small commits to ease reviewing.  Feel free to squash/cherry-pick/…

I could build all of them locally with the exception of sogo where a `keyserver error` regarding the key `0x810273C4` prevented me from doing so.


andryyy:
Looks good to me. Thanks. :) 

You may need to try it a few times for SOGo, as the keyserver URL is a pool with some broken servers in it.

Issue #281, opened 2017-05-16
DKIM signing fails


swdee:
Using commit 861ade77ab42f348dc4484491f946091af56115a (14th May) setting up mailcow with DKIM signing results in DKIM signature to fail upon verification.  

To reproduce the issue, setup mailcow, create DKIM keys, add DNS TXT record, then send test email to 	auth-results@verifier.port25.com or other mail testing provider.  Resulting report indicates DKIM check fails

```
==========================================================
Summary of Results
==========================================================
SPF check:          pass
DomainKeys check:   neutral
DKIM check:         fail
SpamAssassin check: ham
```

With reason;
```
----------------------------------------------------------
DKIM check details:
----------------------------------------------------------
Result:         fail (signature doesn't verify)
```

I have checked Redis keys DKIM_PRIV_KEYS and DKIM_SELECTORS to make sure they are populated with data (they are).   

I added debug config to data/conf/rspamd/override.d/logging.inc to cause rspamd to output DKIM signing debug with;
```
debug_modules = ["dkim"]
```

When sending mail it appears to sign it, but must calculate the wrong hash.  rspamd log output;

```
rspamd-mailcow_1     | 2017-05-16 09:08:34 #8(normal) <f1bf04>; task; accept_socket: accepted connection from fd4d:6169:6c63:6f77::6 port 39880, task ptr: 00007F82A8B82130
rspamd-mailcow_1     | 2017-05-16 09:08:34 #8(normal) <f1bf04>; task; rspamd_message_parse: loaded message; id: <1603319.bLJElE5Sn2@vm32>; queue-id: <F342B3DEAC>; size: 647; checksum: <1b673c983ef7b3a2922cec85ed88684d>
rspamd-mailcow_1     | 2017-05-16 09:08:34 #8(normal) <f1bf04>; lua; settings.lua:262: check for settings
rspamd-mailcow_1     | 2017-05-16 09:08:34 #8(normal) <f1bf04>; task; fuzzy_generate_commands: <1603319.bLJElE5Sn2@vm32>, part is shorter than 100 bytes (3 bytes), skip fuzzy check
rspamd-mailcow_1     | 2017-05-16 09:08:34 #8(normal) <f1bf04>; task; dkim_symbol_callback: skip DKIM checks for local networks and authorized users
rspamd-mailcow_1     | 2017-05-16 09:08:34 #8(normal) <f1bf04>; task; spf_symbol_callback: skip SPF checks for local networks and authorized users
[...]

Issue #272, opened 2017-05-14
DKIM Selectors Empty


stevesbrain:
After upgrading to the latest image (git pull + docker compose pull + docker compose up -d) I've noticed that my DKIM keys list is empty. I try to create a key for an existing image domain, and get told this isn't possible. Create one for a domain that hasn't had one before - works fine.

So I removed the volume containing DKIM keys, then docker-compose up -d again, and I still had all the new keys I had created before. Looks like they're getting stored elsewhere perhaps. I believe this is probably tied with what's being experienced in #271

mcoreix:
Same issue :(

Braintelligence:
For many occurrences you need to docker-compose down first. Did you try that?

mcoreix:
But docker-compose down will also delete volumes !? where data is stored
or am i wrong?

Braintelligence:
#233

andryyy:
DKIM keys are now stored in Redis. It migrates the keys when you recreate the phpfpm container after running docker-compose pull && docker-compose up -d.
The keys will be deleted from the volume and the volume remains empty until we remove it in the future.

Everthing sounds fine, except it did not list the keys from Redis. But this _could_ be a caching issue. Or some missing file updates from git.

Do the keys disappear when you reattach the DKIM volume?

felixoi:
I had the exactly same issue. The old DKIM selectors are not listed, but in a fresh Installation all new selectors are listed.

Braintelligence:
I planned on updating two mailcow instances today but now I'll rather wait :p

Daniel133:
+1

andryyy:
This does not help at all.

I need information like .. does it work after removing the DKIM volume?

andryyy:
Damn. I pushed a new image mailcow/phpfpm... For those who already upgraded, do this please:

Check your domains DNS settings and copy your DKIM TXT record, like this:
```
[...]

PR #286, opened 2017-05-17
DKIM redis migration: store private key correctly

mkuron:
Even after the past weekend's fixes to the broken SQL-to-Redis migration for the DKIM keys (see #272), it still wasn't working for me. I was getting errors like the following in the rspamd logs:

```
dkim_signing.lua:179: cannot make request to load DKIM key for dkim.my-domain.com: nil
```

When comparing [`dkim_add_key()`](https://github.com/mailcow/mailcow-dockerized/blob/0cfd0b55bd88a7d682cae32fd63b2e03db0d93cb/data/web/inc/functions.inc.php#L2403) in the PHP code with the [migration script](https://github.com/mailcow/mailcow-dockerized/blob/e795898b80fabad951f70b25bca5a50062af1961/data/Dockerfiles/php-fpm/docker-entrypoint.sh#L53) or the [rspamd source code](https://github.com/vstakhov/rspamd/blob/ab587d5/src/plugins/lua/dkim_signing.lua#L187), it is clear that the problem is that the private key wasn't stored with the selector -- it was being stored as my-domain.com instead of as dkim.my-domain.com.

#281 could potentially be related to this.

If you have already upgraded, you will need to manually fix things as follows:
Run `source mailcow.conf ; docker-compose exec redis-mailcow redis-cli` in the mailcow-dockerized directory and get the private key as follows:
```
HGET DKIM_PRIV_KEYS my-domain.com
```
Now store that private key again with the selector prefixed to the domain:
```
HSET DKIM_PRIV_KEYS dkim.my-domain.com "-----BEGIN PRIVATE KEY-----[...]\n-----END PRIVATE KEY-----"
```

### DIFF
--- a/data/Dockerfiles/php-fpm/docker-entrypoint.sh
+++ b/data/Dockerfiles/php-fpm/docker-entrypoint.sh
@@ -51,7 +51,7 @@ for file in $(ls /data/dkim/keys/); do
   domain=${file%.dkim}
   if [[ -f /data/dkim/txt/${file} ]]; then
     redis-cli -h redis-mailcow HSET DKIM_PUB_KEYS "${domain}" "$(cat /data/dkim/txt/${file})"
-    redis-cli -h redis-mailcow HSET DKIM_PRIV_KEYS "${domain}" "$(cat /data/dkim/keys/${file})"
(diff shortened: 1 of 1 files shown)


andryyy:
Thanks, that's right.

I will close this PR and rebuild the keys within the script, so users do just need to pull the new image and restart it.

andryyy:
Thanks, fixed it now. I promise I will not mess it up this much in the future. :-)

PR #297, opened 2017-05-19
Turning autocomplete off for Yubi Tokens

stevesbrain:
Autocomplete was on for Yubi Token auth prompt. This made multiple "historical" tokens pop up which is a small annoyance and is not necessary/best practice to have one time tokens recorded/kept in this manner :)


andryyy:
Thanks! :)

PR #311, opened 2017-05-25
Fail2ban-like authentication rate limiting for Postfix, Dovecot and SOGo

mkuron:
Currently, Mailcow does not offer a built-in mechanism for limiting failed login attempts to its services in a fashion similar to what fail2ban does. While there is [documentation](https://mailcow.github.io/mailcow-dockerized-docs/firststeps-syslog/#fail2ban) on how to integrate the latter, it requires manual work, including modifications to the Docker daemon's configuration.

This PR adds a container that runs a simple Python script, which streams the service logs via the Docker API. If it detects failed login attempts, it adds the client's IP address to a list and blocks it via `iptables` if more than 10 attempts are received within 30 minutes. The client is automatically unblocked after 30 minutes of no login attempts.

The client IP address is truncated to 24 bits (IPv4) or 64 bits (IPv6) so that an attacker cannot evade the protection by using different addresses from his provider-allocated network. The `FORWARD` rules protect the dockerized services, while the `INPUT` rules prevent connections to a reverse proxy running on the host.

Dovecot, Postfix and SOGo now all have a common counter for failed authentication attempts. This only leaves one place unprotected: the Mailcow UI. It has a [rudimentary protection](https://github.com/mailcow/mailcow-dockerized/blob/fa9a058cf87cce5de30de9322595c24cb7b73538/data/web/inc/functions.inc.php#L173) already, but it is useless in practice as it can be evaded by deleting the session cookie before every authentication attempt. Perhaps in the future it can be changed to also log failed authentication attempts so that we can have one central instance for rate-limiting authentication failures.

### DIFF
--- a/data/Dockerfiles/fail2ban/Dockerfile
+++ b/data/Dockerfiles/fail2ban/Dockerfile
@@ -1,0 +1,8 @@
+FROM python:2-alpine
+LABEL maintainer "Andre Peters <andre.peters@servercow.de>"
+
+RUN apk add -U --no-cache iptables ip6tables
+RUN pip install docker
+
+COPY logwatch.py /
+CMD ["python2", "-u", "/logwatch.py"]
--- a/data/Dockerfiles/fail2ban/logwatch.py
+++ b/data/Dockerfiles/fail2ban/logwatch.py
@@ -1,0 +1,103 @@
+#!/usr/bin/env python2
+
+import re
+import time
+import atexit
+import signal
+import ipaddress
+import subprocess
+from threading import Thread
+import docker
+
+RULES = {
+	'mailcowdockerized_postfix-mailcow_1': 'warning: .*\[([0-9a-f\.:]+)\]: SASL .* authentication failed',
+	'mailcowdockerized_dovecot-mailcow_1': '-login: Disconnected \(auth failed, .*\): user=.*, method=.*, rip=([0-9a-f\.:]+),',
+	'mailcowdockerized_sogo-mailcow_1': 'SOGo.* Login from \'([0-9a-f\.:]+)\' for user .* might not have worked',
+}
+BAN_TIME = 1800
+MAX_ATTEMPTS = 10
+
+bans = {}
+quit_now = False
+
+def ban(address):
+	ip = ipaddress.ip_address(address.decode('ascii'))
+	if ip.is_private or ip.is_loopback:
+		return
+	
+	net = ipaddress.ip_network((address + ('/24' if type(ip) is ipaddress.IPv4Address else '/64')).decode('ascii'), strict=False)
+	net = str(net)
+	
+	if not net in bans or time.time() - bans[net]['last_attempt'] > BAN_TIME:
+		bans[net] = { 'attempts': 0 }
+	
+	bans[net]['attempts'] += 1
+	bans[net]['last_attempt'] = time.time()
+	
+	if bans[net]['attempts'] >= MAX_ATTEMPTS:
+		print "Banning %s" % net
+		if type(ip) is ipaddress.IPv4Address:
+			subprocess.call(["iptables", "-I", "INPUT", "-s", net, "-j", "REJECT"])
+			subprocess.call(["iptables", "-I", "FORWARD", "-s", net, "-j", "REJECT"])
+		else:
+			subprocess.call(["ip6tables", "-I", "INPUT", "-s", net, "-j", "REJECT"])
+			subprocess.call(["ip6tables", "-I", "FORWARD", "-s", net, "-j", "REJECT"])
+	else:
+		print "%d more attempts until %s is banned" % (MAX_ATTEMPTS - bans[net]['attempts'], net)
+
+def unban(net):
+	print "Unbanning %s" % net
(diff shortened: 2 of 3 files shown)


andryyy:
Hi, that's very cool!
How is the cpu usage here with huge logs?

What do think about reading the logs from Redis? My guess is it would be a bit faster than the Docker API, but I'm not sure.

BAN_TIME and MAX_ATTEMPTS read from Redis with default values if the key is not found would be nice, too. I will implement these options to the UI then. :)

mkuron:
> How is the cpu usage here with huge logs?

As it's going through the Docker API, the absolute size of the logs doesn't matter, only the number of log entries per second. I haven't done any load testing, but I don't think it's a big deal compared to the CPU load caused by the services themselves. It definitely won't cause much more load than a `docker logs --tail 0 -f`.

> What do think about reading the logs from Redis? My guess is it would be a bit faster than the Docker API, but I'm not sure.

I think Docker log streaming doesn't even hit the log file, but is dealt with by the stdout/stderr handlers in the Docker daemon. So Redis should even be slower. Also, reading from Redis requires periodic polling, while reading through the Docker API can be done with the blocking generator as I have implemented it.

> BAN_TIME and MAX_ATTEMPTS read from Redis with default values if the key is not found would be nice, too. I will implement these options to the UI then. :)

That would be nice. I think many installations may want to use a larger MAX_ATTEMPTS than my default value.

andryyy:
> As it's going through the Docker API, the absolute size of the logs doesn't matter, only the number of log entries per second. I haven't done any load testing, but I don't think it's a big deal compared to the CPU load caused by the services themselves. It definitely won't cause much more load than a docker logs --tail 0 -f.

I see, that's very good. 

I will add these options to dev web UI. If you want further options, let me know.

lenovouser:
@mkuron where do the `iptables` commands get executed? On the host system or in each of the Postfix, Dovecot and SOGo containers?

andryyy:
On the host system. Well, it is executed in the container, but since these are containers and the container is privileged, it can use the hosts netfilter.

lenovouser:
That won't work 😞  - Docker is basically bypassing all `iptables` rules. It's pretty annoying and they still haven't implemented anything to fix it. If you disable it you break a s**t load of stuff - so that isn't going to be an option either.

https://github.com/moby/moby/issues/29312

andryyy:
I don't see why it wouldn't work when inserting these rules in the forward chain. It works for a local fail2ban installation. But I might be missing something. Writing from mobile.

mkuron:
> That won't work

No, that works perfectly fine, I did extensive testing.

> Docker is basically bypassing all iptables rules.

Docker doesn't do anything weird about iptables. It simply adds redirect rules to the `FORWARD` table, and my script adds rejection rules to the **beginning** of that table. So they get evaluated before any Docker rules are evaluated.

andryyy:
Does this work with a logging driver other than json?

mkuron:
I haven't tried. If `docker logs` works, then this works. https://docs.docker.com/engine/reference/commandline/logs/ says it works only with the json-file and journald backend. I assume you were hoping to use the redis or syslog backend, which would unfortunately break it.

In general, we should aim for preserving a working `docker logs` command as anyone who has ever used Docker before will try running that and be really confused if they don't get any output.

andryyy:
Yes, I also just yesterday added log_opts which would not work with drivers other than json. But I need to remove the old f2b method from the docs.

mkuron:
Ok. Yes, the old documented method is obsolete since my dockerized fail2ban clone was merged.

PhoenixPeca:
@mkuron, since this blocking feature modifies the host machine's firewall, I think it would be nice to have an option to disable this feature because there might been some cases where a non-mailcow service/app (ex Nextcloud, HumHub, OwnCloud, Ghost) could be blocked from being accessed by an entire ISP IP pool. IMHO, It would be unwise to manually remove a blocked IP pool or wait until 30 minutes has passed every time this happens. **And, it is also an inconvenience to the legitimate user if the IP address he/she is using is in the IP pool where the hacker made the brute force attack.**

.. it would also be convenient to have this option in the mailcow.conf file.

mkuron:
@PhoenixPeca, just comment out the lines in docker-compose.yml. Though you really shouldn't leave your Mailserver unprotected.

Unfortunately, this cannot be made an option in mailcow.conf because docker-compose does not have any concept of "optional" containers.

I don't think the scenario you describe is very likely. Your legitimate users will almost never share a network with an attacker. And if they ever do and repeatedly run into being blocked, you should contact the ISP's anti-abuse department and ask them to terminate the attacker's service for abuse of service and violation of contract.

mkuron:
Last night @andryyy added b9ffcf2 to dev, which implements a whitelist for fail2ban. Perhaps this can help you here, @PhoenixPeca?

PhoenixPeca:
thats great! ill take a look into it.

andryyy:
I've just added a var SKIP_FAIL2BAN for this. Just need to add it to the docs somewhere. 

Many thanks for the responsive fix! Never thought of scroll bars... 👍

PR #313, opened 2017-05-25
Authentication rate limiting for Mailcow UI

mkuron:
I had to turn on error logging for the PHP-FPM container to be able to log messages from the Mailcow UI. Also, I noticed that IPv4-mapped IPv6 addresses (like ::ffff:192.0.2.1) weren't dealt with correctly, but that is fixed now.

Follow-up to #311.

### DIFF
--- a/data/Dockerfiles/fail2ban/logwatch.py
+++ b/data/Dockerfiles/fail2ban/logwatch.py
@@ -13,6 +13,7 @@ RULES = {
 	'mailcowdockerized_postfix-mailcow_1': 'warning: .*\[([0-9a-f\.:]+)\]: SASL .* authentication failed',
 	'mailcowdockerized_dovecot-mailcow_1': '-login: Disconnected \(auth failed, .*\): user=.*, method=.*, rip=([0-9a-f\.:]+),',
 	'mailcowdockerized_sogo-mailcow_1': 'SOGo.* Login from \'([0-9a-f\.:]+)\' for user .* might not have worked',
+	'mailcowdockerized_php-fpm-mailcow_1': 'Mailcow UI: Invalid password for .* by ([0-9a-f\.:]+)',
 }
 BAN_TIME = 1800
 MAX_ATTEMPTS = 10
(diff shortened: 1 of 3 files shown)


andryyy:
Hey, Can you please just use error_report? Afair it doesn't depend on error reporting levels and is therefor slightly more flexible. But I'm trying to implement a Redis log sometime soon. 

Is remote addr always the clients address? With and without reverse proxy?

Beste Grüße 
André

mkuron:
> Hey, Can you please just use error_report? Afair it doesn't depend on error reporting levels and is therefor slightly more flexible.

Ah, thanks. I wasn't aware of `error_log` over `trigger_error`. I've changed it now.

> But I'm trying to implement a Redis log sometime soon.

I'm not sure that that would be better, see https://github.com/mailcow/mailcow-dockerized/pull/311#issuecomment-303975502. It would require polling instead of streaming and thus probably create more load. Also, you'd need to deal with things like log rotation yourself. Right now that is dealt with by the docker daemon automatically.

> Is remote addr always the clients address? With and without reverse proxy?

Yes, it is. That's what [`set_real_ip_from`](https://github.com/mailcow/mailcow-dockerized/blob/edc41b4/data/conf/nginx/site.conf#L36) is for.

andryyy:
Okay, cool. So if we implement a logging to web UI, we can just leave that error_log stuff there to be able to stream the logs.

andryyy:
You declared mailcowdockerized as project name, though the name depends on the name of the current directory. I added this to .env now as fixed name, I will also add this to the docs. 

Changes to the PHP container were not reverted, removed them now.

PhoenixPeca:
Related to #311
---

@mkuron, since this blocking feature modifies the host machine's firewall, I think it would be nice to have an option to disable this feature because there might been some cases where a non-mailcow service/app (ex Nextcloud, HumHub, OwnCloud, Ghost) could be blocked from being accessed by an entire ISP IP pool. IMHO, It would be unwise to manually remove a blocked IP pool or wait until 30 minutes has passed every time this happens. And, it is also an inconvenience to the legitimate user if the IP address he/she is using is in the IP pool where the hacker made the brute force attack.

.. it would also be convenient to have this option in the `mailcow.conf` file.

PR #318, opened 2017-05-28
Clamav dockerfile simplification

TBK:
- Locked to specific Alpine release.
- Fewer layers.
- Size reduction of 49 MB.
image


andryyy:
Thanks, I will add some slight changes to fit the layout (tab indent etc.) to fit @mkuron s changes.

PR #319, opened 2017-05-28
php-fpm dockerfile simplification

TBK:
Locked to specific Alpine release.
Fewer layers.
Size reduction of 403.3 MB.
image

I also checked the docker-entrypoint.sh script againtst http://www.shellcheck.net/ and corected most of the warnings. There are still a few left.

The last part of the dockerfile, from "# Configuring php-fpm" and below is a modified version of the last section of the official php image dockerfile - https://github.com/docker-library/php/blob/76a1c5ca161f1ed6aafb2c2d26f83ec17360bc68/7.1/fpm/alpine/Dockerfile

### DIFF
--- a/data/Dockerfiles/php-fpm/Dockerfile
+++ b/data/Dockerfiles/php-fpm/Dockerfile
@@ -1,17 +1,52 @@
-FROM php:7.1-fpm-alpine
+FROM alpine:3.6
+
 LABEL maintainer "Andre Peters <andre.peters@servercow.de>"
 
-RUN apk add -U --no-cache libxml2-dev icu-dev icu-libs redis mysql-client bash autoconf g++ make openssl
-RUN pecl install redis && pecl clear-cache
-RUN docker-php-ext-configure intl
-RUN docker-php-ext-install intl pdo pdo_mysql xmlrpc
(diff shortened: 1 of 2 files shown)


andryyy:
Thanks, but you shouldn't change the script based on a web check. Things like `rm /data/dkim/{keys,txt}/"${file}"` happen then. I agree that using quotation marks is mostly a good idea, but not always necessary. Just like `-r` for read when we read domain names.
Sometimes you lose functionality of a script with these slight changes. :)

I should nevertheless start using more quotation marks...

It's a very nice approach to save a lot of disk space, thanks!

andryyy:
And a +1 for  `Time to milk the cows`. 🐄

andryyy:
I see this is missing mysqladmin and redis-cli. I assume you just don't use the migration script. But then don't push untested, please.

Edit: It is also missing the PEAR path in include_path, ctype_alnum is not defined etc.
Sorry, this cannot be tested...

TBK:
I am sorry. Apprently I was too tired when doing tesing and tested against the old Dockerfile..... fail :(

I will create a new PR without the changed to the bash script and with all of the dependencies.

PR #328, opened 2017-06-03
support letsencrypt via acme-client

bebehei:
Hi all. I'm currently converting my setup from uberspace to a self-hosted server and this is the perfect project. Thank you for this awesome project!

So I thought, there could be a container added to support letsencrypt by default.

It's not fully tested yet, but I'd like to know what you guys think in general about this!

Things to solve:

- [x] regenerate `dhparam.pem` if it's the fakeoil one
- [x] trigger restart of nginx, dovecot and postfix (any service missing?)
- [x] initial account regeneration

Also another question for understanding: What does `ca.pem` do? I don't see, why there is a CA needed and also there is no reference to `ca.pem` in the code.


chriscroome:
The `ca.pem` file would be the [Let's Encrypt intermediate cert](https://letsencrypt.org/certificates/#intermediate-certificates).

Personally I have been using [acme.sh](https://github.com/Neilpang/acme.sh) as a Let's Encrypt client, after using the EFF Python client for some time, I haven't tried [dehydrated](https://github.com/lukas2511/dehydrated).

data/Dockerfiles/letsencrypt/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM alpine
```
TBK: It is best practice to lock base image version - e.g. alpine:3.6

data/Dockerfiles/letsencrypt/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+RUN apk add --update \
+    bash \
+    curl \
+    openssl \
+    && rm -rf /var/cache/apk/*
```
TBK: Use the --no-cache parameter instead of `&& rm -rf /var/cache/apk/*`

data/Dockerfiles/letsencrypt/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+RUN curl -L https://github.com/lukas2511/dehydrated/archive/master.tar.gz \
```
TBK: Using multiple RUNs adds additional layers, by using && the work is done in one.
`&& curl -L https://github.com/lukas2511/dehydrated/archive/master.tar.gz`

It is also prefered to lock the external depencendy to a specific version:
https://github.com/lukas2511/dehydrated/archive/v0.4.0.zip

The container philosophy is to create an immutable application.

An alternative could also be to use one of the ACME clients that have already been packaged for Alpine:
https://pkgs.alpinelinux.org/package/v3.6/community/x86_64/acme-client (in C)
https://pkgs.alpinelinux.org/package/edge/community/x86_64/lego (in GO)

data/Dockerfiles/letsencrypt/root/etc/dehydrated/config
```diff
@@ -0,0 +1,7 @@
+CONTACT_EMAIL="${LE_CONTACT_EMAIL:-$CONTACT_EMAIL}"
```
TBK: This should just be placed in `data/Dockerfiles/letsencrypt/` - ` data/Dockerfiles/letsencrypt/config`

data/Dockerfiles/letsencrypt/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+  | tar -xzf -
+
+ADD root/ /
```
TBK: `COPY` is prefered over `ADD`.
https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#add-or-copy

data/Dockerfiles/letsencrypt/root/srv/docker.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
TBK: This should just be placed in `data/Dockerfiles/letsencrypt/`. 
With the `COPY `command you can decide where it is copied to.

data/Dockerfiles/letsencrypt/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM alpine
+
```
TBK: Maintainer should be added:
`LABEL maintainer "Andre Peters <andre.peters@servercow.de>"`

bebehei: Question for understanding: Is `LABEL maintainer` equivalent to `MAINTAINER`?

TBK: It is the replacement, MAINTAINER has been deprecated.

docker-compose.yml
```diff
@@ -274,6 +274,31 @@ services:
...
         - /lib/modules:/lib/modules:ro
 
+    acme-mailcow:
+      depends_on:
+        - nginx-mailcow
+      image: mailcow/acme
+      build: ./data/Dockerfiles/acme
+      # All domains to be included in the certificate
+      command: ${MAILCOW_HOSTNAME}
+      #autodiscover.${MAILCOW_HOSTNAME} autoconfig.${MAILCOW_HOSTNAME}
```
bebehei: is https for the autodiscover and autoconfig domain needed?

docker-compose.yml
```diff
@@ -274,6 +274,31 @@ services:
...
+      environment:
+        - CONTAINERS_RESTART=mailcowdockerized_postfix-mailcow_1 mailcowdockerized_dovecot-mailcow_1 mailcowdockerized_nginx-mailcow_1
```
bebehei: Hardcoding `mailcowdockerized` is bad, but neccessary in here. How can we change this?

docker-compose.yml
```diff
@@ -274,6 +274,31 @@ services:
...
+      volumes:
+        - ./data/web/.well-known/acme-challenge:/var/www/acme:rw
+        - ./data/assets/ssl:/var/lib/acme/:rw
+        - /var/run/docker.sock:/var/run/docker.sock:ro
+      dns:
+        - 172.22.1.254
+      dns_search: mailcow-network
+      # TODO: find some way to use always
+      # do not restart the container after exit. Things get worse when we hit let's encrypt's ratelimit.
+      restart: "no"
```
bebehei: Is there any better restart-policy than `"no"`? Using `always` will result in hitting the ratelimit. This is much worse than a stopped container. If hit, the ratelimit cannot get reset manually and requires hours of time to wait. A stopped container can get started and the problem is fixed in seconds.

bebehei:
So, I've taken @TBK0000's comments into account and have rewritten it with `acme-client`. It works now out of the box, but some small questions are still open. Could someone of the dev team @mailcow/contributors review it?

docker-compose.yml
TBK: Could ${COMPOSE_PROJECT_NAME} from the mailcow.conf file be used?

docker-compose.yml
TBK: There is also `on-failure:`

image

https://docs.docker.com/engine/reference/run/#restart-policies-restart

docker-compose.yml
TBK: According to this reddit thread it has to be TLS encrypted.
https://www.reddit.com/r/exchangeserver/comments/3qm2l8/autodiscoverdomaincom_is_getting_ssl_certificate/

And this MS KB:
https://support.microsoft.com/da-dk/help/2772058/-the-name-on-the-security-certificate-is-invalid-or-does-not-match-the-name-of-the-site-error-in-outlook-in-a-dedicated-or-itar-office-365-environment

docker-compose.yml
bebehei: Yeah, I know 'bout that. actually `on-failure` is the bad one.

docker-compose.yml
bebehei: I never heard such a variable `${COMPOSE_PROJECT_NAME}`. Where should this be defined? Nothing in `mailcow.conf`.

TBK: https://github.com/mailcow/mailcow-dockerized/blob/dev/generate_config.sh

bebehei: I see. But I'm merging against `master` 😕

docker-compose.yml
bebehei: Well, reading the docs again:

> If a container is successfully restarted (the container is started and runs for at least 10 seconds), the delay is reset to its default value of 100 ms.

I always had in my mind, this was at 1s instead of 10s. acme-client should fail in less than 10s. So this is safe to use with `on-failure:3` and not hit rate limits.

Thanks.

data/Dockerfiles/acme/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM alpine:3.6
+
+LABEL maintainer "Andre Peters <andre.peters@servercow.de>"
+
+RUN apk add --update --no-cache \
+      acme-client \
+      docker \
```
TBK: I was wondering if wouldn't be possible to achieve the same functionality with API calls (curl) then installing docker wouldn't be necessary - https://docs.docker.com/engine/api/get-started/

bebehei: Yes. docker sucks up 90MB of 100MB in the disk image. I'm aware of that problem. But I just have not known yet another api. Let's have a look. Thanks.

TBK: Here is the API docs - https://docs.docker.com/engine/api/v1.29/#operation/ContainerRestart

bebehei: Well, that's awesome! Thanks for reminding me. curl works like a charm!

data/Dockerfiles/acme/docker-entrypoint.sh
```diff
@@ -18,18 +18,14 @@ while true; do
...
 
 	case "$?" in
 		0) # new certs
 			# cp the new certificates and keys 
 			cp ${ACME_BASE}/acme/fullchain.pem ${ACME_BASE}/cert.pem
 			cp ${ACME_BASE}/acme/private/privkey.pem ${ACME_BASE}/key.pem
 
-			# also generate new dhparams with new certs
-			#(on the first run, the old dhparams may be the fakeoil from git)
-			openssl dhparam -out ${ACME_BASE}/dhparams.pem 2048
```
bebehei: @andryyy Could you please elaborate this?

In contrast to `key.pem`, `dhparams.pem` can be public. So it's ok to use the ones from the git-repository. But is it desirable to use the ones from the git-repository?

I see this PR actually to enable SSL on a zero effort basis. So I don't expect anyone doing manual work for SSL while using this container.

So what do you think: Is it desirable to have every mailcow user running with the same dhparams?

docker-compose.yml
```diff
@@ -280,9 +280,10 @@ services:
       image: mailcow/acme
       build: ./data/Dockerfiles/acme
       # All domains to be included in the certificate
-      command: ${MAILCOW_HOSTNAME} autodiscover.${MAILCOW_HOSTNAME} autoconfig.${MAILCOW_HOSTNAME}
       environment:
         - CONTAINERS_RESTART=mailcowdockerized_postfix-mailcow_1 mailcowdockerized_dovecot-mailcow_1 mailcowdockerized_nginx-mailcow_1
+        # Additional subject alternate names
+        - ADDITIONAL_SAN=
```
bebehei: What about moving these into mailcow.conf?

data/Dockerfiles/acme/docker-entrypoint.sh
bebehei: @andryyy I've seen the explanation in the commit message about dhparams. Sorry for the noise.

andryyy: No problem! :)

docker-compose.yml
andryyy: Yes, good idea.

andryyy:
It is missing any reference to nginx-mailcow to actually respond  to the challenge. :o 

//Edit: Forget it. Sorry. :)

Wow, starting and stopping the services really makes you hit the rate limit very fast.

bebehei:
Testing at first with the `-s` option in acme-client will give you more time. During development of this PR, I've also hit the RL quite often.

Also I have to say, it works like a charm on my machine.

data/Dockerfiles/acme/docker-entrypoint.sh
bebehei: But... why does dovecot regenerate `dhparams.pem` while it has snakeoil certs? So if dhparams get regenerated, but not the certs, ssl will still be broken (dovecot, nginx **and postfix**). What's the point on regenerating dhparams without knowing if SSL termination will be successful?

I can hear mailbox.org loudly crying about the majority of mailservers having only broken SSL.

data/Dockerfiles/acme/docker-entrypoint.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+
+ACME_BASE=/var/lib/acme
+mkdir -p ${ACME_BASE}/acme/private
+
+restart_containers(){
+	for container in $*; do
+		curl -X POST \
+			--unix-socket /var/run/docker.sock \
+			"http:/containers/${container}/restart"
```
Dragnucs, line 10: Shouldn't this strart with http://containers?

bebehei: Well, this was my first thought too. But it's consistently written in the [docker documentation](https://docs.docker.com/engine/api/get-started).

bebehei: Well, I'm currently thinking about this. But I'm sure, that the `containers`-bit is part of the filename in the URL and not some server name. And a possible explanation is that the `//example.com` in an URL is an optional hostname. Because we're using `--unix-socket` to connect to docker, there is no hostname needed and therefore we're using the "default" hostname.

If this is true. That 'd be awesome.

bebehei:
@andryyy What's the status of this PR? Any info when you will merge it?

data/Dockerfiles/acme/docker-entrypoint.sh
RealOrangeOne: I believe you need a double slash in the protocol? `http://`

andryyy: It works like this. :) I thought it was incorrect, too.

andryyy:
@bebehei Waiting for Rspamds 1.6.0 release (will be released within the next hours). Then we can finally drop Rmilter, merge the UI changes and have ARC support in mailcow! :)
Thanks to @vstakhov for his awesome work!

bebehei:
> ARC

What's this?

TBK:
Rspamd 1.6.0 has been release.

TBK:
@bebehei https://rspamd.com/doc/modules/arc.html

mkuron:
Looks like @andryyy already pushed part of this PR to the dev branch.

data/Dockerfiles/acme/docker-entrypoint.sh
```diff
@@ -0,0 +1,43 @@
...
+}
+
+while true; do
+
+	acme-client \
+		-v -b -N -n \
+		-f ${ACME_BASE}/acme/private/account.key \
+		-k ${ACME_BASE}/acme/private/privkey.pem \
+		-c ${ACME_BASE}/acme \
+		${MAILCOW_HOSTNAME} autoconfig.${MAILCOW_HOSTNAME#*.} autoconfig.${MAILCOW_HOSTNAME#*.} ${ADDITIONAL_SAN}
```
lazyfrosch, line 21: It would be nice to be able to configure the full list. So you don't have to use `autoconfig`

Also the list ist wrong, it should be `autodiscover.${MAILCOW_HOSTNAME#*.} autoconfig.${MAILCOW_HOSTNAME#*.}` --> `discover` :wink:

data/Dockerfiles/acme/docker-entrypoint.sh
lazyfrosch: Wild guess: http URL without a hostname (since its a unix socket)

andryyy:
I added some more code and am testing it. Looks good so far!

data/Dockerfiles/acme/docker-entrypoint.sh
andryyy: It is autoconfig for Thunderbird and autodiscover for Outlook. :)

lazyfrosch: Yes, but you have listed `autoconfig` twice.

Also autodiscover can be done via an SRV record (what I want to test and would prefer - since it would work better for multiple domains)

Have to test this for autoconfig, but I bet there is SRV support as well...

andryyy: Oh. Didn't notice this. The dev differs a bit though: https://github.com/mailcow/mailcow-dockerized/blob/dev/data/Dockerfiles/acme/docker-entrypoint.sh

This code checks for autodiscover/autoconfig.{each_domain} and tries to create a certificate.

Yes, a SRV works fine, too. Should be part of the docs.

mkuron:
Can this be closed? The version that's already in master works quite nicely.

skupfer:
Before opening an issue for this question I would like to use this PR...
The guide states mailcow must be running on port 80... Therefore one cannot use this running mailcow on 8080/8443/reverse proxy, right? Could this be achieved in the future?

andryyy:
It should work when the RP is available on port 80.

PR #331, opened 2017-06-04
php-fpm dockerfile simplification take 2

TBK:
This setup is tuned for production.

I have tested it on two machines several times. Think I got all the dependencies, but more testing is appreciated.

### DIFF
--- a/data/Dockerfiles/php-fpm/Dockerfile
+++ b/data/Dockerfiles/php-fpm/Dockerfile
@@ -15,25 +15,25 @@ RUN set -x \
 && apk add --no-cache \
   bash \
   curl \
-  redis \
-  mariadb-client \
   libressl \
+  mariadb-client \
+  redis \
   php7-fpm \
   php7-common \
+  php7-ctype \
+  php7-curl \
   php7-intl \
+  php7-json \
+  php7-openssl \
   php7-pdo \
   php7-pdo_mysql \
-  php7-xmlrpc \
-  php7-redis \
-  php7-openssl \
-  php7-json \
-  php7-curl \
-  php7-ctype \
   php7-pear \
   php7-pear-auth_sasl \
+  php7-pear-mail_mime \
   php7-pear-net_smtp \
   php7-pear-net_idna2 \
-  php7-pear-mail_mime \
+  php7-redis \
+  php7-xmlrpc \
 && pear install Net_IMAP \
 # MISSING apk for php7-pear-net_imap - can be installed once https://github.com/alpinelinux/aports/pull/1359 is merged.
 \
--- a/data/Dockerfiles/php-fpm/Dockerfile
+++ b/data/Dockerfiles/php-fpm/Dockerfile
@@ -11,7 +11,7 @@ RUN set -x \
 && adduser -u 82 -D -S -G www-data www-data \
 \
 # Install Dependencies 
-&& apk add --update \
+&& apk update \
 && apk add --no-cache \
   bash \
   curl \
--- a/data/Dockerfiles/php-fpm/Dockerfile
+++ b/data/Dockerfiles/php-fpm/Dockerfile
@@ -12,9 +12,28 @@ RUN set -x \
 \
 # Install Dependencies 
 && apk add --update \
-&& apk add --no-cache bash curl redis mariadb-client libressl php7-fpm \
-php7-common php7-intl php7-pdo php7-pdo_mysql php7-xmlrpc php7-redis php7-openssl php7-json php7-curl php7-ctype \
-php7-pear php7-pear-auth_sasl php7-pear-net_smtp php7-pear-net_idna2 php7-pear-mail_mime \
+&& apk add --no-cache \
+  bash \
+  curl \
+  redis \
+  mariadb-client \
+  libressl \
+  php7-fpm \
+  php7-common \
+  php7-intl \
+  php7-pdo \
+  php7-pdo_mysql \
+  php7-xmlrpc \
+  php7-redis \
+  php7-openssl \
+  php7-json \
+  php7-curl \
+  php7-ctype \
+  php7-pear \
+  php7-pear-auth_sasl \
+  php7-pear-net_smtp \
+  php7-pear-net_idna2 \
+  php7-pear-mail_mime \
 && pear install Net_IMAP \
 # MISSING apk for php7-pear-net_imap - can be installed once https://github.com/alpinelinux/aports/pull/1359 is merged.
 \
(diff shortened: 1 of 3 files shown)


data/Dockerfiles/php-fpm/Dockerfile
```diff
@@ -1,17 +1,62 @@
...
 
-COPY ./docker-entrypoint.sh /
+# Add group + user - 82 is the standard uid/gid for "www-data" in Alpine
+RUN set -x \
+&& addgroup -g 82 -S www-data \
+&& adduser -u 82 -D -S -G www-data www-data \
+\
+# Install Dependencies 
+&& apk add --update \
+&& apk add --no-cache bash curl redis mariadb-client libressl php7-fpm \
```
bebehei: Docker's best practices is to split up packages onto multiple lines and sort it. See [Dockerfile best practices](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#sort-multi-line-arguments)

TBK: Done :)

data/Dockerfiles/php-fpm/Dockerfile
```diff
@@ -12,9 +12,28 @@ RUN set -x \
...
+  php7-redis \
+  php7-openssl \
+  php7-json \
+  php7-curl \
+  php7-ctype \
+  php7-pear \
+  php7-pear-auth_sasl \
+  php7-pear-net_smtp \
+  php7-pear-net_idna2 \
+  php7-pear-mail_mime \
```
bebehei: Well, I have to emphasize the "sorted". Sorry for bugging you with this, but only then the rule makes sense.

TBK: I have sorted them by package and subpackages in an alphabetical manner.

andryyy:
I don't think you should introduce a global gitattriubtes file with this PR. 

Also I don't think it is best practice to force all commands in one RUN line. We also lose the ability to switch between PHP versions. 

@mkuron What's your take?

mkuron:
If you want to add a gitattributes file, it should cover all file types we use (.py, .php, .lua, ...). It's not very useful with just one.

I think we should continue building upon the official PHP image and not try to install the packages in Alpine. That's more flexible for installing PHP modules and has versioning. So I'm also not in favor of this PR. I haven't checked, but it probably doesn't reduce build time by that much either -- PHP modules didn't take very long to compile.

The one thing in this PR that does make sense to me is merging multiple `RUN`s to reduce the number of layers. This is recommended by Dockerfile best practices, but shouldn't be taken to the extreme where legibility suffers. Just combine everything into a few `RUN`s with related commands.

TBK:
I added the gitattributes file for shell files because it effects the Docker images build on Windows, it can be removed, expanded or turned into another PR.

It is smaller compared to the current "Official PHP image" (made my Docker folks) based image (uncompressed a lot, compressed a bit).
The official PHP image is setup for development, not production.
Space saving is always a plus in my book, since storage cost money.

Compilation takes a lot longer then just downloading and unpacking of packages.
I made a VM (Hyper-V) one CPU, 1024 MB RAM, 30 GiB disk (stored on SSD), 100/100 Mb/s connection to test it. I measured the command execution time with this script:
```
#!/bin/sh
ts=$(date +%s%N);
"$@";
tt=$((($(date +%s%N) - $ts)/1000000));
echo "Time taken: $tt milliseconds"
```

See the "Time used: xxxxxxxxxx milliseconds" line.

Current dockerfile clean build:
image

Current dockerfile with base image downloaded:
image


New dockerfile clean build:
image

New dockerfile with base image downloaded:
image


If we compare the builds where the base image was not downloaded, then it is roughtly 129 seconds for the current and 16 seconds for the new dockerfile.

My primary reason for making an all Alpine repository based version it is make it easy to create an ARM version of MailCow, the only thing required is to change the base image (https://github.com/docker-library/official-images#architectures-other-than-amd64)

Since this is a production image and it is pretty static, there is no need for several RUN commands.

Several RUNs can be used when a part of the image is often iterated upon and you want to use the docker build cache mechanism to build the new image faster.

mkuron:
One big advantage of building on top of official images for standard software like PHP is that they will always be well-tested and contain settings that resemble the project's current best practices. They are also likely to get updates to new versions faster and we can decide in our own time whether we want the 5.6 branch, the 7 branch, the 7.0 branch, or the 7.1 branch.

> It is smaller compared to the current "Official PHP image" (made my Docker folks) based image (uncompressed a lot, compressed a bit).

Where does that difference come from? Unstripped binaries, static linking, PEAR cache, anything like that? In the end, both should contain a pretty standard install of PHP. Perhaps some size reductions can be proposed to the PHP official image.

> If we compare the builds where the base image was not downloaded, then it is roughtly 129 seconds for the current and 16 seconds for the new dockerfile.

That's surprisingly much. All we do is compile a few PHP extensions and PEAR modules. Any more detailed insights?

> My primary reason for making an all Alpine repository based version it is make it easy to create an ARM version of MailCow, the only thing required is to change the base image

This could also be solved by convincing the PHP maintainers to build official images for the most common other architectures. Furthermore, PHP isn't the only official image Mailcow uses; others include Bind9, MariaDB, Redis and Memcached. We wouldn't want to have to build and maintain all of them ourselves.

> Since this is a production image and it is pretty static, there is no need for several RUN commands.

Agreed, but this can easily be accomplished in the existing Dockerfile. If all of the `RUN` commands are merged, things like GCC and Autoconf would never end up in the image, saving space.

TBK:
Alpine and also the PHP packages gets used a lot and are frequently updated (on edge).
If you want php5 then it is simply to replace the 7 with 5 (https://pkgs.alpinelinux.org/packages?name=php5*&branch=v3.6&repo=&arch=&maintainer=), if you want 7.0 then change the base image to 3.5.

I have not investigated the "bloating" any further.

Example of the stragegy to taken with Alpine/dockerfile when compiling stuff to minimize cleanup:
```
&& apk add --no-cache python3 ffmpeg opus ca-certificates \
&& apk add --no-cache --virtual .build-deps git python3-dev libffi-dev libsodium-dev musl-dev gcc make \
\
# Install pip dependencies
&& pip3 install --no-cache-dir -r requirements.txt \
\
# Clean up build dependencies
&& apk del .build-deps
```

resystit/bind9 is not an official image. I have made several PRs to it so it currently is based on Alpine 3.6 (x86_64). It is my intention to talk with the dev about making images for different architecture.

The official MariaDB image is pretty big, I have not looked at reducing the size as of yet. My guess is it should be easy to shave off a couple of hundred MBs.

Redis and Memcached are available in official ARM varieties (https://hub.docker.com/r/arm64v8 & https://hub.docker.com/u/arm32v7/). Currently the Alpine flavors are missing since there is a problem with the maintainers script, should be fixed soon.

Issue #326, opened 2017-06-02
Running Mailcow Dockerized on Docker for Windows?


jaimbo:
Hi, I currently have the standalone version of Mailcow running in a VM on a Windows box, but seeing as how it hasn't had any updates for a while and you recommend moving to the Docker version. Is it possible to run the Dockerized version inside Docker for Windows?

mkuron:
Docker for Windows in the end just runs a Linux kernel in a HyperV VM, so I don't think it would be much different than on Linux. Mailcow is fairly standard in terms of how it interacts with the Docker platform; the only exception is the ipv6nat container (and the fail2ban container in the dev branch) which directly sets iptables rules on the host -- but you can disable those for a start.

If Mailcow doesn't work on Windows for whatever reason, you could also install Docker inside a Linux VM.

jaimbo:
Perfect I'll give it a bash after work and I'll report back if there are any issues.

Do you think there will be any problems in performance? I'm sure I've seen the Docker Devs state in a few places that Docker for Windows is only meant for development and testing.

andryyy:
I have no idea. :-) I think you may also want to install the "Windows Subsystem for Linux" => https://msdn.microsoft.com/de-de/commandline/wsl/install_guide

Braintelligence:
I'm pretty sure docker for windows existed before the subsystem for Linux was released. Are you sure it always runs on a Linux kernel?

jaimbo:
Seems to be getting stuck on:

Creating mailcowdockerized_clamd-mailcow_1 ... done

mkuron:
The Windows Subsystem is not needed. It provides an API translation layer so that Linux binaries can be executed on top of the Windows kernel. I'm pretty sure that translation wouldn't support the necessary kernel APIs for containers.

Docker on Windows runs a lightweight Linux VM in the background if you're running Linux containers (there is a diagram somewhere in the documentation that shows that). Only if you're running Windows containers you wouldn't need that.

@jaimbo, can you provide any logs? I don't think anyone has ever tried that with Mailcow.

jaimbo:
@mkuron What kind of logs were you after?

I managed to get Mailcow up and running, even managed to get my databases backed up and transferred over from the old install. The problem above was caused by Docker for Windows not pulling the data from the mailcow.conf. So I instead declared all the vars in a .bat and then ran the commands and that sorted it.

[...]

PR #332, opened 2017-06-04
Replace symlink to PHP script

mkuron:
There is one symlink in Mailcow (besides the .env => mailcow.conf one) that might cause trouble on Docker for Windows, see #326.

This PR replaces it with `require_once` inside the PHP script.


andryyy:
Great, thanks!

PR #347, opened 2017-06-11
Add Tooltip to mailcow Apps

PhoenixPeca:
Basically, this allows the admin to set an optional custom description of a specific mailcow app that can be read on mouse hover. Additionally, I have also added `htmlspecialchars` to escape double quotes in some html attributes. The descriptions are in english by default, but users can always customize them in `vars.local.inc.php` so I don't think we'll need a dedicated translation entries in the `lang` folder.
sdgdfg

### DIFF
--- a/data/web/inc/vars.inc.php
+++ b/data/web/inc/vars.inc.php
@@ -42,11 +42,13 @@ $PASSWD_REGEP = '.{4,}';
 $MAILCOW_APPS = array(
   array(
     'name' => 'SOGo',
-    'link' => '/SOGo/'
+    'link' => '/SOGo/',
+    'description' => 'SOGo is a collaborative software (groupware) server.'
   ),
   // array(
     // 'name' => 'Roundcube',
-    // 'link' => '/rc/'
+    // 'link' => '/rc/',
+    // 'description' => 'Roundcube is a web-based IMAP email client.',
   // ),
 );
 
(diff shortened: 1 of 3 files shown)


data/web/inc/vars.inc.php
```diff
@@ -42,11 +42,13 @@
 $MAILCOW_APPS = array(
   array(
     'name' => 'SOGo',
-    'link' => '/SOGo/'
+    'link' => '/SOGo/',
+    'description' => 'SOGo is a collaborative software (groupware) server.'
```
mkuron: Change the descriptions to something less technical, e.g. "Access your email, contacts and calendar" and "Access your email".

mkuron (changes requested):

data/web/inc/vars.inc.php
PhoenixPeca: @mkuron how about this one "SOGo is a multi-purpose groupware server that provides native Microsoft ActiveSync support."? 😄

mkuron: Remember, this page is used by the end user, not that much by the admin. They might never have even heard about "groupware" or "ActiveSync". Also, ActiveSync is not used in the browser through the link here, so it doesn't need to be mentioned in the description.

PhoenixPeca: Any suggestions? 😄

PhoenixPeca: How about this?

"SOGo is a web-based client for email, address book and calendar."
"Roundcube is a light-weight alternative of SOGo webmail."

mkuron: The first one is good. Change the second one to "Roundcube is a web-based email client" (i.e. drop the SOGo reference and don't claim it's lightweight -- it's a pretty massive piece of PHP and JS after all).

PR #354, opened 2017-06-12
Add missing backslash in php-fpm Dockerfile

mkuron:
Missed this in #352


andryyy:
Thanks, already added it but have not yet pushed it. :(

PR #356, opened 2017-06-13
Fix unbound root hint download

mkuron:


andryyy:
That was not a typo, just for the progress bar. :)

mkuron:
Interesting, never seen that before. It looked like a misplaced comment to me...

PR #358, opened 2017-06-13
Added the "Authentication-Results" header

v0ctor:
This header allows MUAs to understand if received emails have passed SPF, DKIM and DMARC checks.

This also works with Rspamd 1.6.0 (https://rspamd.com/announce/2017/06/12/rspamd-1.6.0.html and https://rspamd.com/doc/modules/milter_headers.html).

### DIFF
--- a/data/conf/rspamd/local.d/rmilter_headers.conf
+++ b/data/conf/rspamd/local.d/rmilter_headers.conf
@@ -1,0 +1,1 @@
+use = ["authentication-results"];


andryyy (commented):
I activated in dev. 

We had to make Rspamd merge milter replies for or own LUA scripts to work. It was a longer WIP.

It was also necessary to change/add extended_headers due to Rmilter removal.

Better open an issue beforehand. :)

v0ctor:
Oh, sorry :(

By the way, it would be a great idea to add the header to internal emails in addition to external ones.

PR #361, opened 2017-06-15
Properly gitignore the SSL certificate

mkuron:
@andryyy, I saw that in 5aff0a3d6dbbfb30f6115515a1b194343052e77d you tried to gitignore the example SSL certificate. That doesn't work because the files were force-added. gitignore does not apply to files that are already in the repository.

Here is a suggestion how to fix the problem: we move the example certificate to a different folder and have the generate_config.sh script copy it to the right location if there is no certificate. If there is already a (custom) certificate in the ssl folder, it is not overwritten.


andryyy:
Yes, thanks, I totally missed it and couldn't find my mitake. :)

PR #365, opened 2017-06-16
Include a customized SOGo integrator plugin for Thunderbird

mkuron:
SOGo provides the SOGo Integrator plugin for Thunderbird that makes it behave more like a proper SOGo client. For example, it syncs the list of subscribed IMAP folders, ensures that Thunderbird displays the same address books and calendars that are also visible in the SOGo web interface, and makes the Global Address List available. This saves you the effort of manually creating these subscriptions in Thunderbird.

SOGo integrator needs to be customized with the server name, the name of the Global Address List, and a list of additional plugins to be installed. I decided to install the SOGo Connector (which adds CardDAV support to address book and makes it possible to edit ACLs for contacts and calendars), the Sieve plugin (for editing filters) and the IMAP ACL plugin (for editing ACLs for mail folders). These plugins are all [recommended by SOGo](http://wiki.sogo.nu/Thunderbird).

With my changes, the plugins will automatically be customized whenever reconf-domains.sh is run. This script is already responsible for telling the SOGo server which domains are configured on Mailcow, so it seemed like the logical place to also customize the plugins.

All a user needs to do is configure his email account in Thunderbird and set it to be the default account, then download and install [https://_mailcow.example.com_/thunderbird-plugins/sogo-integrator-31.0.4-_example.com_.xpi](https://mailcow.example.com/thunderbird-plugins/sogo-integrator-31.0.4-example.com.xpi) (note that the file name contains the domain name as there is a separate customized plugin for each domain). Thunderbird will restart twice and everything will be set up correctly. I will soon provide documentation to explain all the steps for the user and put a link onto the account page.


mkuron:
It's all working now. Let me know what you think.

Note that there is some slight brokenness in SOGo Connector with recent Thunderbird versions, but I have already submitted a patch: https://github.com/inverse-inc/sogo-connector.tb31/pull/27. Until that's merged, steer clear of the "All Address Books" address book and don't try adding/editing contacts from within an email message.

andryyy:
Let me check it, looks very promising! Thanks!

Braintelligence:
Thanks I've read it. 
This is awesome. Please keep it up!

Mailcow's the best! :)

PR #370, opened 2017-06-18
Autodiscovery for eM Client and Outlook 2007/2010

mkuron:
[eM Client](http://www.emclient.com) is an alternative to Thunderbird and Outlook. It natively supports CalDAV and CardDAV. It can autodiscover mail servers through the same mechanism as Outlook. As it doesn't support EAS, we need to send it the IMAP configuration.

~~eM Client can only be identified by its lack of User Agent when performing autodiscovery. All other common clients correctly send a User Agent header, so we can use that as an identification criterion.~~
Autodiscovery clients include in their request whether they are mobile devices (iPhone, Windows Phone, etc.) or desktop clients (Outlook, eM Client) using a response scheme attribute. We now differentiate based on that to determine whether a client gets IMAP or EAS. For desktop clients, with this PR, we now only use EAS for Outlook 2013 or higher on Windows. All other versions only support IMAP.

This PR also adds extra `<Protocol>` sections for CalDAV and CardDAV, which are ignored by Microsoft Outlook but used by eM Client to automatically set up calendars and address books.

### DIFF
--- a/data/web/autodiscover.php
+++ b/data/web/autodiscover.php
@@ -35,7 +35,8 @@ if (strpos($data, 'autodiscover/outlook/responseschema')) { // desktop client
 	if ($config['useEASforOutlook'] == 'yes' &&
 	    strpos($_SERVER['HTTP_USER_AGENT'], 'Outlook') !== FALSE && // Outlook
 	    strpos($_SERVER['HTTP_USER_AGENT'], 'Windows NT') !== FALSE && // Windows
-	    preg_match('/Outlook (1[5-9]\.|[2-9]|1[0-9][0-9])/', $_SERVER['HTTP_USER_AGENT']) // Outlook 2013 (version 15) or higher
+	    preg_match('/Outlook (1[5-9]\.|[2-9]|1[0-9][0-9])/', $_SERVER['HTTP_USER_AGENT']) && // Outlook 2013 (version 15) or higher
+	    strpos($_SERVER['HTTP_USER_AGENT'], 'MS Connectivity Analyzer') === FALSE // https://testconnectivity.microsoft.com doesn't support EAS for Outlook
 	) {
 			$config['autodiscoverType'] = 'activesync';
 	}
--- a/data/web/autodiscover.php
+++ b/data/web/autodiscover.php
@@ -33,6 +33,10 @@ if ($config['useEASforOutlook'] == 'no') {
 	}
 }
 
+if (!isset($_SERVER['HTTP_USER_AGENT']) || empty($_SERVER['HTTP_USER_AGENT'])) { // eM Client sends no user agent
+	$config['autodiscoverType'] = 'imap';
+}
+
 $dsn = "$database_type:host=$database_host;dbname=$database_name";
 $opt = [
 		PDO::ATTR_ERRMODE            => PDO::ERRMODE_EXCEPTION,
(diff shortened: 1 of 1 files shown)


K2rool:
I tested it about hour and half ago using Outlook 2010 which never worked i was going to say your changes don't work correctly, but since then you did a other commit 08506f7 which i tested and it works perfectly.

mkuron:
Thanks for testing, @k2rool. I've also tested on a number of different clients and think this PR is ready for merging now. Further testing would be appreciated as I don't actually have access to older Outlook versions or non-iOS mobile platforms.

I'm going to submit another PR soon that adds documentation of which clients are supported and how to set them up.

andryyy:
Does Outlook 2010 EAS not work?

K2rool:
Not far as i know, i try to add it as exchange it's not even doing any web request.

Getting this error all the time

image

mkuron:
EAS was added in Outlook 2013. That's what the version >= 15 check in my patch is for, it ensures that Outlook 2007 and 2010 are set up with IMAP. Outlook versions older than that didn't have autodiscover and are long out of support anyway.

andryyy:
Strange. I really thought EAS was included in 2010.

Let me check this PR, but it looks good so far. Thanks!

andryyy:
Are you willing to add this to https://mailcow.github.io/mailcow-dockerized-docs/u_e-autodiscover_config/ ? :)

mkuron:
That page is actually still valid.

In a few days I'm going to provide step-by-step guides for all clients that support either EAS or IMAP+CalDAV+CardDAV. On last count, that included Outlook 2013/2016 for Windows, Thunderbird, eM Client, KMail/Kontact, macOS, iOS, Android, and Windows Phone. Outlook for Mac does not support EAS or CardDAV/CalDAV, while Evolution lacks CardDAV support.

I was thinking about putting these somewhere on the Mailcow user account page. The docs repository provides documentation to the server admin, and this is clearly an end-user topic. Also, from #365 we need to provide a domain-specific download link for SOGo Integrator for Thunderbird, so this link can't be in the docs repo.

andryyy:
That would be amazing, send me a PayPal.me link if you have any (or a Bitcoin address?) for a donation for this!

andryyy:
We still need to pass the correct port for the mail/web server. Don't assume it's always 443. Needs to be tested with and without RP.

mkuron:
> We still need to pass the correct port for the mail/web server. Don't assume it's always 443.

Actually, Outlook (and the other clients) always contact https://autodiscover.example.com/, i.e. always HTTPS, always port 443. If they can reach our autodiscovery script, we already know that Mailcow's Nginx owns port 443 (be it directly or via reverse proxy). So no need to do anything extra.

andryyy:
You can still set them up manually to use another port.

mkuron:
But not with autodiscovery.

andryyy:
Yes, sorry, I was talking about the EAS server.

We can try to use $_server[http_host], I think.

andryyy:
Oh well, that's stupid. If we reach autodiscover we are already on port 443. Forget it. :)

andryyy:
EAS is not selected for Outlook 2013 with this......

andryyy:
Outlook/15.0 (15.0.4927.1000; C2R; x86)
Microsoft Office/15.0 (Windows NT 10.0; MAPI 15.0.4927; Pro)

Needs to match both, I will add a fix later.

mkuron:
Ah, thanks for noticing. I did not test the Click-to-run version of Office, only the MSI version. What was Microsoft thinking there giving them two different user agents...

PR #372, opened 2017-06-19
data/web: Update syncjob password input to type password

lazyfrosch:
So the password is not readable in clear text.

### DIFF
--- a/data/web/edit.php
+++ b/data/web/edit.php
@@ -527,7 +527,7 @@ elseif (isset($_SESSION['mailcow_cc_role']) && ($_SESSION['mailcow_cc_role'] == 
 					<div class="form-group">
 						<label class="control-label col-sm-2" for="password1"><?=$lang['edit']['password'];?></label>
 						<div class="col-sm-10">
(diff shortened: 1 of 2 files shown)


andryyy:
Could you please push this to dev? :)

lazyfrosch:
Ooops, okay. Updated

Btw. you should merge master -> dev to avoid merge / rebase problems every time :)

andryyy:
master to dev? the dev stage was master at one point.

mkuron:
@lazyfrosch is right. The current problem is that bug fixes are directly put into master and then independently reimplemented on dev. This produces merge conflicts when changes are proposed against master and the base branch of the PR is then changed to dev.

The correct way to do bugfixes on master (and new features on dev) would be to merge master into dev every time a change is made on master.

andryyy:
I should not commit pushes to master. Problem would be solved.

lazyfrosch:
@andryyy well, there can be good reasons why. Just make sure to merge it to dev afterwards. Then dev is always prepared to be merged to master.

andryyy:
Hm, I once did it. But sometimes I am lazy. I should start doing it and also use better comments for commits!

lazyfrosch:
@andryyy no worries, I'm amazed by the stack so far :wink:

PR #373, opened 2017-06-19
dovecot/imapsync_cron: Use passfiles to hide passwords from ps

lazyfrosch:


andryyy:
Looks good! I have no idea: will the temp passwd files be removed?

lazyfrosch:
@andryyy Yes, `File::Temp` deletes the files on object destruction (when Perl exits in some controlled way)

andryyy:
Nice, let's merge it.

PR #389, opened 2017-06-22
Client configuration guides

mkuron:
This pull request adds client configuration guides. The user can see them when they log into the Mailcow web UI. I have included a step-by-step setup guide for all clients that support IMAP, CalDAV and CardDAV (or Exchange) and can be set up easily through autoconfiguration. Three popular client are missing: Outlook 2011 and 2016 for Mac support neither DAV nor Exchange (also, 2016 uses the MX record instead of autodiscover.xml to discover the mail server and 2011 has no autodiscovery at all) and Gnome Evolution has to be configured fully manually (specifying IMAP and SMTP addresses and CardDAV and CalDAV URLs). I have tested the guide for the latest version of each client.

For Android, CalDAV/CardDAV should also work, but it requires a third-party app. If anybody wants to provide such a guide, please do.

---------

This PR also includes some extra changes to the web UI that I made while creating the guides. These are:
- fix for a blank display name when Outlook autodiscovers the IMAP configuration
- a diagnostic page that shows which DNS records should be created for mailcow and whether they have been created correctly (if not, the current value is displayed, see screenshot below for an example)
- a generator for mobileconfig files for Apple iOS / macOS that is used by the client configuration guide
- support for non-standard port numbers in autoconfig/autodiscover/mobileconfig via SRV records (though they are only pulled from the first domain, all other domains are assumed to have the same records). If a SRV record target is set to `.`, that service is removed from the generated XML.

dnsrecords

---------

I am aware that the HTML layout of the configuration guides it pretty basic. I'd appreciate if someone else could beautify it later on ;-) .


andryyy:
Very cool! 👍 I think I can add the text to the lang files tomorrow.

One of the problems I see is with the TLSA records. acme-mailcow will recreate the key when we remove a name from the certificate. We can stop it from doing so automatically, but it is just a workaround as we cannot response to further challenges successfully.

mkuron:
Here's the change for handling non-standard port numbers (they are pulled from SRV records).

> acme-mailcow will recreate the key when we remove a name from the certificate.

Really? I thought that's what the `-N` flag is for, it only creates a new key if none is available.

mkuron:
Actually, the private and account key are not tied to a domain. Just remove lines 102 and 103 from the ACME script and we're good.

andryyy:
No, it complains about the account key on new requests... The private key can be the same, yes, but we cannot remove that line and use the snakeoil one...

I will add a check and remove/regenerate the private key when it is really needed.

mkuron:
Ok, sounds good. You could just include a hash of the snakeoil private key in the script and check against that.

andryyy:
Good idea. But I also forgot it checks ACME_BASE/privat/private.key and not ACME_BASE/private.key.

I think I run into this problem when updating an instance with a LE certificate from "outside" (certbot via old method).

Should be fixed in a bit.

data/web/diagnostics.php
```diff
@@ -0,0 +1,273 @@
...
+  foreach (explode(':', $ip6_full) as $part) {
+    $ptr6 .= str_pad($part, 4, '0', STR_PAD_LEFT);
+  }
+  $ptr6 = implode('.', array_reverse(str_split($ptr6, 1))) . '.ip6.arpa';
+}
+
+function get_tlsa($host, $port, $starttls = '') {
+  if ($starttls)
+    $starttls = ' -starttls ' . $starttls;
+  return '3 1 1 ' . trim(shell_exec('echo | openssl s_client -connect ' . $host . ':' . $port . ' -servername ' . $host . $starttls . ' 2>/dev/null | openssl x509 -noout -pubkey | openssl pkey -pubin -outform DER | openssl sha256 | awk \'{print $2}\''));
```
andryyy: Sorry, I will not use shell_exec anymore. You can use openssl via PHP for this.

data/web/autodiscover.php
```diff
@@ -1,36 +1,22 @@
...
+
 /* ---------- DO NOT MODIFY ANYTHING BEYOND THIS LINE. IGNORE AT YOUR OWN RISK. ---------- */
 
 error_reporting(0);
 
 $data = trim(file_get_contents("php://input"));
 
 // Desktop client needs IMAP, unless it's Outlook 2013 or higher on Windows
-if (strpos($data, 'autodiscover/outlook/responseschema')) { // desktop client
+if (strpos($data, 'autodiscover/outlook/responseschema') !== FALSE) { // desktop client
```
andryyy: This is wrong in master?

andryyy:
SPF and DMARC should only be checked for existence. 

Edit: A SRV check for autodiscover should also exist. There are clients to only support A or CNAME or SRV (yes, they suck, but we should mention it).

A 3 0 1 TLSA record is also valid, though not recommended with short-lived LE certs. We should at least explain that somewhere. Or link to the docs, I don't know.

andryyy:
Clientconfig.inc.php=> Read the env vars please, I don't want any new configuration file for this. We can pass them in docker-compose.yml. Please also don't declare functions in the committed files, just provide them via functions.inc.php. I think there are too many new files for such a relatively small function.

I also think it is better to not do DNS checks in the UI at all. They may confuse users as they are delayed - I see this happening @ Servercow, where I implemented something like that. Just print stuff like the public key digest (TLSA), MX etc.

I think most of the code works fine for _you_, but this will not work for all setups out there. Some stuff is very hacky or relies on other configurations or success of previous operations (examples: What happens when the Integrator failed to build? What if there is no SRV record for the first domain in SQL?).
Same applies to the CNAME check with dig acme-mailcow: `dig A autoconfig.${SQL_DOMAIN} +short | tail -n 1` - `tail -n1` is a dirty hack. That's not how you check for a CNAME. This can fail with newer versions of dnsutils.
I used so many hacks like these in the previous mailcow versions. Most work temporarily, some may work forever. But it is just wrong.

And please don't use foreign APIs. :-) I don't know their policy, rate limits, etc...

The change regarding ssl => ssl-example was not well thought through, too. Works on new installations, removes the ssl directory when updating old installations. Just because something seems correct, doesn't mean it _is_ correct. We need much more testing of PRs. There are so many different installations...

data/web/inc/clientconfig.inc.php
```diff
@@ -0,0 +1,143 @@
...
+     'activesync' => array(
+       'url' => 'https://'.$mailcow_hostname.'/Microsoft-Server-ActiveSync',
+     ),
+);
+
+function get_client_config() {
+  global $pdo, $client_config;
+  $config = $client_config;
+  
+  // use the SRV records of the first domain to obtain the correct port numbers
```
andryyy: What if they don't exist? They are not mandatory.

data/web/autodiscover.php
mkuron: Yes, but it doesn't actually matter, it still worked. `strpos` should never be compared in a non-typesafe way as it may return both `0` and `FALSE`.

data/web/inc/clientconfig.inc.php
mkuron: The default values are not overwritten if the SRV records don't exist. They are already in the `$config` array.

mkuron:
> SPF and DMARC should only be checked for existence.

SPF is being checked to ensure that `$mailcow_hostname` is permitted by the SPF rule. DMARC I now only check for it being present.

> A SRV check for autodiscover should also exist. There are clients to only support A or CNAME or SRV (yes, they suck, but we should mention it).

Ok, added that.

> A 3 0 1 TLSA record is also valid, though not recommended with short-lived LE certs. We should at least explain that somewhere. Or link to the docs, I don't know.

Is there ever a reason to use a "3 0 1" instead of a "3 1 1", even with longer-lived certificates?

> Clientconfig.inc.php=> Read the env vars please, I don't want any new configuration file for this. We can pass them in docker-compose.yml. Please also don't declare functions in the committed files, just provide them via functions.inc.php. I think there are too many new files for such a relatively small function.

This is not a configuration file, it just declares functions. That `$client_config` array is not meant to be modified by the user. I'll remove the array and put the functions into functions.inc.php

> I also think it is better to not do DNS checks in the UI at all. They may confuse users as they are delayed.

It's useful -- it certainly helped me find that I had forgotten to add one DNS record. Putting a warning at the top that this information may be out of date for up to 24 hours should help avoid confusion.

> I think most of the code works fine for you, but this will not work for all setups out there.

I tried to make it as generic as possible. Specifically, that's why I added all the SRV handling. I don't need it, I'm running a standard setup.

> Some stuff is very hacky or relies on other configurations or success of previous operations (examples: What happens when the Integrator failed to build? What if there is no SRV record for the first domain in SQL?).

When the integrator fails to build, nothing happens. It is simply not built and the download link to it will 404. That seems like correct behavior to me.
When no SRV record is present, it is assumed that no domain has one. The reason I make this assumption is that autoconfig.php does not know what domain it is being requested for, so not all paths that lead to `get_client_config()` know the domain.

> Same applies to the CNAME check with dig acme-mailcow: dig A autoconfig.${SQL_DOMAIN} +short | tail -n 1 - tail -n1 is a dirty hack. That's not how you check for a CNAME. This can fail with newer versions of dnsutils.

I don't want to check for the presence of a CNAME, I want to check for the presence of an A record or a CNAME that leads to one. I didn't find any option to `dig` that allows me to do it without `tail`.

> And please don't use foreign APIs. :-) I don't know their policy, rate limits, etc...

I know that, and you've already replaced it in this PR.

mkuron:
I've made a number of changes that try to address your concerns. Could you please take another look and let me know if there's anything else you want me to change?

mkuron:
I've now removed the stupid hack of using the first domain's SRV records in autoconfig.php. I can just get the domain name (example.com) from the HTTP host name (autoconfig.example.com), obviously.

mkuron:
Merged current master and fixed a problem due to `pem_to_der` being nested inside `generate_tlsa_digest`. PHP [doesn't actually support](https://stackoverflow.com/questions/415969/what-are-php-nested-functions-for) nested functions.

mkuron:
@andryyy, I saw that you already copied a few little changes from this PR to the dev branch. What do you want me to do so we can get the rest merged? A lot of this stuff would be quite useful for many people, e.g. the client configuration guide or the Apple mobileconfig generator.

mkuron:
I've updated the PR to current master.

I also removed the code that pulls port numbers from SRV records as that is likely not needed by anyone; environment variables should suffice. (I still kept the checks for SRV records with an empty target addres though as we [document](https://mailcow.github.io/mailcow-dockerized-docs/prerequesite-dns/#the-advanced-dns-configuration) that as the official way to announce services as not provided.)

Is there any other things you want me to remove before this can be merged, @andryyy?

andryyy:
I don't really see why this should be in the UI. It still has flaws and is very large for what it does.

```
+    'sogo' => array(
 +      'server' => $mailcow_hostname,
 +      'port' => '443',
 +      'ssl' => 'on',
 +    ),
 +    'http' => array(
 +      'server' => $mailcow_hostname,
 +      'port' => '443',
 +      'ssl' => 'on',
```

I don't want a user to update config vars just because he changed a port or bound it to localhost (what happens by the way when a port is given with an IP address, like 127.0.0.1:143?).
This PR introduces much more configuration complexity to mailcow, than it already has. :/ You could also read $_SERVER['HTTP_HOST'] for some urls to get the correct port. And check if SSL is enabled by reading $IS_HTTPS. 

I would like mailcow to use even less variables. We should also get rid of that "ssl = on" stuff, I don't really see a  need for it. SSL is always on for IMAP and SMTP, even for the UI.

The DNS propagation delay is still a problem. The SPF check should not show a value. There is no recommended value, setups differ. Same for some other records. I also wonder why it says "X" for servercow.de's SPF. :o 

Autodiscover/config can also be A, is this checked?

This has many "ifs" and "buts" and "except for". :/

mkuron:
Thanks for the comments.

> I don't really see why this should be in the UI.

Which part? The config guides need to be in the UI (and not the docs repository) because they contain domain-specific information. Also, they are for the user, not for the server admin, so they need to be in a place where the user finds them.

> I don't want a user to update config vars just because he changed a port 
> You could also read $_SERVER['HTTP_HOST'] for some urls to get the correct port. And check if SSL is enabled by reading $IS_HTTPS.

Ok, I'll change it to get that information from HTTP_HOST.

> or bound it to localhost (what happens by the way when a port is given with an IP address, like 127.0.0.1:143?).

Good point, I hadn't considered that case. Should I just extract the port number from the variable?

> This PR introduces much more configuration complexity to mailcow, than it already has. :/ 

It is aiming to reduce configuration complexity. It should not add any new settable variables; instead it should pull everything from existing environment variables. It even gets rid of the current global `$autodiscover_config` (the only reason why I still check it is in case somebody modified `useEASforOutlook`). Previously, that global needed to be modified manually when somebody ran services on non-standard ports.

> We should also get rid of that "ssl = on" stuff, I don't really see a need for it. SSL is always on for IMAP and SMTP, even for the UI.

I'm happy to do that. I'll just assume everywhere that SSL is used and make the appropriate changes. Can I also assume everywhere that HTTPS is used?

> The DNS propagation delay is still a problem. The SPF check should not show a value. There is no recommended value, setups differ. Same for some other records.

I can remove the DNS page entirely if you don't like it. The SPF check does not check against the value, but I still wanted to provide a suggested value for the most common set up. Same goes for DMARC.

> Autodiscover/config can also be A, is this checked?

No, but I can add that.

mkuron:
I made the changes you requested. Please have another look, @andryyy.

Since that was the most controversial part, I removed the DNS page for now. I'll post a new PR some time in the future for just the DNS page.

Issue #401, opened 2017-06-24
Major Problems with acme Container


brunick:
I cannot update nor restore my installation, caused by this error while docker-compose pull
```
Pulling acme-mailcow (mailcow/acme:1.4)...
ERROR: manifest for mailcow/acme:1.4 not found

```
current head, already removed everything and tried on a clean system

mkuron:
Confirmed. @andryyy hasn't pushed that version yet to https://hub.docker.com/r/mailcow/acme/tags/. You can run `docker-compose build acme-mailcow` in the meantime or replace that container with version 1.3 in docker-compose.yml.

brunick:
all right, i sorta found out, that i could ignore the pull of acme by "--ignore-pull-failures" and my server could be recreated - i also got some problems with rspamd, the whitelist from the mailcow does not work anymore...

andryyy:
And that's why you automate these tasks. :P Sorry.

andryyy:
When was your last update? Are the whitelists still present? Can you run `docker-compose exec php-fpm-mailcow curl http://nginx:8081/settings.php` and post the output (censored)?

brunick:
this is the output, i've received a couple of mails from hardwareluxx, these mails did get the following header: "X-Spamd-Result: default: False [6.20 / 15.00]"
```
root@mx:~/mailcow-dockerized# docker-compose exec php-fpm-mailcow curl http://nginx:8081/settings.php
settings {
        score_info###### {
                priority = low;
                from = "/^((?!.*@########.de|.*@#########.de|.*@########.de|.*@#########.com|.*@hardwareluxx.de).)*$/";
                rcpt = "/info\+.*@######.de/";
                rcpt = "info@######.de";
                apply "default" {
                        actions {
                                reject = 15;
                                greylist = 5.5;
                                "add header" = 6.5;
                        }
                }
        }
        whitelist_info###### {
                from = "/(.*@########.de|.*@#########.de|.*@########.de|.*@#########.com|.*@hardwareluxx.de)/";
                priority = high;
                rcpt = "/info\+.*@######.de/";
                rcpt = "info@######.de";
                apply "default" {
                        MAILCOW_WHITE = -999.0;
                }
                symbols [
                        "MAILCOW_WHITE"
                ]
        }
```

andryyy:
Can you forward this email to info@servercow.de ? 

[...]

PR #404, opened 2017-06-25
Automated builds using Travis-CI

mkuron:
@andryyy, here's a first suggestion for building on Travis (as discussed in https://github.com/mailcow/mailcow-dockerized/issues/401#issuecomment-310859925). You can use an API token instead of your real Docker Hub password, and you store it in an encrypted variable on the Travis web site. For testing, you can remove the last three lines that limit builds to master and push this to a new branch. Once it's working, put the file into master. I have only done very limited testing so far, but it should be quite easy to get this set up.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,15 @@
+sudo: required
+
+services:
+  - docker
+
+script:
+- echo 'Europe/Berlin' | MAILCOW_HOSTNAME=build.mailcow ./generate_config.sh
+- docker-compose pull --ignore-pull-failures --parallel
+- docker-compose build
+- docker login --username=$DOCKER_HUB_USERNAME --password=$DOCKER_HUB_PASSWORD
+- docker-compose push
+
+branches:
+  only:
+  - master


andryyy:
Hi, thanks, I stopped at creating the encrypted variables yesterday evening. So this is pretty much what was missing. :-)

andryyy:
Eh, this does not check if images exist. This was why I did not want to use Docker Hub. Slight changes to the (example) Debian image can break stuff. I want the images to be tested locally. If I forget to push them, this is a fallback.

Will add it when I am back.

mkuron:
> Eh, this does not check if images exist.

It should. It pulls the old images and will only rebuild and upload if they are outdated.

> Slight changes to the (example) Debian image can break stuff.

That shouldn't happen, but can obviously not be ruled out, however unlikely.

> If I forget to push them, this is a fallback.

Ok, sounds reasonable.

ntimo:
Any news on this? Automated building of images would be great and would also increase the transparency.

PR #405, opened 2017-06-25
Fix EAS autodiscover for certain Outlook variants

mkuron:
Fixes the issue discovered by @andryyy (https://github.com/mailcow/mailcow-dockerized/pull/370#issuecomment-310895640). Evidently, the MSI-installed and the click-to-run-installed versions of Office have different user agents.

PR #406, opened 2017-06-25
Fix EAS autodiscover for certain Outlook variants (second try)

mkuron:
Oversight in #405. This should finally do it... I removed the Windows check altogether as the Mac version turned out to not support autodiscover anyway.


andryyy:
I was working on a fix and didn't see this.

Btw: the connection analyzer _does_ work with EAS.

mkuron:
It didn't for me, but feel free to remove that line if it works for you.

PR #427, opened 2017-07-02
Client configuration guides

mkuron:
This pull request supersedes #389, trying to address all of @andryyy's concerns. It is greatly simplified and contains way fewer `if`s and special cases. It no longer contains the DNS status page, but I will resubmit that for discussion in a future PR.

This PR is much more lightweight than the previous one. It only adds 52 lines to existing files and edits 56 lines -- and most of these are trivial changes. All the remaining code is inside two new files, one for the client configuration guides and one for the Apple mobileconfig generrator.

Improvements:

- Client configuration guides
- macOS/iOS mobileconfig generator
- non-443 HTTPS ports in autoconfig etc.
- disabling POP service via SRV record
- fix display name in Outlook IMAP autodiscover
- allow multiple calls to TLSA generator and support Sieve STARTTLS
- proper treatment of bind addresses in the port number environment variables

### DIFF
--- a/data/Dockerfiles/sogo/Dockerfile
+++ b/data/Dockerfiles/sogo/Dockerfile
@@ -8,7 +8,6 @@ ENV GOSU_VERSION 1.9
 # Prerequisites
 RUN apt-get update && apt-get install -y --no-install-recommends \
 		apt-transport-https \
-		bind9-host \
 		ca-certificates \
 		cron \
 		gnupg \
--- a/data/Dockerfiles/sogo/Dockerfile
+++ b/data/Dockerfiles/sogo/Dockerfile
@@ -8,6 +8,7 @@ ENV GOSU_VERSION 1.9
 # Prerequisites
 RUN apt-get update && apt-get install -y --no-install-recommends \
 		apt-transport-https \
+		bind9-host \
 		ca-certificates \
 		cron \
 		gnupg \
--- a/data/web/autoconfig.php
+++ b/data/web/autoconfig.php
@@ -5,6 +5,16 @@ if (empty($mailcow_hostname)) {
   exit();
 }
 
+$domain_dot = strpos($_SERVER['HTTP_HOST'], '.');
+$domain_port = strpos($_SERVER['HTTP_HOST'], ':');
+if ($domain_port === FALSE) {
+  $domain = substr($_SERVER['HTTP_HOST'], $domain_dot+1);
+  $port = 443;
+} else {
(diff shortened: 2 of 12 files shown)


mkuron:
I have re-checked all the logic and believe this is ready to be merged. Could you please have a look, @andryyy?

felixoi:
I think @andryyy decided to switch from bind9 to unbound, or? Just to be neat we could use the unbound-host package instead of the bind9-host one.

data/Dockerfiles/sogo/Dockerfile
```diff
@@ -8,6 +8,7 @@ ENV GOSU_VERSION 1.9
 # Prerequisites
 RUN apt-get update && apt-get install -y --no-install-recommends \
 		apt-transport-https \
+		bind9-host \
```
felixoi: I think @andryyy decided to switch from bind9 to unbound. Just to be neat we could use the unbound-host package instead of the bind9-host one, or is there a complex reason to use bind9?

mkuron: Actually, I chose to use bind9 to [match](https://github.com/mailcow/mailcow-dockerized/blob/9040d45/data/Dockerfiles/acme/Dockerfile#L10) the acme container, which also uses `dig` (that's the only other place in Mailcow that performs an explicit DNS lookup).
From a technical perspective, it makes no difference whether we use `dig` or `unbound-host`, so we can just keep `dig` here.

mkuron:
I just added proper treatment of bind addresses in the port number environment variables. On the current master, autodiscovery was broken if a bind address was present.

mkuron:
Fixed merge conflict. @andryyy, can we please get this merged? I have wasted way too much time on resolving merge conflicts over the past few weeks...

If you don't think this is ready to be merged: please let me know what you want me to change.

andryyy:
I will merge this to dev and check it. I don't know if it will be merged to master. I think this just belongs to the docs...

andryyy:
Just because you disable autodiscover EAS it does not mean you should not show it in the guide. It is perfectly fine to disable EAS AS and still make it available. The default is IMAP/SMTP then, but EAS is still fine.

andryyy:
And please remove the integrator stuff.

mkuron:
Ok, will send another PR later today.

andryyy:
I saw you just omit the protocol in `<Server>https://<?=$configuration['carddav']['server'];?>` - this may work in emClient, but in other clients, too? :-/

The Android configuration guide will be very different for every phone. We may recommend an app for those, don't know...

For every change in a clients configuration guide, mailcow would need to be updated. That's strange. I would really love to see this guide in the docs, not in the admin UI.

PR #444, opened 2017-07-09
DNS diagnostics page

mkuron:
This PR adds a page that shows what DNS records should be created and whether they have already been created and have the correct value. Obviously, the latter may be delayed by up to 24 hours. This was originally part of #389 and depends on #427.


andryyy:
I would really like to have this in the docs. There is too much that cannot be shown reliable.

mkuron:
>  There is too much that cannot be shown reliable

It's reliable, you just need to wait 24 hours between making DNS changes and looking at this page. There is a warning at the top of the page about this.

> I would really like to have this in the docs..

The docs already contain a list of all relevant records, this PR just makes it a lot easier to see whether everything is correct and what values the installation-specific records (e.g. TLSA) should have.

mkuron:
Why did you merge this already? As I said, it depends on #427 (due to changes in functions.inc.php and vars.inc.php), which is not merged yet.

andryyy:
Besides the diagnostics.php in the root directory (?) it does not show anything.

Edit: It does now, but empty values for some things.

andryyy:
Remove the SPF stuff and DMARC, please. Or just write "SPF" or something, don't tell the users this single record is fine (I know you already check for a covered SPF, but still this is misleading). The DMARC record can actually be removed completly. Or mark it as optional. This should not be added without knowing anything about it, nobody needs it when they don't know what it is and does. :-)

PR #449, opened 2017-07-10
Autoconfig/Autodiscover unification

mkuron:
@andryyy merged #427 and #444 this morning and later reverted them with a few comments. This PR contains the stuff that he did not comment on, which I thus assumed to be ok as-is. It's all just simple stuff that fixes some little bugs and unifies variable usage between autoconfig (Thunderbird) and autodiscover (Outlook etc.).

- use `$autodiscover_config` everywhere
- non-443 HTTPS ports in autoconfig etc.
- disabling POP service via SRV record
- fix display name in Outlook IMAP autodiscover
- fix eM Client CardDAV/CalDAV autodiscover
- allow multiple calls to TLSA generator and support Sieve STARTTLS
- iOS mobileconfig generator

I will later post separate PRs for the client configuration page and the DNS status page. I think this PR should be merged regardless of whether the other two are accepted.


mkuron:
> I saw you just omit the protocol in `<Server>https://<?=$configuration['carddav']['server'];?>` - this may work in emClient, but in other clients, too? :-/

The protocol is there (`https://`). I replaced `$configuration['carddav']['url']` with `$configuration['carddav']['server']` (i.e. removed the protocol and put that directly into the XML) as the Apple mobileconfig generator needs the hostname (not a URL).

mkuron:
Thanks!

PR #451, opened 2017-07-10
DNS diagnostics page

mkuron:
@andryyy merged #427 and #444 this morning and later reverted them with a few comments. This PR just contains the DNS diagnostics page updated according to his comments, which I have included and addressed below. It needs to be merged after #449 as it depends on that PR.

----------

> Remove the SPF stuff and DMARC, please. Or just write "SPF" or something, don't tell the users this single record is fine (I know you already check for a covered SPF, but still this is misleading). The DMARC record can actually be removed completly. Or mark it as optional. This should not be added without knowing anything about it, nobody needs it when they don't know what it is and does. :-)

Done that. I now link to the assistant from the documentation instead of showing a recommended value. Instead of a checkmark or cross, it now just shows the current value and that it's optional.

> Besides the diagnostics.php in the root directory (?) it does not show anything.

Ooops, sorry. Fixed that now.


schulluk:
Can you please fix the conflicts?
I'm really interested in this extension :-)
Thank you for your work!

mkuron:
@lukas-schulze: I resolved the conflicts.

PhoenixPeca:
@mkuron, I think this PR is great! Have tested it locally. Any news about it getting merged? But there is one thing that bugs me, domainkey/DKIM record failed the detection (at least for my side, can you please verify?)

PhoenixPeca (commented):
Would be better if it can also check the [Advanced DNS Configuration](https://mailcow.github.io/mailcow-dockerized-docs/prerequesite-dns/#the-advanced-dns-configuration) as defined in the docs

PhoenixPeca (commented):
It might be better using [ipify](https://www.ipify.org/) to retrieve the public IP. The ipv6 curl checker just doesn't work anymore.

mkuron:
It all worked for me when I last tried it a few weeks ago, including DKIM. Advanced DNS configuration (SRV records) is supported and it shows up as soon as you use non-default ports. The IP checkers were set up by @andryyy, so he would need to fix them if they are broken. @andryyy didn't want to merge this PR so far because he found the time delay of DNS records confusing for users, so it will probably just remain something people can optionally merge locally.

jhesketh:
@mkuron Regarding the time delay concern, I think this could be addressed by a simple note that changes may take some time to happen. You could even look up an existing record's TTL and use that as a minimum.

If the user confusion is a concern, I believe it is more important to highlight this as a user may have made the DNS changes but will still need to wait and become confused when the effect hasn't taken place yet.

schulluk:
There might be also another optimization: I have multiple domains and it takes quite a while to load the page. It would be nice if the requests would be send asynchronously and provide some instant feedback to the user

develth:
@lukas-schulze how many domains you have set up?
Currently i didn´t have a huge delay. 

But +1 with the asynch setup - i will have a look on it. 

Is it important to have the Diagnostics for each Domain on the Page? I don´t think so. 
What about to make a search input field and just show the Diagnostics of the cow-host and the searched domain?

schulluk:
@develth: I currently have 16 domains in mailcow

What about a 2-level check?
Only check base dns entries on page load (async) and do all the others on user interaction "Check all for this domain"?

andryyy:
@mkuron Are you fine with me changing it to a modal? We can add a button in the domain table like "check dns". Could run as ajax call then.

develth:
@andryyy i would agree -IMHO the best place for domain config. And the global DNS overview into configuration?

mkuron:
@andryyy, that‘s fine with me.

@develth, I‘d suggest we show the global configuration on every domain-specific modal so that we don‘t need to spread DNS between multiple pages.

develth:
@mkuron I wouldn´t put general stuff in domain specific sections ( maybe you just would show the domain DNS fields to Domain Admins - and they should not see the Global DNS overview).

What about:

* **Configuration Administration-> Bottom**: A Icon Indicator if DNS State is ok (green ok, red missing). This flag could be set by Cron / AJAX. Beside is a Button / Link that show the Detail of the cow DNS.
* **Configuration Mailboxes -> Domains**:  New Multi Action DNS Diagnostic. Show DNS for selected Pages. Add a table row with indicator if DNS is ok or not. this flag could be set by Cron / AJAX.
* **Configuration Mailboxes -> Domains Edit Bottom**: DNS Indicator with Collapsable or Modal Detail. Indicator Cron / AJAX - Collapsed / Modal loaded by AJAX request.

I think that would be the best UX. DNS Settings are important to ensure mail delivery, so lets move them into users focus.

PR #468, opened 2017-07-22
Client configuration guide link

mkuron:
This supersedes #450.

I was able to move everything to the docs repository (https://github.com/mailcow/mailcow-dockerized-docs/pull/37). When the page in the documentation is accessed via the link on the user page, the necessary variables (host name, whether Outlook EAS is enabled, etc.) are passed via the URL anchor and the documentation is dynamically modified via JavaScript. The variables are never sent to the server as the anchor is only dealt with by the browser.

~~#449 needs to be merged before this PR works.~~

### DIFF
--- a/data/web/lang/lang.de.php
+++ b/data/web/lang/lang.de.php
@@ -108,6 +108,7 @@ $lang['user']['user_settings'] = 'Benutzereinstellungen';
 $lang['user']['mailbox_settings'] = 'Mailbox-Einstellungen';
 $lang['user']['mailbox_details'] = 'Mailbox-Details';
 $lang['user']['change_password'] = 'Passwort ändern';
+$lang['user']['client_configuration'] = 'Konfigurationsanleitungen für E-Mail-Programme und Smartphones anzeigen';
 $lang['user']['new_password'] = 'Neues Passwort';
 $lang['user']['save_changes'] = 'Änderungen speichern';
 $lang['user']['password_now'] = 'Aktuelles Passwort (Änderungen bestätigen)';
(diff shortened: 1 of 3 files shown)


andryyy:
Can you add a note to "Apple macOS / iOS" for the "Mailcow.mobileconfig" link when clicked without coming from a mailcow installation? Or remove that part with JS? A note saying what is configured by that config file would be nice, too. :-) Cal/CardDAV + IMAP etc.

Windows should be "Windows >= 8" or "Windows Mail" ... something like that. Do they support EAS?

Another info box in the docs would be nice, like... "When coming from a mailcow installation, all guides will have prefilled values blabla according to your installation blabla". And if they came from a mailcow, notify that all values were prefilled etc. You know what I mean? :-) Can you do that?

mkuron:
Ok, give me a few minutes. I'll post another pull request for that soon.

andryyy:
Thanks. 👍 

Works for me in Chrome, but not in Firefox though. :/ I browsed the docs without coming from mailcow, then logged in to mailcow and browsed the docs again. Did it cache that I am not logged in?

mkuron:
I'll check and fix it if necessary.

andryyy:
Whereever autoconfig/autodiscover is used, it should say that. There should also be a guide to setup a client manually.

I know this is quite a lot work. :/ I can try to find some time later.

mkuron:
Manual configuration is quite a lot of work (at least if you want not just IMAP but also CalDAV and CardDAV), so I relied on automation wherever possible. That's also the reason why there is no guide for Evolution -- it cannot be autoconfigured. We could make one generic guide that lists server address, port numbers, DAV URLs, etc., but we should rely on working autoconfiguration to make our lives easier.

PR #486, opened 2017-08-02
Disable ClamAV phishing filter

mkuron:
ClamAV has a phishing filter built in, but it generates too many false positives as it only has some very simple heuristics for detecting phishing. Any email it suspects to be phishing is rejected with a status message claiming it contains a virus. Obviously this is not true.
We can disable ClamAV's phishing filter without risk because rspamd also checks for phishing itself. Since rspamd does not only do simple text analysis, but also considers headers and the training of its Bayesian filter, it is much more suited than ClamAV to deliver accurate judgements over whether an email is really phishing.

### DIFF
--- a/data/Dockerfiles/clamd/Dockerfile
+++ b/data/Dockerfiles/clamd/Dockerfile
@@ -15,6 +15,8 @@ RUN apk add --update \
 	&& chmod 750 /run/clamav \
 	&& sed -i '/Foreground yes/s/^#//g' /etc/clamav/clamd.conf \
 	&& sed -i '/TCPSocket 3310/s/^#//g' /etc/clamav/clamd.conf \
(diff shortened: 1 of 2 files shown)


andryyy (commented):
Thanks, good idea. 

But I wonder if Rspamd ClamAV module is able to ignore phishing matches. There is an example for Eicar, I think.

mkuron:
@andryyy, one can indeed add a symbol based on that example and use a force_action to let ClamAV-detected phishing pass. However, this would let an email that contains both a virus and phishing pass (while an email that only contains a virus would not pass) -- and we never want viruses to pass. So the only way to ignore phishing in ClamAV is to disable the checks like I did.

andryyy:
Okay, then this should be better. :-)

PR #512, opened 2017-08-17
#511 change SQL statement

### LINKED ISSUE #511: Problems after migrating from mailcow 0.14 with le certificates
Hello,

i tried to migrate from non dockerized mailcow to dockerized mailcow.
I had a lot of Backup MXs for high availability of other domains.

The acme container or better the LetsEncrypt script inside fails a lot of times. The rate limit for failed requests is reached because the backup mxs have no matching A records.

A quick view in code shows that there is no limit and it tries to create certificates for all domains. No matter if its backup mx or not.

topelrapha:
try to fix #511


mkuron (approved):

andryyy:
Thanks!

Issue #524, opened 2017-08-22
Proposition Improvement : update.sh docker pull before docker stop


JOduMonT:
Hi;

Would be better if the docker pull the new version before stopping all service ?
Such as yesterday I did a ./update.sh
and during the pull something when wrong
- my connection was super slow
- the pull stop in the middle of SOGo

It's only a proposition ;)

Thanks for you're support and your awesome compilation of software

mkuron:
Sounds reasonable as it helps in keeping the downtime as short as possible. You should submit a pull request with that change.

JOduMonT:
I hope I did well ;)

normanu:
This would be preferable, but is still not the case.
As we speak I'm updating and it pulled down the stack.
Dovecot image was not available so it is building the image itself, this takes a long time.
Would have been better if the services where not down in the mean time.

andryyy:
See the other issues, this led to problems and was reverted.

The images are all online and available: https://hub.docker.com/r/mailcow/dovecot/tags 

Please check your compose file against origin/master.

PR #525, opened 2017-08-23
move docker stop after docker pull

JOduMonT:
as proposed at: https://github.com/mailcow/mailcow-dockerized/issues/524
move docker-compose down after docker pull


mkuron (approved):

lazyfrosch:
Technically, is there a real need to "down"?

`up -d` should restart all necessary containers.

Correct would be even:

     docker-compose up -d --remove-orphans

This will remove old containers as well.

andryyy:
Can you base this on dev?

PR #536, opened 2017-08-27
Rspamd user blacklist/whitelist improvements

mkuron:
- Make regex matching case-insensitive.
  - The web UI normalizes all white-/blacklist entries to lower case, but email addresses in an SMTP conversation may have any (mixed) case.
- In addition to the Envelope-From, match the From header.
  - The average end user doesn't know about Envelope-From (or how to extract it from the email headers in their email client), so they expect to be able to add a From address to their blacklist/whitelist.
  - Some mailing lists and newsletters send messages with different Envelope-From and From, so this change is actually necessary to be able to correctly handle certain messages.
  - This is not a security/spam issue as Envelope-From is just as easily forged as From headers.
- For domain-global whitelist/blacklist entries, settings.php should not print all email addresses in that domain (potentially making the settings map really large), but a domain regex.
  - The `ucl_rcpts` function was already capable of this, but needed a trivial bug fix and be called with `$type = 'domain'`.

### DIFF
--- a/data/conf/rspamd/dynmaps/settings.php
+++ b/data/conf/rspamd/dynmaps/settings.php
@@ -182,6 +182,47 @@ while ($row = array_shift($rows)) {
 			"MAILCOW_WHITE"
 		]
 	}
+	whitelist_header_<?=$username_sane;?> {
+<?php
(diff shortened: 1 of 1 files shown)


andryyy:
I wonder that you could follow this mess. 👍 That's one of the worst scripts in mailcow. Thanks for fixing it!

mkuron:
Well, it's still messy, but it could have been worse...

On a different note: are you going to do another release soon (merge dev into master)? Dev contains a few useful fixes and little improvements.

PR #543, opened 2017-09-02
Formatting

Dennis14e:
I know the pull request is not small.
It still needs to be checked whether the code still works as I have not tested it.


mkuron:
This is probably not a good idea. It will cause merge conflicts for everyone who has local changes in their PHP code. Don't get me wrong, I agree that a consistent coding style helps improve readability, but intoducing it has a major one-time inconvenience.

Dennis14e:
Fine.

PR #544, opened 2017-09-02
Use different exit code on updates check

max-baz:
Enables some scripting, e.g. putting this in a cron job:

    ./update.sh --check && <notify-me-about-available-updates>

By the way, what is the recommended way to follow updates? I'm preparing to switch from Mail-in-a-box, and there users receive an email when a new release is available. I found this `./update.sh --check` which with this PR merged will be able to help me follow updates, but is there maybe an alternative that everyone is currently using?

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -51,10 +51,11 @@ case "${1}" in
 		git fetch origin ${BRANCH}
 		if ! git diff origin/${BRANCH} --quiet; then
 			echo "Updated code is available."
+			exit 0
 		else
 			echo "No updates available."
+			exit 3
 		fi
-		exit 0
 	;;
 esac
 


mkuron:
You can subscribe to the RSS feed, https://github.com/mailcow/mailcow-dockerized/commits.atom . It's for the master branch, which contains the releases.

mkuron (approved):

andryyy:
Can you change -1 to something from 3 to 255? 👍 I am not sure about which exit codes can/should be used right now. But I think we should avoid a negative int.

max-baz:
Huh, a negative was not intended to be there, thanks for catching this! :)

I see all around that script `exit 1` is used, do you want me to use `exit 3` only in this particular case?

andryyy:
Oh, okay. :) Hm, 3 would be good!

max-baz:
Then 3 it is :+1:

PR #549, opened 2017-09-03
Change primary name of SSL certificate

tbsbckr:
Changes the primary name for the SSL certificate to the hostname of mailcow, all other DNS names will be added as SAN (same procedure as before)


mkuron (approved):

andryyy:
Thanks to both of you! 👍

PR #565, opened 2017-09-09
Forwarding hosts: treat soft reject like greylist

mkuron:
The _greylist_ action actually turns into _soft reject_ by the time it reaches the _force_actions_ module. So we should treat _soft reject_ the same as _greylist_ when we deal with forwarding hosts, meaning the message should not be flagged as spam.

The old behavior is inconsistent: mails from forwarding hosts are flagged as spam on scores higher than 4.0, while mails from other hosts are flagged above 5.0. This PR makes that consistent by using 5.0 for both.


andryyy:
Nice, thanks!

And many more thanks for cleaning up the issues! I owe you some Bitcoins. :)

PR #567, opened 2017-09-09
a little bit of security

JOduMonT:
Hide the version of NGINX, block XSS and more...
inspired by : https://gist.github.com/plentz/6737338

tested with : https://securityheaders.io
it pass the project from Grade : E fo B

### DIFF
--- a/data/conf/nginx/site.conf
+++ b/data/conf/nginx/site.conf
@@ -1,3 +1,4 @@
+server_tokens off;
 # includes to http {
 proxy_cache_path /tmp levels=1:2 keys_zone=sogo:10m inactive=24h  max_size=1g;
 server_names_hash_bucket_size 64;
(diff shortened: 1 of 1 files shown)


mkuron (approved):

andryyy:
This breaks Roundcube for those who have it installed. Should be in the docs.

K2rool:
If your talking about the broken Roundcube issue that the non docker version had months ago that was due to the use of X-Frame-Options "DENY"; which break Roundcube due to the frame holding <pre>/skins/larry/watermark.html</pre> getting blocked which makes Roundcube near unusable.

I think config changes on this pull will work fine with users on Roundcube, not too sure about SoGo as I don't rely use it that much to be told.

.

andryyy:
I see. :)

andryyy:
Breaks SPDY: ERR_SPDY_PROTOCOL_ERROR

That's why you never just add some options blindly that sound like they make anything more secure. :-/

mkuron:
Hmm, probably one of the SSL lines. The headers are unlikely have such an effect. They are the more important part anyway as they prevent XSS.

andryyy:
Yes :( have had no time to check it yet, commented those lines for now.

PR #568, opened 2017-09-09
Limit DOS Attack

JOduMonT:
base on: https://security-24-7.com/hardening-guide-for-postfix-2-x/

### DIFF
--- a/data/conf/postfix/main.cf
+++ b/data/conf/postfix/main.cf
@@ -99,3 +99,10 @@ smtp_sasl_auth_enable = yes
 smtp_sasl_password_maps = proxy:mysql:/opt/postfix/conf/sql/mysql_sasl_passwd_maps.cf
 smtp_sasl_security_options = 
 smtp_sasl_mechanism_filter = plain, login
+# Limit Deny of Service Attacks
+default_process_limit = 100
+smtpd_client_connection_count_limit = 10
+smtpd_client_connection_rate_limit = 30
+queue_minfree = 20971520
+header_size_limit = 51200
+smtpd_recipient_limit = 100


mkuron (changes requested):
I don't think any of your changes are necessary. If you look at the Postfix documentation, you can see that the default values are already pretty similar to what you are suggesting:
```
default_process_limit = 100
smtpd_client_connection_count_limit = 50
smtpd_client_connection_rate_limit = 0
queue_minfree = 1.5*message_size_limit
header_size_limit = 102400
smtpd_recipient_limit = 1000
```
The biggest differences are the client connections (which we don't care about because we have a fail2ban-like authentication rate limiter) and the recipient list size (for which 100 may be too small for some people -- imagine sending out Christmas greetings to all your friends).

JOduMonT:
I forget about the fail2ban

but for the recipient_limit if think is valid
because if people use their postfix as a mailing list server they will have issue and will be considered as spammer.

ah and the header size still a good idea too ;)

mkuron:
Your header size is only a factor of two smaller than the default. So we don't gain that much in terms of security with the change, but risk problems in some rare edge cases where messages have lots of headers.

Limiting the number of recipients is a valid point, but 100 will occasionally be too low for some people. There is nothing wrong with sending messages to a larger number of people at once -- if it is original content sent to actual people's email addresses and doesn't happen regularly, it is quite unlikely that you'd be considered a spammer for that. I agree that the default of 1000 may be a bit high, but the Postfix developers chose it for a reason. It prevents excessive use without interfering with any legitimate use. Besides, anyone wanting to abuse this can just send the same email multiple times to different sets of recipients.

JOduMonT:
thanks you for you're explanation and probably research ;)
i'll go with your common sense then!
which make sense for me too!

PR #573, opened 2017-09-10
serve internet explorer js files too

cwildfoerster:
We're already serving all JS and CSS files, so we should serve the internet explorer specific fix files to, so we don't run into problems on local or corporate installtion without direct internet access.


andryyy:
Hm, or remove them altogether? I don't even know for sure we support IE < 9 right now.

cwildfoerster:
Well, if we're still using Bootstrap v3, then i would say yes, just keep them, they're not that big and annoying.

PR #577, opened 2017-09-11
Added getting current timezone via /etc/localtime

K2rool:
The Current version only gets the current timezone on Debain based distros, I've tested my edited version on Centos 7 and Arch Linux and it got the current system timezone no issue.

### DIFF
--- a/generate_config.sh
+++ b/generate_config.sh
@@ -16,7 +16,7 @@ if [ -z "$MAILCOW_HOSTNAME" ]; then
   read -p "Hostname (FQDN): " -ei "mx.example.org" MAILCOW_HOSTNAME
 fi
 
-[[ -a /etc/timezone ]] && TZ=$(cat /etc/timezone)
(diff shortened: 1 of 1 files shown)


andryyy:
Some systems don't have a realpath binary, so this command fails for Ubuntu 14.04 for example.

K2rool:
Hmm readlink Would work just as well that should be in Ubuntu 14.04, am petty sure Ubuntu 14.04 got /etc/timezone file anyway so the new added line is not going run anyway.

andryyy:
No, it throws an error when it doesn't exist.

andryyy:
readlink is part of the coreutils, isn't it? Have not seen a distro without it. :)

K2rool:
realpath is a part of coreutils too but was added later to it than readlink was.

K2rool:
Edit never mind did not notice youe merged it,

PR #607, opened 2017-09-15
[Feature] Add FTS using SOLR

### LINKED ISSUE #107: Feature request: Integration of solr in dovecot
Like in the non-dockerized version of mailcow it would be great to integrate solr plugin into dovecot for better search performance in SOGo. No problem to add it by myself, but would be great to have it as default. Or did you remove it due to performance issues concerning jetty & co?

evilstiefel:
This pull request adds support for full-text search using Apache SOLR as described here:

https://wiki.dovecot.org/Plugins/FTS/Solr

It adds another service to docker-compose which hosts an instance of Apache SOLR configured with a Dovecot schema for mail indexing. This enables fast searching through the inbox when using IMAP (and all web interfaces that use it, e.g. RoundCubeMail and SOGo).

The docker service has a build environment that was taken from **lmmdock/dovecot-solr**. Thus, _we can change the image source to his repository_. For completeness sake, I've included the entire Dockerfile-directory to locally build it.

This will increase the memory footprint and also adds another volume to the managed services. The user experience is quite nice though.

Fixes #107.


mkuron (changes requested):
Nice! Two major changes are needed though:
- Remove the solr_core directory. That stuff doesn't belong into the Mailcow repo. Instead, please modify the Dockerfile so it copies it from /opt/solr/example.
- The solr-vol-1 mount overlaps with the Docker image. You should probably mount it to /opt/solr/solr_core/data.

Furthermore, I have two questions:
- How big is the extra memory consumption? If it's major, then we probably won't want to enable SOLR by default, in which case the next question becomes important.
- What happens if SOLR is unreachable? Does Dovecot eventually give up contacting it or does it get stuck? What's the timeout for that?

evilstiefel:
Regarding your questions:

1. the extra memory consumption is considerable. Compare e.g. ~200MiB for mailcow-sogo and then around ~400MiB for the SOLR container. I don't know how it scales with a lot of users, but since SOLR is somewhat of an in-memory DB and runs in a JVM, there is some additional ressources required.

2. if the SOLR host is not reachable, Dovecot switches back to the default search behaviour within a couple of seconds (if even that much). If the host is up, but SOLR doesn't respond to requests correctly, dovecot does a fallback immediately.

I could remove the build-core files, but do keep in mind that those are not just copied from the example-folder, since the schema for the dovecot-core is different. Since it is derived from the docker-image lmmdock/dovecot-solr, we could remove the Dockerfiles completely.

andryyy:
Solr does indeed eat a bunch of memory. It will run in oom situations if not adjusted accordingly and also needs some further maintenance.
I have not added it for a good reason. :/

You won't need that old fixed scheme when running a current SOLR. It does auto create a dynamic scheme very well since a few months/years. That's default since v5 iirc. You would also need to disable this mode explicitly - haven't checked the configs tbh, but I guess you did that.
I also _think_ you used that old scheme you can find on the internet? It fails for some messages and has other flaws. :/ And, as I said previously, not needed anymore. 

Dovecot developed a commercial FTS plugin which they focus on since a long time. I think SOLR is a pita to maintain and will very often crash our cow. I stopped using it a longer time ago for these reasons.

evilstiefel:
The maintenance is minimal (really, just running some CURLs to commit and optimize the database from time to time). I did not create the schema, I just used the previouisly specified image and adjusted the mounts accordingly.

The reason why I invested some time into this: currently, searching for mails (especially body) is very slow. 10000 mails in my Inbox, a search without an index takes ~40 seconds. Most users would probably think it timed out at that time. With the index, the same quey takes maybe half a second.

I don't know about dovecot's commercial FTS plugin, and it's probably not applicable to this project anyway. I would also be interested in a less memory-hungry solution, but my only personal experience is with this tool, at least so far.

mkuron:
> I could remove the build-core files, but do keep in mind that those are not just copied from the example-folder, since the schema for the dovecot-core is different.

@evilstiefel, at least all of these language-specific word lists should not be in Mailcow. If they are not in the examples folder already, then you can use the Dockerfile to download them from somewhere. We shouldn't need to maintain separately. Only Mailcow-specific config files should be here.

> Since it is derived from the docker-image lmmdock/dovecot-solr, we could remove the Dockerfiles completely.

We do want to have the Dockerfile here so we can build updated images ourselves and don't depend on an image built by someone we don't know.

>  adjusted the mounts accordingly.

You still have a data directory with two files checked in -- I think you can remove them since you mount the volume there.

Dragnucs:
What about Elasticsearch. I don't know how it compares to Solr in terms of memory consumption but its maintenance is near zero.

mkuron:
@Dragnucs, [fts-elastocsearch](fts-elastocsearch) is broken with recent Dovecot versions. Fts-solr seems to at least be well-maintained as its an official plugin. There is an official plugin for Lucene, but Dovecot recommends to use SOLR instead.

evilstiefel:
> You still have a data directory with two files checked in -- I think you can remove them since you mount the volume there.

I chose the mount point because all cores are created there. If any user decides to create another one, they might get unexpected results when trying to persist the data. If we assume the dovecot-core is the only one and we don't care about the user "messing things up", I can adjust the mount.

I'll take a look into the added language files tomorrow.

mkuron:
@evilstiefel, it's the users's obligation to ensure correct mount points if he modifies the default setup. So solr-vol-1 should be mounted such that it only contains the actual data files and none of the configuration. Configuration files should be inside the container, except for ones that many users need to modify (which probably doesn't apply to SOLR, but to things like Postfix).

evilstiefel:
I've updated the pull request according to your comments. Now the only issue remaining is how to implement this "politically". Certainly, an option like with the ACME-client would make sense to not enable SOLR on memory-limited systems or for other reasons. I know that @andryyy is looking into this currently (at least what the maintainability of adding this PR to the project would be like).

data/Dockerfiles/solr/solr_core/conf/solrconfig.xml
```diff
@@ -0,0 +1,1482 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
mkuron: Does this file contain any customizations? I didn't find it referenced in the Dovecot FTS setup guides around the internet. If it's the default, maybe you can leave it out or copy from examples.

data/Dockerfiles/solr/solr_core/core.properties
```diff
@@ -0,0 +1,4 @@
+#Written by CorePropertiesLocator
```
mkuron: Looks auto-generated. Does it even need to be checked in?

docker-compose.yml
```diff
@@ -338,6 +338,21 @@ services:
         - /var/run/docker.sock:/var/run/docker.sock:ro
         - /lib/modules:/lib/modules:ro
 
+
+    solr:
+      build: ./data/Dockerfiles/solr
+      restart: always
+      volumes:
+        - solr-vol-1:/opt/solr/server/solr/dovecot
```
mkuron: You're adding files to /opt/solr/server/solr/dovecot in the Dockerfile and then mount a volume on top. That will cause trouble, at least when updating the docker container in the future. I think you should mount to /opt/solr/server/solr/dovecot/data so that the volume only contains data and no config.

Also, you still have data/index/{segments_1,write.lock} checked in. Those will almost certainly be created automatically and shouldn't be in git.

docker-compose.yml
```diff
@@ -338,6 +338,21 @@ services:
...
+      dns:
+        - 172.22.1.254
+      dns_search: mailcow-network
+      networks:
+        mailcow-network:
+          ipv4_address: 172.22.1.248
```
mkuron: You don't need to specify an explicit IP her. Automatic assignment is fine.

mkuron (changes requested):

evilstiefel:
The solrconfig.xml contains customizations, e.g. disabling soft autocommits.

Additionally, while the core.properties file might look auto-generated, it is necessary (because it only gets generated when manually adding a core). It is basically an index of available cores within SOLR.

I've remapped the volume and fixed a build-error in the Dockerfile (no brace expansion when not running bash, whoops).

data/Dockerfiles/solr/Dockerfile
```diff
@@ -2,12 +2,13 @@ FROM solr:latest
...
-RUN cp -R /opt/solr/example/files/conf/lang /opt/solr/server/solr/dovecot/ \
-    && cp /opt/solr/example/files/conf/{protwords.txt,stopwords.txt,synonyms.txt,currency.xml,elevate.xml} \
-       /opt/solr/server/solr/dovecot/
-
 USER root
 
+RUN cp -R /opt/solr/example/files/conf/lang /opt/solr/server/solr/dovecot/conf/
+
+RUN ["/bin/bash", "-c", "cp /opt/solr/example/files/conf/{protwords.txt,stopwords.txt,synonyms.txt,currency.xml,elevate.xml} /opt/solr/server/solr/dovecot/conf/"]
+RUN mkdir -p /opt/solr/server/solr/dovecot/data/index
```
mkuron: This can't work. You're mounting a volume to /opt/solr/server/solr/dovecot/data. If the _index_ directory needs to be created manually (and is not automatically created when SOLR starts up), you need to create an entrypoint script that does it.

mkuron (dismissed):
Looks pretty good now, though I have yet to test it and and you still have one mount overlap.

evilstiefel:
Fixed. It does automatically create the index.

extremeshok:
Any progress on this PR ?

hachre:
I tried this out but it doesn't seem to be working. Solr doesn't store any data in its assigned volume. I went into the Solr container and checked out the path where it should be and noticed that the `data` directory doesn't have `solr` assigned as its owner and group. Could this be the problem? 

```
solr@578b265ae38e:/opt/solr/server/solr/dovecot$ ls -la
total 20
drwxr-xr-x 1 solr solr 4096 Dec 25 09:07 .
drwxr-xr-x 1 solr solr 4096 Dec 25 08:41 ..
drwxr-xr-x 1 solr solr 4096 Dec 25 09:08 conf
-rw-r--r-- 1 solr solr   95 Dec 25 08:35 core.properties
drwxr-xr-x 2 root root 4096 Dec 25 08:40 data
```

However when I try to chown the data directory, I get:
```
chown: changing ownership of 'data': Operation not permitted
```

Braintelligence:
A full-text search would be gorgeous! Please implement this at least as an optional choice if it needs better hardware!

hachre:
I have now used `chown 8983:8983 /var/lib/docker/volumes/mailcowdockerized_solr-vol-1/_data/` in the host system (taking the uid from `cat /etc/passwd` within the container) to fix the permission for that folder and it seems to be working now. 

So this is something that definitely seems to need fixing.

hachre:
Picking up on the conversation here: https://github.com/mailcow/mailcow-dockerized/issues/722#issuecomment-353856962

Yes, I think with a feature like we already have for enabling / disabling clamav, this would be a very nice addition to mailcow!

andryyy:
Shouldn't we just use the official solr image, precreate the core and then insert the schema via API? 

There is an example for this: https://github.com/docker-solr/docker-solr-examples/tree/master/schema-api-startup
We wouldn't rely on third party images then. The "Tipps" section regarding hdr and df may be of interest, too: https://wiki.dovecot.org/Plugins/FTS/Solr

We could create a variable in mailcow.conf for the heap size.

hachre:
So about an hour later and it's done indexing my entire mailbox. The resulting index is 680 MB for a mailbox of 11.5 GB. When searching from within Roundcube, however, I see no speedup and the Dovecot log with `mail_debug = yes` shows me that it's still opening every single mail in the mailbox. Is this not supposed to work outside of the Inbox? 

Edit: My previous message (now deleted) about it working in SOGo was false. It doesn't work there either.

hachre:
It works correctly from within SOGo. No messages are opened and searches are very fast. I wonder what's wrong with my Roundcube setup...

hachre:
Okay, so after reading that link you posted, @andryyy, I came to the conclusion that I didn't run both commands needed to trigger a full manual indexing. I was missing the `doveadm fts` command. I now ran these and it has started to index more mails than before: 
```
docker-compose exec dovecot-mailcow doveadm fts rescan -A
docker-compose exec dovecot-mailcow doveadm index -A -q "*"
```

It did seem like 680 MB was somewhat small for a 11.5 GB mailbox. Will repeat my tests when this is done!

evilstiefel:
Good catch with the permissions. That's either something that changed with baseline or didn't occur during my tests (either because of my pre-existing setup of volumes or something else).

This already doesn't depend on external images except for solr:latest (which is the same image that is used in the "schema-api-startup" example).

That SOGo and Roundcube don't make use of the generated indices is something I cannot test on my setup, I used IMAP only and it worked flawlessly.

Technically, you shouldn't need to manually start an index, as they should be created when you first issue a IMAP SEARCH command. That of course means that the initial startup is slow, but following searches should then take advantage of the index. To make the user experience as nice as possible, I would advise the admin run `doveadm fts rescan` for all user accounts when first switching to FTS though. This doesn't matter for new setups, of course.

hachre:
@evilstiefel What client did you use to test it with? Wouldn't most non-webmail clients run the search in their own offline cache? I will try my tests with Thunderbird as well if Roundcube and SOGo fail again after this. I do remember there being an option in Thunderbird to explicitly have the search "run on the server".

evilstiefel:
It's true that a lot of standalone email clients will simply use the offline cache, precisely since not a lot of IMAP servers offer such functionality. I tested it by talking with the server directly to take that out of the picture. K9-mail for Android is an example where online searches are performed, and I do seem to recall that Thunderbird also had an option to perform the search online, I'd have to look it up again, though.

Edit:
According to Thunderbird's documentation about IMAP functionality [1], full text search is only available for locally stored and indexed messages. Otherwise, only the headers of each "partially synchronized" message are searched.

2nd edit:
Actually, in Thunderbird, if you right-click on a folder for an IMAP-account and select **Search Messages...**, you get an advanced search dialog in which you can also select to search online. Never knew about this, but may be that helps during testing:
online-search

[1] https://support.mozilla.org/en-US/kb/imap-synchronization#w_performance-and-search

hachre:
Thanks @evilstiefel. My new indexing round is done and the index is about 1.2 GB large now. However I'm still not seeing speedups in Roundcube and SOGo. I've now tried to execute a basic search with `openssl` on the command line and found the following error message in the Dovecot log when attempting my search (without `mail_debug = yes`):
```
Dec 25 14:06:45 orion dovecot: imap(user@example.com): Error: fts_solr: Invalid XML input at 1:0: not well-formed (invalid token) (near: {
dovecot-mailcow_1    | Dec 25 14:06:45 orion dovecot: imap: Error:   "responseHeader":{
dovecot-mailcow_1    | Dec 25 14:06:45 orion dovecot: imap: Error:     "status":0,
dovecot-mailcow_1    | Dec 25 14:06:45 orion dovecot: imap: Error:     "QTime":88,
dovecot-mailcow_1    | Dec 25 14:06:45 orion dovecot: imap: Error:     "params":{
dovecot-mailcow_1    | Dec 25 14:06:45 orion dovecot: imap: Error:       "q":"{!lucene q.op=AND}(hdr:bios OR body:bios)",
dovecot-mailcow_1    | Dec 25 14:06:45 orion dovecot: imap: Error:    )
```

The commands I used to attempt to search were these:
```
openssl s_client -connect example.com:993
a1 login user@example.com password
a2 select Archive/2008
a3 search text "bios"
a4 search text "bios"
```

When I run the `a3` line, that error message first appears in the log. Running `a3` and `a4` takes the same amount of time (about 12 seconds), so it looks like the cache is not being used.

When I run the search, I see the following almost instantly in the `solr` log, however:
```
solr_1               | 2017-12-25 14:32:56.170 INFO  (qtp257895351-16) [   x:dovecot] o.a.s.c.S.Request [dovecot]  webapp=/solr path=/select params={q={!lucene+q.op%3DAND}(hdr:bios+OR+body:bios)&fl=uid,score&sort=uid+asc&fq=%2Bbox:98d2af2b01163f5a36140000f49fd5c3+%2Buser:user@example.com&rows=2053} hits=5 status=0 QTime=3
```

Hits 5 is the amount of mails that have that word. So it seems the query reaches `solr` and also executes near instantly there but because of the error in the Dovecot log, Dovecot falls back to its direct search (which opens all mails in `Archive/2008` to find the text).

Any ideas?

hachre:
So I found out they switched the default response type from XML to JSON in a recent version, that's why that happens. See: https://issues.apache.org/jira/browse/SOLR-10494

The fix is to add the following line to the `solrconfig.xml` file in one specific area:
```
<str name="wt">xml</str>
```

See here:
https://github.com/lmm-git/docker-dovecot-solr/pull/2/files#diff-7f9d451cf2754ca8ba70dee3682255bb
and here:
https://github.com/lmm-git/docker-dovecot-solr/issues/1

I'm going to change it and try soon.

hachre:
I have attempted to change the file but despite seeing it changed within the instance, it is still sending JSON output, so I'm not sure what else is missing here.

hachre:
### It works!

So what needs to be changed is edit the file `data/Dockerfiles/solr/solr_core/conf/solrconfig.xml` and change the following area:
```
  <requestHandler name="/select" class="solr.SearchHandler">
    <!-- default values for query parameters can be specified, these
         will be overridden by parameters in the request
      -->
    <lst name="defaults">
      <str name="echoParams">explicit</str>
      <int name="rows">10</int>
      <!-- <str name="df">text</str> -->
    </lst>
```
to look like this:
```
  <requestHandler name="/select" class="solr.SearchHandler">
    <!-- default values for query parameters can be specified, these
         will be overridden by parameters in the request
      -->
    <lst name="defaults">
      <str name="echoParams">explicit</str>
      <int name="rows">10</int>
      <str name="wt">xml</str>
      <!-- <str name="df">text</str> -->
    </lst>
```

Don't change the other occurrence that already exists in the file and says `<str name="wt">json</str>`.

#### So in summary, we need the following changes to make it work:
1. fix the permission problem of the `data` dir (it had `root:root` permissions, instead of `solr:solr`)
2. change the `solrconfig.xml` file as described

Could you do these in your pull request, @evilstiefel?

Thank you!

I kept an eye on memory consumption and it is between 1 and 3 GB while very busy indexing, so it is quite substantial. We will definitely need an on/off toggle for this whole feature. But the results are great: It is working with both Roundcube and SOGo and searches for mail contents are taking less than 2 seconds now across all folders of my 200k mailbox. Great!!!

evilstiefel:
I'd say, contributions welcome - and thanks for your findings. I can take a look and incorporate the changes myself in two days when I get back home to my computer.

hachre:
I have next to no Docker experience (just started using it for the first time a few days ago when I started checking out mailcow), so I feel less than comfortable attempting these changes myself :)

evilstiefel:
That's fine, I'll also include a change so that we can have a stable SOLR version and not be surprised by further upstream decisions and make our own deliberate changes when it comes to upgrading SOLR.

mkuron:
> I'll also include a change so that we can have a stable SOLR version

Make sure the version you choose (e.g. solr:7 or solr:7.2) will be supported for some time to come and don’t pin to a specific version like solr:7.2.0. we still want to pick up security updates and bugfixes automatically.

Braintelligence:
So excited about this feature!

evilstiefel:
I've commited the changes, pre-creating the solr-core data folder with correct permissions, changing the configuration for the return format (xml) and locked solr to version 7.2.x (so solr:7.2). We could argue about 7 in general, but I thought 7.2 and all minor fixes was a sensible choice.

This does not yet include
- an on/off config option
- merging the latest upstream changes (even though there are no conflicts)

mkuron:
> We could argue about 7 in general, but I thought 7.2 and all minor fixes was a sensible choice.

Could you check what the support policy on solr is? It looks like they never released a 7.1.1 or higher before 7.2.0 came out, but 5.5 and 6.6 are still getting updates. So locking to 7.2 may mean we will never get an update. Judging from https://lucene.apache.org/solr/guide/7_0/solr-upgrade-notes.html, there shouldn't be backwards-incompatible changes within a major version, so _solr:7_ should be fine and give us updates for a few years.

hachre:
From http://lucene.apache.org/solr/community.html:

> ### About versions
> Apache Solr is under active development with frequent feature releases on the current major version. The previous major version may still receive security- and bug fixes as point releases, thus taking the role of the LTS (Long Term Support) version. Older versions are considered EOL (End Of Life) and will not be further updated. For this reason it may also be difficult to obtain community support for EOL versions.
> 
> Large changes or changes that break compatibility with existing functionality are normally only included in the next major version.

They consider 7.x the current STABLE branch, which will receive features but not break compatibility.
6.4.x is the LTS branch, which will receive only bugfixes.
Once a 8.0 comes out, the then current 7.x version will move into LTS mode and the 6.4.x will move into obsolete mode.

So, yes, it appears to me as well that 7.x is the better choice than 7.2.

evilstiefel:
There doesn't appear to be a "roadmap" for future releases. It looks like 6.x was supported for short on 2 years, as was version 5. @hachre sniped me with a similar observation, so I think we can use solr:7 ;)

data/Dockerfiles/solr/solr_core/conf/solrconfig.xml
andryyy: Version does not match, pathes could be wrong etc. I don't think it is a good idea to copy a core config as xml. We should just use the API to create the core and its config...

andryyy:
Please don't just copy the other images' files. Use the API (see examples posted earlier). We don't even know wether this core config is any good. Is it even compatible with newer versions? There are pathes which could be wrong, too.

What about that select handler change in the Dovecot wiki?

The curl cronjobs will exit with errors unless we catch the request and return dummy data or something like that. How does Dovecot handle a failing FTS plugin? Do we even need to commit by curl, when Solr has an auto-commit feature? Some important Solr options (heap size etc.) must be configurable in mailcow.conf. A docker entrypoint script can use the API to configure these parameters on startup.

evilstiefel:
We can think about adding the core programmatically to try and get rid of the solrconfig.xml (although we would still need to somehow include the change for the **/select** path to still return XML and not json).
I will not invest time to use the api (which really is just sending curl POST-requests to localhost) to recreate the schema that is provided by the official dovecot git-repository [1]. I don't even know if it's feasable to not just create fields (like in the example), but also the other more advanced schema options. In this case, it's not just any file from an image, but a copy of a file from dovecot's documentation.

[1] https://raw.githubusercontent.com/dovecot/core/master/doc/solr-schema.xml

hachre:
I think the schema is fine to be included as a file, isn't it? The concern is about the `solarconfig.xml` I believe.

mkuron:
The schema is not that much of a concern because it is provided by the Dovecot developers. You shouldn’t commit the file though but retrieve it from the Dovecot repo instead from within the Dockerfile — the `ADD` command can deal with remote files.

In the config file, we need to understand what the individual options mean and write a file that is as minimalistic as possible. And things like the heap size need to be configurable via environment variables.

evilstiefel:
We can definitely minify some config contents to the absolutely necessary, which I think is a good idea. I don't necessarily agree with the idea to add a dependency from a remote repository to a file that should never be changed outside of our control (since the schema depends on the dovecot version, and which fields it expects).

For that reason, I would advocate against linking ADD during build to a remote Github repository, be it only because it adds a build-dependency for a network resource that needs to be available. Of couse, users shouldn't need to build often and Github should always be available. Still, the file doesn't need to be "freshly downloaded" and with the case of the incompatibility that arose from the recent Solr update, I think it's a valid argument.

mkuron:
> be it only because it adds a build-dependency for a network resource that needs to be available

We already depend on dovecot.org being available to build the dovecot-mailcow image and github.com for sogo-mailcow, so don't worry about that.

> Still, the file doesn't need to be "freshly downloaded"

There is just no need to have it in the Mailcow repo. We don't have any default config files here for all the other components, just custom stuff and files with major modifications. All the default configs come from the source tarballs at docker-build time.

> since the schema depends on the dovecot version

You can put the version branch into the URL: https://raw.githubusercontent.com/dovecot/core/master-2.2/doc/solr-schema.xml

Braintelligence:
@mkuron Do you think we'll achieve implementation of this in Q1 2018? 😸

andryyy:
No, since this is just copied from here and there.

Just use the offical images, don't copy unnecessary stuff from other images and use the API to create the core and the schema. That's the supported way described in the Solr images. It is also not usable for everyone. Solr does heavily depend on your setup and load. It needs more options exposed to mailcow.conf.

Plus: Solr changes a lot, the API calls make sure we don't copy crappy configs or configs we don't even fully understand and hope it just works.

Braintelligence:
@andryyy Sounds like you don't support this PR. Maybe SOLR will be integrated by one of the maintainers in the future?

andryyy:
I support this as long as we use solr as it is described with examples in their repo.

mkuron:
@Braintelligence, I fully agree with @andryyy on this. If the SOLR Docker maintainers say we should use the API, we should do that. It would eliminate the config file we don’t completely understand and ensure we don‘t use any outdated config parameters. The schema should still come from Dovecot‘s repository, but not be checked into our repository.

Braintelligence:
@mkuron I completely trust your decisions guys 👍.

lavdnone:
thank you, tested it - works fine

- as per dovecot recomendation https://wiki.dovecot.org/Plugins/FTS/Solr
they suggest using schema.xml they provide in /usr/local/share/doc/dovecot/solr-schema.xml
only difference are some types like tdates tlongs - date search works without full-text index and we would hardly utilize solr.TrieDateField format
```
    <fieldType name="booleans" class="solr.BoolField" sortMissingLast="true" multiValued="true"/>
    <fieldType name="tdates" class="solr.TrieDateField" docValues="true" precisionStep="6" positionIncrementGap="0" multiValued="true"/>
    <fieldType name="tlongs" class="solr.TrieLongField" docValues="true" precisionStep="8" positionIncrementGap="0" multiValued="true"/>
    <fieldType name="tdoubles" class="solr.TrieDoubleField" docValues="true" precisionStep="8" positionIncrementGap="0" multiValued="true"/>
```

- params.json and solrconfig.xml are two ways to set same settings, and can be merged, like for example Velocity plugin
   https://lucene.apache.org/solr/guide/6_6/velocity-response-writer.html
- solrconfig.xml adding `<str name="wt">xml</str> `is because of Velocity plug-in activation, which is fine
- those curl commits through cron are he same way API would do configuration, softCommit can be replaced in solrconfig.xml  to cover it: 
```
<autoCommit>
  <maxTime>300000</maxTime>
</autoCommit>
```
- maybe put out autoCommit interval in to admin panel - it is thousands of a seconds after last index update.
- heavy curl update?optimize is good in cron but maybe with time setup from admin panel and with optimize now button
- Suggested using active api is for different purpose of scalability and dynamic swamps. We want static docker-compose so either json or solrconfig.xml should be good
- core.properties gives it a name - useful
- Docker images and compose seem to be good - nothing to take away - all use internal provided files

- only file needed outside is solrconfig.xml for people to add lines depending on file locks if they need any etc

extremeshok:
Excellent work so far,
Any progress or eta on this for a merger ?

evilstiefel:
I've read @lavdnone's remarks and have not forgotten about this - but I lack the time right now to start learning and playing around with the API SOLR provides to create files programmatically. While I understand the project owner's concerns about the implementation, they are no easy fixes for me right now.

Also, having things configurable through the admin panel is outside my expertise since it belongs to an entirely different component with which I am not involved. I can certainly see myself spend more time on the docker-part/configuration in the coming weeks, so don't give up yet ;).

andryyy:
The API will help us to not break things over time when Solr changes. It is not only for dynamic swamps or something @lavdnone wrote.

It is fatal to always assume your understanding is the correct understanding. You can always suggest changes and explain why you _think_ they are necessary or correct. Don’t write they recommend using the xml when you don’t seem to understand why the API should be preferred (catch field type errors or field type changes, not overwriting/removing useful new options, config validation and more). I know this PR will work as it is. But we shouldn’t copy files from other Solr images and break future builds due to changes in Solr. It is even recommended by Solr to use the API.

Whatever. I really appreciate this PR and thank evilstiefel for sharing this! It will work as it is right now, but let’s try to implement it using the API. We can also make use of it in the mailcow UI.

evilstiefel:
Don't get me wrong, I never assumed my understanding is perfect or the correct one, neither did I say that anyone from SOLR suggests we use the XML as is :). I do understand why using the API is preferable, since it's a layer of abstraction that we then don't have to worry about anymore (assuming the API is stable).

The only thing I did say is that right now, I don't have a lot of free time to wrap my head around the issue, but that I've not forgotten about it.

The only thing I can't do is change the web interface to make it configurable. As always, thanks for your input!

andryyy:
Sorry @evilstiefel, I was not directing this to you. :-)

lavdnone:
In previous post I went through the changed files to share what I know about those and provide sources which I used for deployment - not sure if it is optimal or fastest way, again just info out from deployment experience

Braintelligence:
Do I understand correctly, that implementing #1007 would make SOLR obsolete?

evilstiefel:
I don't think it does, unless dbox and other formats also magically index message contents and make those available to dovecot (and not just mail flags). Nothing like that was mentioned in that issue though.

extremeshok:
@Braintelligence 
> Do I understand correctly, that implementing #1007 would make SOLR obsolete?

No it does NOT. mdbox speeds up the actual message indexing and processing, not full text searching

fts_lucene with the new/modern lucene++ can replace the need for solr. This is how my current setup is.

Braintelligence:
@evilstiefel @extremeshok Thanks for clarifying! Is fts_lucene easier to set up than SOLR?

evilstiefel:
@Braintelligence you can read about it in the [Dovecot wiki](https://wiki2.dovecot.org/Plugins/FTS/Lucene). They state themselves that they would choose SOLR over Lucene because of dependency on an outdated library, but that is a seperate discussion and shouldn't be done in this PR.

extremeshok:
@evilstiefel the "outdated" library was replaced / superseded with this https://github.com/luceneplusplus/LucenePlusPlus , its maintained by "spotify.com"

Ill push a patch later this week with lucene support

andryyy:
They also have their own (paid) module now.

lavdnone:
This solr setup seems very stable. Full text search works like magic. Resource consumption minimal.
Initially was looking to use elasticsearch as I use elasticsearch cluster for other stuff, but eventually gave up as I was unable to break or find flaw in mentioned here solr setup.
Another reason for this to merge.

andryyy:
Ressource consumption does depend on usage. You cannot generalize Solr setups. That’s not a reason for a merge.

evilstiefel:
I'm working on using the API to create a core. Currently, I'm a bit hampered by the fact that there is no API method to change the schemaFactory (since I don't want a managed schema, but use the schema.xml provided by dovecot). So, this is two steps forward, one step back (doesn't currently work as-is).

evilstiefel:
I've added the necessary changes so that we can use the managed_schema by recreating the dovecot schema using api-calls. This eliminates all config files and instead uses a startup-script that I borrowed from @andryyy's link from way earlier which I modified to suit our needs.

Right now it is only run when building the image and is not a new entrypoint. Maybe this has to change?

extremeshok:
@lavdnone 
> . Resource consumption minimal.

Have you tested with more than 10 million emails and rebuilding indexes ?? We went back to fts_lucene with luceneplusplus because of this

lavdnone:
@extremeshok not with 10 mil, no. I was actually waiting to see if you tested it with your email pressure.
My understanding was that dovecot recommended solr https://wiki2.dovecot.org/Plugins/FTS/Lucene
Solr and Elasticsearch use Apache Lucene shards which, I assumed came from luceneplusplus code.
Calculated index size to be around 1/50 and shot it to SSD tire.

Fts through Luceneplusplus library shooting index and searches directly from files? without Apache Lucene server?
Is solr:7 docker resourse hungry on 1000 emails a minute flow? or is it dovecot fts for solr freaking out?

Thanks for info in advance.

andryyy:
Thanks a bunch @evilstiefel, that's the way we should use it. I will add some more options to the UI etc.

mkuron:
This needs a config variable to disable as its memory requirements can be huge. Also, the RAM check from #1135 needs to set that variable appropriately (e.g. enable ClamAV if 1.5 GB available and SOLR if 4 GB available).

andryyy:
Sure, I am working on it. 👍 It will be disabled by default.

tramtrist:
I updated an existing stable environment to dev and am getting errors with solr.  Really want to get indexed search working as I use it all the time.  Also using roundcube.  Any advice?
: Error: fts_solr: Indexing failed: connect(172.22.1.2:8983) failed: Connection refused (2 attempts in 0.199 secs)

tramtrist:
Well I was able to get solr started but its throwing an error about not having a dovecot container.. 
org.apache.solr.common.SolrException: Could not find collection : dovecot
Anyone out there? @evilstiefel @andryyy

evilstiefel:
Hey @tramtrist, I saw your original question, and so far I can only say that "it works" with the commit that I checked in. I know that is super unsatisfactory, but I don't know what changed upstream in the dev-branch/solr after it was merged, but @andryyy did say on IRC that he is working on merging it into master with the remaining proper changes (like making it configurable through the config).

Unfortunatelly, I cannot investigate the issue right now.

andryyy:
Dev is pretty unstable right now. Solr works fine, but there are probably a thousand over things that can break Solr today. 

I will merge it asap, but I’m on vacation.

Only a very few more options like those @evilstiefel mentioned need to be integrated.

tramtrist:
Thanks guys :dancer:

tramtrist:
@evilstiefel @andryyy You're right.. It does work.. All you need to do is run the right start command in the solr container
./solr start -m 1g -d /opt/solr/server/ 
..Did the trick
I'll keep an eye out for the merge.. but for now I can search to my heart's content!

BenSchweikert:
Hi,
any progess here? I am waiting for this feature for month.

andryyy:
nope

tmikaeld:
I just found Mailcow and it's the perfect, it's almost the same stack we use manually!

However, Solr support is a **requirement** for us. 

And yes, Solr has to be disabled by default, because it can easily chew through 4GB of RAM even for 10k e-mails and i don't think auto-enabling it is a good idea.

Memory Heap Settings should also be configurable since it's not possible to assume much about index size and usage needs.

We're currently running Solr on 4 VM's on 5M e-mails, 25 users, taking up 300GB of space, solr consume 4GB of RAM in total.

We use the default Dovecot schema @ /usr/share/dovecot/solr-schema.xml

I'm really hoping that this can be merged officially soon!

zeigerpuppy:
This looks like a reasonably safe addition from the changes shown.  I agree it's probably best not to enable in the default config due to memory consumption but I'm wondering what the recommended way is to test it with an install based on the stable (master) branch?

- Is it recommended to just apply the changes here as a patch?
- Or is it better to use the dev branch?

andryyy:
Don't use dev, please. :) It is a graveyard at the moment. I will include it soon, if you guys really want it.

zeigerpuppy:
Ah great, including FTS in master would be awesome, it makes a huge difference for user searches.  Really enjoying using mailcow (donation on the way soon!).

extremeshok:
@andryyy please add support for both fts_lucene and fts_solr , disabled by default

I can submit patches for fts_lucene if required

tribut:
I agree, `fts_lucene` provides the same benefit to the user that solr does for small installations while using (almost) no additional memory when idle. It is disconcerting that the library is somewhat unmaintained, but I guess it can be an acceptable trade-off.

andryyy:
I will build it with lucene. 

No additional memory besides all the data that needs to be stored in memory for FTS, you mean?

No additional memory usage plus an indexing service is not possible. They always eat a lot of RAM. That’s its purpose. :-D

tribut:
Yes, `fts_lucene` does appear to hold any fts data in memory. It reads the index from disk for every query, which is far less memory intensive than running the solr daemon.

extremeshok:
Compared to running a dedicated SOLR and the resources it eats, we talking a few MB compared to GB's. SOLR server also requires massive CPU resources during the hard commit and rebuilds.

This was my original guide on using fts_lucene

https://extremeshok.com/6821/dovecot-fts_lucene-lucene-full-text-search-aka-solr-replacement-for-centos/

andryyy:
Let’s check lucene then. I will give it a try locally. 

"It's a much better idea to use fts-solr instead, which has much more features and is more stable."

So no.

zeigerpuppy:
I found some useful resources on this choice:
- http://www.lucenetutorial.com/lucene-vs-solr.html
- https://help.atmail.com/hc/en-us/articles/201566404-Installing-Apache-Solr-with-Dovecot-for-fulltext-search-results

and from the Dovecot docs:
- https://wiki2.dovecot.org/Plugins/FTS/Solr
- https://wiki2.dovecot.org/Plugins/FTS/Lucene

The dovecot authors definitely [recommend Solr]( https://wiki2.dovecot.org/Plugins/FTS/Lucene): 
 
> NOTE: Although the fts-lucene plugin works, it's using CLucene library, which is very old and has some bugs. It's a much better idea to use fts-solr instead, which has much more features and is more stable. 

I have installed Solr before and it was moderately resource hungry but once the indexes are built it only indexes new mail, so the CPU utilisation is generally moderate as an ongoing issue.  It also has much more advanced features, like defining the phrase length for searches and easier scripting of updates. 



I'd put my vote behind solr but I am probably less CPU/memory constrained than some users.  Another consideration is long-term maintenance of the FTS solution, Solr clearly leads in this respect.

extremeshok:
Majority of the mailcow user base would be happy with quicker searching and almost no resource impact aka lucene.

My suggestion was to add compiled support for both, to the dovecot image, and the user can choose what they want to use.

lavdnone:
Guess adding doveadm fts rescan to mailcow admin web interface would also make sense to switch or flush and reindex users

andryyy:
Indeed

PR #627, opened 2017-09-21
rspamd: move all config to local.d so override.d can be used for customizations

mkuron:
Rspamd has multiple levels of configuration: /etc/rspamd, /etc/rspamd/local.d and /etc/rspamd/override.d. The first one contains the default configuration, and the second and third override the default settings (in that order). For details, see https://github.com/vstakhov/rspamd/blob/master/conf/rspamd.conf. The main difference between levels two and three is that local.d gets merged with and override.d replaces existing options, but that doesn't make a difference for the settings that Mailcow uses.

Previously Mailcow was using both local.d and override.d, but this is not necessary. This PR moves all configuration to local.d so that override.d can be used by the user for his own customizations. This prevents merge conflicts when a Mailcow update and the user both change the same file. Now user configuration in override.d takes absolute precedence.

I tested this PR by changing the web UI password and the debug log level from within override.d and everything worked as before. This PR should thus not introduce any changes visible to the user.


andryyy:
Moving to local.d will merge existing config options - or even new options! - we don't want to use.

It is not a "either this or that". You _will_ run into problems. You would need a configdump for each change to see what was merged and/or - even better - check the modules LUA for options that will be switched on. Configdump may not be enough.

I don't see a need for that PR. I personally never run into any merge conflict with Rspamd configs. We already use a gitignore to ignore all untracked files.

I promise this PR would turn into a problem - sooner or later. :/

mkuron:
> Moving to local.d will merge existing config options - or even new options! - we don't want to use.

I don't get it. Anything that's in the defaults will get used and we overrode it either in local.d or override.d if we needed to. The lua modules don't even know that there are separate config files at the three levels. Have a look at https://github.com/vstakhov/rspamd/blob/master/conf/rspamd.conf -- all four files that we used to have in override.d are read from there.

> I don't see a need for that PR. I personally never run into any merge conflict with Rspamd configs.

I had edits in the files in override.d and ran into conflicts previously.

mkuron:
> Moving to local.d will merge existing config options - or even new options! - we don't want to use.

New options (from /etc/rspamd) are always used as long as we don't override them, so I don't get your point.

> check the modules LUA for options that will be switched on

All config files that we had in override.d are used by https://github.com/vstakhov/rspamd/blob/master/conf/rspamd.conf and not by any Lua modules. So the only thing that even knows that there are multiple files is the config file parser, the modules don't care.

> I don't see a need for that PR. I personally never run into any merge conflict with Rspamd configs.

I had edits in the files in override.d and ran into conflicts previously.

> I promise this PR would turn into a problem - sooner or later. :/

The only thing that changes is the merge behavior. Root-level options are not affected as they always get overridden. The only options that are affected are ones that are mergeable -- and these are only the `secure_ip` option (where it doesn't matter) and the `upstream` block (where we override all parameters that are set in the default config).

andryyy:
No, override.d does not merge any default options. All other options are disabled. Whatever you write to override.d is all that module/worker will read.
When you, for example, move the proxy worker to local.d, the defaults in /etc/rspamd/worker-proxy.inc will be merged, while they are ignored right now. 
Please just leave this closed, these changes will cause trouble sometime...

mkuron:
What you're saying doesn't match the FAQ: https://rspamd.com/doc/faq.html#what-are-the-locald-and-overrided-directories . Override.d only overrides top-level entries, so e.g. our worker-proxy.inc inherits the `timeout = 120s;` from the default config.

mkuron:
So I've diffed the config dump and see that the only difference is in the `upstream` section of worker-proxy. This is incorrectly merged, but it makes no difference.

```diff
         upstream {
+            local {
+                default = true;
+                hosts = "localhost";
+            }
             name = "localhost";
             default = true;
             hosts = "rspamd:11333";
         }
```

I have now submitted a new PR that uses optional include files instead to achieve the same goal. That way, the files that are currently in override.d can stay there if you think that is safer and at the same time the user can put his customizations into separate files.

andryyy:
There have been other PRs that had flaws where everybody assured they are totally fine. I don’t want you to understand (I really don’t want to explain every time why XY will get us into trouble), but please respect this decision. It is not okay to move these files from here to there. The merge behavior _is_ different and I _want_ to override some changes - also in favor for future Rspamd releases.

Still, as I said before: sometimes a config dump is not enough to find the changes. I had modules act differently because some options where not explicitly empty or whatever. 

The update script _should_ solve most conflicts in favor for mailcow.

Other people may have configured mailcow in a way that _this_ PR will give them merge conflicts.

mkuron:
Alright, then have a look at #629. That changes nothing about existing behavior and only includes some new files if they are there.

> The update script _should_ solve most conflicts in favor for mailcow.

Many, but by far not all. Merging files is not an easy task for a computer. That's why one ideally has user-defined configuration in separate files.

PR #629, opened 2017-09-22
rspamd: Include files for customizations in override.d

mkuron:
Supersedes #627. While that had no adverse effects on behavior, @andryyy didn't like that there is a slight danger of behavior changing in the future.
I'm now including an extra file (*.custom.inc) if present into each file inside override.d which can be used for customizations. That way, none of the Git-managed files need to be changed anymore for customizations.

I've also removed `enable_password` from the worker-controller config as that is already set to a sensible default by rspamd (`q2`, and that is blacklisted by the web UI so you can't actually log in without changing it first). I don't think it's documented anywhere what previous default hash actually is, so this change also ensures that nobody can get into an unconfigured web UI by brute-forcing that hash.


andryyy:
"I don't think it's documented anywhere what previous default hash actually is" -> Some random string I don't know. :-) But it is much better this way, I agree.

I am not sure about what happens when we merge this. Will everyone run into a merge conflict? Have you tried?

mkuron:
If they‘ve changed the password line, they will of course get a merge conflict in that line, there‘s no way around that. update.sh merges with `-Xtheirs`, so that should be silently resolved. Those who update manually need to remove the line manually. In either case, they need to manually set the password again by putting it into the new included file.

I guess the situation is no different with any other change to the rspamd config — if someone changed the line we touch (or one close by), they get a conflict. The good thing about this PR is that this will be the last time ever this happens because customizations can now go into separate files.

mkuron:
Fixed in #814

Issue #615, opened 2017-09-18
troubles with autodiscover with the new outlook 2016 assistant


mritzmann:
Outlook 2016 has problems with the autodiscover function of Mailcow. The old setup wizard works, the new one does not. Anybody got the same problem?

---

new wizard:

![new_assistant](https://user-images.githubusercontent.com/4533780/30533220-881337e0-9c58-11e7-8ae9-75c7f8a375aa.png)

old wizard:

![old_assistant](https://user-images.githubusercontent.com/4533780/30533229-911aa01c-9c58-11e7-9e5e-06388f223f98.png)

The configuration looks valid to me:

![test](https://user-images.githubusercontent.com/4533780/30533245-a550091e-9c58-11e7-9c86-e588415fbf13.png)

mkuron:
Please use the old one for now. Outlook 2016 for Mac has always shown the same issue (it uses a new JSON-based autodiscover mechanism that is not documented by Microsoft and not supported by us), so I suspect they ported that to the Windows version now.

andryyy:
Sounds fun.

Would love to wireshark that communication.

mritzmann:
If anyone is interested in debugging: I can donate an Office365 account with Outlook 2016.

ChessSpider:
quick way to open old style:
in RUN
```C:\Program Files (x86)\Microsoft Office\root\Office16\OLCFG.EXE```
then button ```email accounts``` => ```manual setup or additional server types``` => **Exchange ActiveSync**

I had to look for this, as autoconfigure on Outlook 2016 now configures as SMTP/IMAP :(

andryyy:
@mkuron Any idea why it configures as IMAP?

mritzmann:
> I had to look for this, as autoconfigure on Outlook 2016 now configures as SMTP/IMAP

Tested it yesterday with the old setup-assistant. Account was set up as ActiveSync. Only the new setup-assistant doesn't work.

andryyy:
Strange. But it auto-configured it as IMAP?

mritzmann:
automatically as activesync

mkuron:
I'll test and fix it later this week. The IMAP thing is probably a user agent issue again.

andryyy:
@mritzmann I meant the new config guide. :) Thanks mkuron! :)

mkuron:
[...]

PR #636, opened 2017-09-26
Preliminary support for Outlook 2016’s autodiscover.json

mkuron:
This adds preliminary support to Outlook 2016's JSON-based autodiscovery mechanism.

My PHP script behaves exactly like Microsoft Exchange does as far as I can tell, though unfortunately Outlook doesn't pick up the configuration correctly. Until Microsoft publishes documentation on the file format or someone reverse engineers all the remaining protocol details, we will not be able to provide full autodiscovery support for Outlook 2016.

Partial fix for #615

### DIFF
--- a/data/conf/nginx/site.conf
+++ b/data/conf/nginx/site.conf
@@ -86,6 +86,14 @@ server {
     try_files /autodiscover.php =404;
   }
 
+  location ~* ^/Autodiscover/Autodiscover.json {
+    fastcgi_split_path_info ^(.+\.php)(/.+)$;
+    fastcgi_pass phpfpm:9000;
+    include /etc/nginx/fastcgi_params;
+    fastcgi_param SCRIPT_FILENAME $document_root$fastcgi_script_name;
+    try_files /autodiscover-json.php =404;
+  }
+
   location ~ /(?:m|M)ail/(?:c|C)onfig-v1.1.xml {
     fastcgi_split_path_info ^(.+\.php)(/.+)$;
     fastcgi_pass phpfpm:9000;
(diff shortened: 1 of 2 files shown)


andryyy:
Nice. Should we merge this? Or wait?

andryyy:
Outlook for mobile auto-discovering works with this.

mkuron:
Oh, it works for you? Then I guess you can merge it. Can you test the desktop version too?

ChessSpider:
I think autodiscover for mobile was always working ? i think this issue is for outlook specifically?

andryyy:
Outlook for iOS fails without this.

andryyy:
I did some more research on this...
I used to use Outlook 2016, not Outlook from Office 365. I decided to try O365 for a month and, well, there it is: the "new" dialog. Sadly autodiscover.json does not work here. Well, it kind of does. It queries the json file, then the xml, which in the end returns the correct XML stuff. But it looks like MS doesn't want ActiveSync in Office 365's Outlook.
Returned to Outlook 2016 retail and everything is fine.
Outlook for iOS auto-configures fine with EAS when this .json file is used.

mkuron:
Microsoft is not making it easy for us... I had tried Outlook 2016 retail and it showed the new dialog...

But since this PR appears to improve things in at least a few situations that didn’t work previously, let‘s merge it.

andryyy:
grafik

Oh, PP. :P 

Okay, sure, let's merge it.

andryyy:
I have just tested it again with full debugging... the iOS mobile app definitely _only_ calls the json autodiscover file and configures fine! Crazy.

andryyy:
@mkuron Have you tried removing ActiveSync from protocols and returning only "Autodiscoverv1" as supported protocol? Does it fall back to this? I just removed Outlook 365. :-/

edit: this is most likely an Office 365 thing. I installed another plain Outlook 2016 (not from any Office 365 package they sell on eBay) and it worked ootb.

mkuron:
No, it only ever requests `ActiveSync`.

PR #638, opened 2017-09-27
ACME needs to wait for MySQL to be ready

mkuron:
Today, I installed a Docker update on my server and it restarted all containers. Unfortunately, the acme container came up before the mysql container and thus all my autoconfig.* and autodiscover.* SANs were removed from the certificate.
```
acme-mailcow_1       | Found Let's Encrypt or mailcow snake-oil CA issued certificate with SANs: autoconfig.example.com autodiscover.example.com cloud.example.com mailcow.example.com
acme-mailcow_1       | ERROR 2003 (HY000): Can't connect to MySQL server on 'mysql-mailcow' (111 "Connection refused")
acme-mailcow_1       | ERROR 2003 (HY000): Can't connect to MySQL server on 'mysql-mailcow' (111 "Connection refused")
acme-mailcow_1       | Found A record for mailcow.example.com: 192.0.2.42
acme-mailcow_1       | Confirmed A record mailcow.example.com
acme-mailcow_1       | Found A record for cloud.example.com: 192.0.2.42
acme-mailcow_1       | Confirmed A record cloud.example.com
acme-mailcow_1       | Found orphaned SAN autoconfig.example.com autodiscover.example.com in certificate, moving old files to /var/lib/acme/acme/private/2017-09-27_04_40_20.bak/, keeping key file...
```

This PR fixes this race condition by making ACME wait for MySQL to come up.

### DIFF
--- a/data/Dockerfiles/acme/docker-entrypoint.sh
+++ b/data/Dockerfiles/acme/docker-entrypoint.sh
@@ -80,6 +80,11 @@ else
 	fi
 fi
 
+while ! mysqladmin ping --host mysql -u${DBUSER} -p${DBPASS} --silent; do
+	echo "Waiting for database to come up..."
+	sleep 2
+done
+
 while true; do
 	if [[ "${SKIP_LETS_ENCRYPT}" =~ ^([yY][eE][sS]|[yY])+$ ]]; then
 		echo "SKIP_LETS_ENCRYPT=y, skipping Let's Encrypt..."
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -249,7 +249,8 @@ services:
     acme-mailcow:
       depends_on:
         - nginx-mailcow
-      image: mailcow/acme:1.19
+        - mysql-mailcow
+      image: mailcow/acme:1.20
       build: ./data/Dockerfiles/acme
       init: true
       dns:


chriscroome:
> Today, I installed a Docker update on my server and it restarted all containers.

Would it be worth adding something to the documentation to make it clear that upgrading Docker CE will cause this behavior? (With Xen you can upgrade it without VMs restarting.)

Docker CE updates are quite frequent (I think the site says they aim at every three months but it seem to be more often than that).

Also does it make sense to manually shutdown the containers, upgrade and then restart them, for example:

```
docker-compose down
apt update ; apt upgrade
docker-compose up -d
```

mkuron:
> Would it be worth adding something to the documentation to make it clear that upgrading Docker CE will cause this behavior?

Restaring the Docker daemon isn‘t the only thing that could have caused this. Rebooting the host would have too, and that would have affected Xen too.

There is no point in documenting this behavior as it is wrong and the pull request fixes it...

> Also does it make sense to manually shutdown the containers, upgrade and then restart them

You can do that, but it is not necessary. All Mailcow containers come up just fine when the Docker daemon is restarted.

Braintelligence:
@mkuron Do you have an idea why I might have got a mail from letsencrypt that my certificates are deprecating today but the online certificate on the mailcow webGUI shows a certificate that was issued on 13th september and is valid through december?
Was it some update of the ACME stuff that made this weird? Could this repeat itself?

chriscroome:
@Braintelligence are they are for different sets of domains?

Braintelligence:
@chriscroome They're for autoconfig.foo.bar, autodiscover.foo.bar and mail.foo.bar.

chriscroome:
@Braintelligence if both certs are for the same set of domains then I guess it might because you got two separate certs in the past, perhaps the old one was from before the acme container was added to the project? In any case, simply ignore the email :-)

andryyy:
Oh, good catch. Can you merge it? Will be back from vacation on Sunday.

Issue #619, opened 2017-09-19
Higher Docker-Compose version issue


jpbaril:
Hi,

I just ran the update script. At the end of the update process container were not able to start and I received errors because the docker-compose file format version changed to 2.3 and my system's Docker engine version (1.12) does not support it. Indeed, I run Mailcow-dockerized on CoreOS and it only supports Docker compose version 2.1. Fortunately, I did not delete the old container images when asked and I had a backup of my modified docker-compose file.

Would that be technically possible to modify the update script to first show a message and ask for approbation before updating if the to-be-updated docker-compose file will now requires a higher Docker engine version ?

Thanks

Braintelligence:
There are two possible ways to "fix" this:

Either there are OS requirements that do not include CoreOS and there's nothing to do here.

Or we decide to include CoreOS in the supported OSes and have to work on this.

Are there others who need this?

andryyy:
We can consider it. 👍 Is there a roadmap etc. when they will support version xy?

jpbaril:
I don't remember seing that Mailcow had explicit OS or Docker version requirements.

My point is that requirements indeed changed without me knowing and that broke my setup when I ran the Update script. So I was just asking if it was feasable in any technical way in the script  to dynamicaly check if requirements are met before updating.

As for CoreOs, they plan to update to Docker engine 1.13 in October and then follow quickly with Docker newest version (ref: https://github.com/coreos/bugs/issues/1930#issuecomment-328696579). But even if they do this time there is nothing garantiing that in the future you will not decide to update to a higher version before CoreOs do. That's why I was asking if there is a way to check for requirements in the script, or maybe just inform of the new requirements in the docs and I will have a look there before updating.

mkuron:
Perhaps the update script could parse https://raw.githubusercontent.com/docker/docker.github.io/master/_includes/content/compose-matrix.md and check whether the Docker version installed supports the version currently set in docker-compose.yml.

jodumont:
I think it in the same kind of idea/error

this morning I try to install mailcow

mailcow-dockerized $ docker-compose pull
Pulling redis-mailcow (redis:alpine)...
ERROR: The Docker Engine version is less than the minimum required by Compose. Your current project requires a Docker Engine of version 17.06.0 or greater.

I look and I'm sorry I have the last release of Docker-compose which is docker-compose version 1.16.1, build 6d1ac21  

https://docs.docker.com/compose/install/#prerequisites

[...]

PR #664, opened 2017-10-14
Compatible CoreOS Stable

JOduMonT:
related to : https://github.com/mailcow/mailcow-dockerized/issues/619
If MailCow follow Docker EE release instead of Docker CE it will keep this compatibility


luxflux:
Does this work? It's quite different to the changes in `dev` branch?

PR #686, opened 2017-10-26
Fixing error with somaxconn

### LINKED ISSUE #685: redis-mailcow complaing about somaxconn
When starting mailcow, redis keeps telling 

># WARNING: The TCP backlog setting of 511 cannot be enforced because /proc/sys/net/core/somaxconn >is set to the lower value of 128.

A Solution for this is to add 

>       sysctls:
>       - net.core.somaxconn=511

in the docker-compose.yml in the redis-mailcow section
My Current branch is: dev
I am using the latest commit  b807712

Oratorian:
Fixing error -> The TCP backlog setting of 511 cannot be enforced because /proc/sys/net/core/somaxconn is set to the lower value of 128

Fixes #685

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -43,6 +43,8 @@ services:
             - mysql
 
     redis-mailcow:
+      sysctls:                                                                     
+        - net.core.somaxconn=511
       image: redis:alpine
       volumes:
         - redis-vol-1:/data/


mkuron:
I only see this message once after starting — I wouldn’t consider that spamming.

Actually I don’t think we even need such a large TCP backlog as our Redis will never have more than a few dozen simultaneous connections. Maybe 20 from PHP (which has one connection per request and php-fpm only allows a certain number of processes to run at the same time) if the admin UI gets lots of traffic, 10 from the different syslog instances and a few dozen from rspamd (which has a connection pool). So we could either ignore the message if it only comes up once during startup or configure redis to use a smaller TCP backlog.

Oratorian:
my redis server keeps me telling that after each exchangesync i do with my mailclients

mkuron:
Ok, I don‘t use Exchange, so I don’t see that.

I don’t think SOGo (which is responsible for EAS) even uses Redis, so I wonder where that comes from. Can you provide logs for all services around the time of the somaxconn message?

andryyy:
Hm. Any updates?

Oratorian:
Ok, seems fixed. Was a major System issue which kept my mailcow restarting after each connection attempt from an activesync exchange connection.

Your were right the message just pops up once after restart.
Pull Request can be closed.

Issue #692, opened 2017-10-27
rspamd sometimes does not display history


DasWiesel2305:
Since the last update via the "./update.sh" script I have some weird issue with rspamd.
Sometimes rspamd can not display the history in the UI ($DOMAIN/rspamd).
It is prompting instantly with "timeout while connecting to server" or "gateway timeout", but everything is working as it should only the rspamd history is not working.

I've gathered the following suspicious entries from the docker logs of the rspamd container.

2017-10-27 22:03:21 #18(controller) <61c729>; lua; history_redis.lua:194: got error timeout while connecting the server when getting history: no value

The strange thing is that it is working sometimes after showing this message, if I spam "Update". 
Once it has loaded and I like to update once again it is quitting with the above mentioned errors.

If you need any further information, I will gladly provide to you.

DasWiesel2305:
I've also found that rspamd is causing one CPU core to lock at 100%, if the above mentioned phenomenom is present.

https://www2.pic-upload.de/img/34193560/1_core_100_cpu.png

I can solve it by restarting the rspamd docker container.

markirb:
I think this is a duplicate of
https://github.com/mailcow/mailcow-dockerized/issues/575

After playing around a bit, it seems that sometimes the JSON API request does not return anything. I suspect some kind of runtime or memory issue of PHP with the default of 2000 last log entries returned. The returned JSON is around 7MB for me, so this thing is quite big...
The underlying code needs to decode the JSON from rspamd and reencode only the rows as JSON.
So this is huge overhead... maybe this could be avoided.

Another way which worked for me was to reduce the number of rows:
When creating a 
data/conf/rspamd/local.d/history_redis.conf
with a reduced size, e.g. 
```
servers = redis:6379;
nrows = 400;
key_prefix = "rs_history";
compress = true;
```
(restart rspamd-mailcow afterwards)
it worked reliably again for me.

axd:
@markusg Your solution doesn't work for me.

I've researched and I'm receiving this error:
`SyntaxError: JSON.parse: unterminated string at line 1 column 1060611 of the JSON data`

The JSON it's truncated and it's incomplete on the response.

andryyy:
Do you have that exact line?

markirb:
@axd 
Is this the output of the Firefox JSON parser? I've noticed the same thing sometimes.
But i believe this is a problem of Firefox, as when you look at the RAW data, it is not truncated....
[...]

PR #701, opened 2017-10-30
Rspamd History fix small error

markirb:
I found this while playing around with
https://github.com/mailcow/mailcow-dockerized/issues/692
might be related, but i think this still does not fix it


andryyy:
Oh, thanks!

PR #703, opened 2017-10-30
Adding a primary key in domain_admins table

### LINKED ISSUE #702: Missing primary key in domain_admins table
After following the [migration guide](https://mailcow.github.io/mailcow-dockerized-docs/migrate-mc014/) I saw that there's no primary key in the ```domain_admins``` table.
Therefore you can add 2 domain admins while one is set active and the other one is inactive and the database won't know what to do.

MAGICCC:
Fixes #702

### DIFF
--- a/data/web/inc/init_db.inc.php
+++ b/data/web/inc/init_db.inc.php
@@ -248,8 +248,8 @@ function init_db_schema() {
           "active" => "TINYINT(1) NOT NULL DEFAULT '1'"
         ),
         "keys" => array(
-          "key" => array(
-            "username" => array("username")
+          "primary" => array(
+            "" => array("username", "domain")
           )
         ),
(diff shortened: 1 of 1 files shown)


andryyy:
Thanks! I will also add more foreign keys later. :-)

andryyy:
And I reverted it... Didn't check the file before, my fault. But why push it untested?

MAGICCC:
Tests were made and the script worked well with our mailcow. 
These changes where made, because it occured that the table ```domain_admins``` had non-unique combinations of the columns username-domain-active. Therefore a primary key on the columns (username, domain) is recommended as a minimum. The existing index on username was removed, because the new primary key does the same thing (via partial index usage).
Nevertheless additional foreign keys that ensure the existence of a username and a domain would of course increase the consistency of the mailcow database.
By the way: What was wrong with the PR? Maybe you can provide a better solution.

andryyy:
I fixed it.

Your changes were not even applied without changing the version in the head. Change it and you will see the error. :-)

PR #707, opened 2017-11-01
Disable Buttons on Mailbox / Domain when no permission to interact with them

develth:
Hi!

It was a bit confusing for Domainadmins when they had a Button "Add Domain", but can´t add a Domain and get a error after trying to save.

So this PR disables the Actions on the Domain table for Users that don´t have a permission to use this action.


andryyy:
Ah, damn, actually forgot to merge it... 

Thanks!

Issue #625, opened 2017-09-20
SOGo / Roundcube with subdomain


ghost:
Hi,

as in the title posted, how can I configure a subdomain for SOGo (instead of mydomain/SOGo) or for Roundcube (instead of mydomain/rc).

I haven't found any config, with which I can configure a subdomain. Do I need to install a nginx on the host (besides docker) and use this for a subdomain? Can someone give me a short howto, because I'm new to this stuff?!

Thanks in advance :-)

PhoenixPeca:
Hello @mwrnck!

> Do I need to install a nginx on the host (besides docker) and use this for a subdomain?

Yes. You will need to set the host machine's nginx as a reverse proxy for the dockerized nginx-mailcow. To get yourself started, just follow the steps as stated in the [documentations](https://mailcow.github.io/mailcow-dockerized-docs/firststeps-rp/)

ghost:
Thanks @PhoenixPeca ,
can you give an example-config for the nginx with a ssl-config? I don't get it to work:
First I change the bindings in the mailcow.conf and after this I create a new nginx file on the host, where I insert the lines of the docu.
If I open mail.example.de I get a blank page. Can you or anyone else help me with this?

Braintelligence:
Would it be acceptible to just create a subdomain that points per HTTP301 to mydomain.com/SOGo ?
That's what I do in such cases, if it's only about having a more beautiful link.

axd:
@mwrnck There's a thread open to achieve something like you want but it's in standby: https://github.com/mailcow/mailcow-dockerized/issues/393.

But as @Braintelligence says, that's the only way (for now) to make it work. (https://domain.com/SOGo).

Issue #528, opened 2017-08-23
Whitelabeling for mail.domain


jaketame:
Is it possible to configure additional domain names automatically into the SAN such as

mail.domain1.org
mail.domain2.org?

Cheers

chriscroome:
Yes, add them using the `ADDITIONAL_SAN` environmental variable in the `mailcow.conf` configuration file, see [this line](https://github.com/mailcow/mailcow-dockerized/blob/master/generate_config.sh#L80).

jaketame:
Thanks for reply, was hoping for a more automated way / when add new domain - tickbox to add as SAN?

chriscroome:
Why do you want additional `mail.example.org` and `mail.example.org.uk` sub-domains, why don't you just use the main domain for the server -- you can add additional email domains via the web interface and these are added to the SAN list for `autodiscover.` and `autoconfig.` automatically, or perhaps I'm misunderstanding what you are asking?

jaketame:
So I can provide some sort of white labelling to sites I host and provide them with there own mail.domain to connect to with a validate cert.

chriscroome:
Fair enough, I'm not the one to answer this, perhaps it would be worth making the issue title clearer regarding the aim?

mkuron:
Just adding extra domain names is not sufficent to get white-labeling. MAILCOW_HOSTNAME is used in a number of places throughout the configuration files and in the web UI code. Mailcow in its current state is not made for white-label setups.

kunago:
I see the usage in easy migration process. Once you set your emails in clients and use MAILCOW_HOSTNAME, it's then difficult to simply switch servers.

What about having some 2 extra fields in the UI for each domain where you could set the incoming and outgoing domain names? Or maybe more then 2 depending on the amount of services.

jaketame:
I give a client mail.there-domain.com, which just points to my mailserver, its a bit neater and nice for them. They don't need a full whitelabel service. The certificate is to ensure they don't have errors client side by using a different domain connecting to the server.

kunago:
@jaketame I do the same thing, however, autoconfig points to MAILCOW_HOSTNAME which may be confusing for some.

mkuron:
[...]

PR #709, opened 2017-11-01
Verfify additional subdomains

markirb:
Me and obviously some other users (https://github.com/mailcow/mailcow-dockerized/issues/528) don't use only one domain, but want to provide the service under additional subdomains for user convenience. (mail.* but also roundcube.*...  whatever)

I want to give a starting porint, with the way i do it at the moment:

This introduces an ADDITIONAL_SUBDOMAINS env variable, where you can specify additional subdomains, which are then verfied with letsencrypt for all SQL_DOMAINS.
This saved me A LOT of typing for the ADDITIONAL_SAN 😄 

However there are two limitations:
1. You cannot go beyound 100 names per certificate, which makes it applicable to only a limited range of  usecases.
2.  The autoconfig.* content is not yet remapped - i don't think this is necessary: I use  the feature to not have to explain to the user that they have to use the main mailcowdomain.

What also could be done is the convenience redirect (https://github.com/mailcow/mailcow-dockerized/issues/625).

Any thoughts on this?

### DIFF
--- a/data/Dockerfiles/acme/docker-entrypoint.sh
+++ b/data/Dockerfiles/acme/docker-entrypoint.sh
@@ -163,8 +163,20 @@ while true; do
   declare -a SQL_DOMAIN_ARR
   declare -a VALIDATED_CONFIG_DOMAINS
   declare -a ADDITIONAL_VALIDATED_SAN
-  IFS=',' read -r -a ADDITIONAL_SAN_ARR <<< "${ADDITIONAL_SAN}"
+  IFS=',' read -r -a TMP_ARR <<< "${ADDITIONAL_SAN}"
   log_f "Detecting IP addresses... " no_nl
+
+  unset ADDITIONAL_WC_ARR
+  unset ADDITIONAL_SAN_ARR
+  for i in "${TMP_ARR[@]}" ; do
+    if [[ "$i" =~ \.\*$ ]]; then
+      ADDITIONAL_WC_ARR+=(${i::-2})
+    else
+      ADDITIONAL_SAN_ARR+=($i)
+    fi
+  done
+  ADDITIONAL_WC_ARR+=('autodiscover')
+
   IPV4=$(get_ipv4)
   IPV6=$(get_ipv6)
   log_f "OK" no_date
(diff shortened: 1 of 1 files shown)


data/Dockerfiles/acme/docker-entrypoint.sh
```diff
@@ -192,7 +185,7 @@ while true; do
   done
 
   # Unique elements
-  ALL_VALIDATED=(${VALIDATED_MAILCOW_HOSTNAME} $(echo ${VALIDATED_CONFIG_DOMAINS[*]} ${ADDITIONAL_VALIDATED_SAN[*]} | xargs -n1 | sort -u | xargs))
+  ALL_VALIDATED=($(echo ${VALIDATED_MAILCOW_HOSTNAME} ${VALIDATED_SUBDOMAINS[*]} ${ADDITIONAL_VALIDATED_SAN[*]} | xargs -n1 | sort -u | xargs))
```
andryyy: The hostname should be first here, any reason you changed it?

docker-compose.yml
```diff
@@ -219,6 +219,7 @@ services:
       image: nginx:mainline-alpine
       command: /bin/sh -c "envsubst < /etc/nginx/conf.d/templates/listen_plain.template > /etc/nginx/conf.d/listen_plain.active &&
         envsubst < /etc/nginx/conf.d/templates/listen_ssl.template > /etc/nginx/conf.d/listen_ssl.active &&
+        export ADD_WILDCARDS=$$(IFS=','; for S in $$ADDITIONAL_SUBDOMAINS; do A=$$A' '$$S'.*'; done; echo $$A) &&
```
andryyy: This works in all test cases?

data/Dockerfiles/acme/docker-entrypoint.sh
markirb: Yep, the reason is, your hostname might be mail.example.com and if you add mail.* you have to remove the duplicate. Otherwise ACME-verfication will not work.

docker-compose.yml
markirb: I find this ugly too, but did not find any way around it, to keep it as simple as it is right now.

The double dollars are for escaping docker-compose.
/bin/sh is from alpine, so should always give same behaviour.

If ADDITIONAL_SUBDOMAINS contains semicolons or is otherwise malformed, then it breaks the nginx config.
If its empty or contains a comma separated list, it works as expected.

markirb (commented):
This was deliberate, to have no space at the end in case the variable is empty. However nginx does not care about that extra whitespace

data/Dockerfiles/acme/docker-entrypoint.sh
mkuron: MAILCOW_HOSTNAME has to be the first one. That‘s what LE will use for the common name field, and we don‘t want one of the autoconfig domains to be used there. Please change this line back to what it used to be and instead delete MAILCOW_HOSTNAME from VAPIDATED_DOMAINS if it‘s in there.

mkuron (changes requested):

data/Dockerfiles/acme/docker-entrypoint.sh
markirb: Yeah, that is a valid point, I haven't considered that

markirb:
This was changed to fit the review a long time ago 🤔
I'm still using this code without any problems on my mailcow setup ever since.

I updated it today to fix a merge conflic.
Do you want  anything else changed or if you do not want to merge should it be closed or done another way?

mkuron:
It still seems kind of redundant to me. We have `ADDITIONAL_SAN` already and it‘s really not that difficult to add extra domains to it... either way you‘re likely not adding and removing domains every day. So it would mainly serve to confuse people who don‘t find the difference between `ADDITIONAL_SAN` and `ADDITIONAL_SUBDOMAINS` self-explanatory.

markirb:
Jeah, sure that's a valid point. 
But also if you do not have ADDITIONAL_SUBDOMAINS in your config it does not do anything.
And it saves you from having to keep your domains in sync in two places, which is especially useful if you have a handful of domains, as in my case the ADDITIONAL_SAN spans over several lines without any possibility to commment or format the stuff. 

So another way would be to have a nice way of a format for those lines, what would help clarity.
I think you really do not want more than a handful of domains in the SAN right now.

markirb:
Hey,

as the functionality is indeed a duplicate of ADDITIONAL_SAN, I now changed the whole thing to be within ADDITIONAL_SAN.
So now you can do
ADDITIONAL_SAN=mail.*,subdomain.whatever.com
and the wildcard automatically expands to include all of your SQL_DOMAINS.

I left out everything unnecessary else I did above.
So it boils down to this simple functionality, which is everything I ever used...
I hope you find this more useful

markirb:
Before this goes stale and to avoid confusion: forget all the upper posts.
This PR does two things:
1. cleanup the entrypoint.sh of ACME to not have the duplicate code for autoconfig.* and autodiscover.* 
2. recognizes subdomain.* from ADDITIONAL_SAN to allow the same mechanism for custom subdomains.

andryyy:
Can you fix the script to match the current version?

markirb:
Done.
Also removed the autoconfig.* as you did in the current version.

andryyy:
Oh, nice, thanks!

Let me run a few tests, then we will merge it. 👍

andryyy:
Thanks, I will add it to the docs when we push the new image.

PR #715, opened 2017-11-04
Update rspamd

mkuron:
rspamd updated its APT packages last night, so we can now fis https://github.com/vstakhov/rspamd/pull/1898 in Mailcow.


andryyy:
Nice, that's a very cool fix! Thanks!

PR #589, opened 2017-09-12
Rspamd: better scores for mx_check symbols

marrco:
discussion on https://github.com/mailcow/mailcow-dockerized/issues/576

PR #730, opened 2017-11-08
Rspamd: increase task timeout

mkuron:
Since #589, we wait up to 8 seconds for mx_check. Incidentally, the default task timeout is only 8 seconds, so we sometimes run into the latter timeout. This PR increases it. The 4 extra seconds should be plenty as all other modules are much faster than the mx_check.


andryyy:
Thanks 👍

PR #732, opened 2017-11-09
Better Dovecot auth failure logging

extremeshok:
Log all failed authentication attempts, with the "service auth"
ie allows one to view failed login usernames when postfix authorizes against dovecot.

sample output in the log will be like this
````
auth: Info:(test@domain.com,199.99.99.1,<5Ff/7Htd6QCpAM+Y>): invalid credentials
````

https://wiki2.dovecot.org/Logging
````
auth_verbose=yes enables logging all failed authentication attempts.
`````

### DIFF
--- a/data/conf/dovecot/dovecot.conf
+++ b/data/conf/dovecot/dovecot.conf
@@ -3,6 +3,7 @@
 # --------------------------------------------------------------------------
 auth_mechanisms = plain login
 #mail_debug = yes
+auth_verbose=yes
 log_path = syslog
 disable_plaintext_auth = yes
 # Uncomment on NFS share

### CODE REVIEW

**Comment by mkuron** [comment]
Is the old error message that only contains the IP address still logged? If not, then you need to change the fail2ban container to parse the new message.

Also, the log messages that fail2ban currently parses (see below) already contain both user name and remote IP address, so your change might not be necessary.
```
RULES[2] = '-login: Disconnected \(auth failed, .+\): user=.*, method=.+, rip=([0-9a-f\.:]+),'
RULES[3] = '-login: Aborted login \(no auth .+\): user=.+, rip=([0-9a-f\.:]+), lip.+'
RULES[4] = '-login: Aborted login \(tried to use disallowed .+\): user=.+, rip=([0-9a-f\.:]+), lip.+'
```

mkuron:
Closing this as requested details were never provided.

PR #743, opened 2017-11-13
Add support for LDAP auth.

danielmorlock:
This will add a minimal LDAP auth support for admin users and mailboxes:

1. Create a LDAP user (inetOrgPerson) having `uid` and to also update mailbox password having one or more `mail` attributes.
1. Create a mailcow (domain) admin (matching `uid`) and/or mailboxes (matching one of `mail`)
2. Configure LDAP in `inc/vars.local.inc.php` as follows:  
```php
<?php
$ldap_config = array(
	'uri' => 'ldap://your-ldap-host',
	'base_dn' => 'dc=mx,dc=example,dc=org',

	// Optional params (see defaults below):
	//'bind_admin_dn' => null,
	//'bind_admin_pass' => 'null,
	//'user_filter' => '(|(uid=%*)(mail=%*))',
);
```
3. Login using mailcow admin username and/or mailbox
4. User is search in LDAP using the specified filters (by default using `uid` and `mail` attributes)
5. LDAP user login is tried using the given password
6. On success, the corresponding passwords are updated in the mailcow db
7. The native login is repeated (now using the updated passwords).

### DIFF
--- a/data/Dockerfiles/phpfpm/Dockerfile
+++ b/data/Dockerfiles/phpfpm/Dockerfile
@@ -33,6 +33,7 @@ RUN apk add -U --no-cache libxml2-dev \
   imagemagick \
   libtool \
   librsvg \
+  openldap-dev \
   && pear install channel://pear.php.net/Net_IDNA2-0.2.0 \
     channel://pear.php.net/Auth_SASL-1.1.0 \
     Net_IMAP \
@@ -42,7 +43,7 @@ RUN apk add -U --no-cache libxml2-dev \
 	&& docker-php-ext-enable redis apcu memcached imagick \
 	&& pecl clear-cache \
 	&& docker-php-ext-configure intl \
-  && docker-php-ext-install -j 4 intl pdo pdo_mysql xmlrpc gd zip pcntl opcache \
+  && docker-php-ext-install -j 4 intl pdo pdo_mysql xmlrpc gd zip pcntl opcache ldap \
   && docker-php-ext-configure imap --with-imap --with-imap-ssl \
 	&& docker-php-ext-install -j 4 imap \
 	&& apk del --purge autoconf g++ make libxml2-dev icu-dev imap-dev openssl-dev cyrus-sasl-dev pcre-dev libpng-dev libpng-dev libjpeg-turbo-dev libwebp-dev zlib-dev imagemagick-dev \
@@ -55,7 +56,6 @@ RUN apk add -U --no-cache libxml2-dev \
   echo 'opcache.save_comments=1'; \
   echo 'opcache.revalidate_freq=1'; \
 } > /usr/local/etc/php/conf.d/opcache-recommended.ini
-
 
 COPY ./docker-entrypoint.sh /
 
--- a/data/web/inc/functions.inc.php
+++ b/data/web/inc/functions.inc.php
@@ -144,9 +144,17 @@ function verify_ssha256($hash, $password) {
 		return false;
 	}
 }
-function check_login($user, $pass) {
+function check_login($user, $pass, $skip_ldap = false) {
 	global $pdo;
 	global $redis;
+	global $ldap_config;
+
+	if(!$skip_ldap && $ldap_config) {
+		$retval = ldap_check_login($user, $pass);
+		if($retval !== false)
+			return $retval;
+	}
+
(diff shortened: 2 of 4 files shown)


mkuron:
Interesting, thanks for submitting that. Proper LDAP support requires a lot more though. My main concern with your suggested solution is that you mirror the data instead of using it directly. This requires people to log into the web UI before they can connect to any service or receive mail and they need to log in whenever they have changed their LDAP password.

Postfix, Dovecot and SOGo can directly access an LDAP server, just like they currently use the MySQL database. Of course, that requires changes in many more places and is more work to test.

danielmorlock:
This is indeed not a full-LDAP-support as discussed here: https://github.com/mailcow/mailcow/issues/284
So this is thought to be a workaround until full-LDAP-support is on your roadmap.

mkuron:
Closing this because it will be too confusing for end users if their mail and LDAP passwords get out of sync. A better solution until we get proper LDAP support (if we ever do) would be to extract the password hashes from LDAP and import them into the Mailcow database at 5-minute intervals so the users don't notice that the data is mirrored.

etique57:
Sorry to reopen the subject, but would this pull request be a better fit?

https://github.com/mailcow/mailcow-dockerized/compare/master...str8labs:feature-ldap

mkuron:
@etique57, that looks like a proper solution. It is unclear to me though how LDAP and SQL backends interact with each other and whether one can store user settings in SQL while the user account itself is in LDAP.

andryyy:
This is something we can start to work with after all. LDAP should be configurable via UI, some fields should be greyed out as not-editable, a table should list inactive LDAP users etc. - it still needs a lot of work.

That's a very good start, yes. Can you check the SOGo source mixing?

mkuron:
Please post a pull request so we can discuss improvements there.

One thing that just came to my mind: `LDAP_MAIL_ATTR` needs to accept multiple values. `mail` is only allowed once by Microsoft's schema, so to work with Active Directory we would need to read `proxyAddresses` as well.

andryyy:
I think alias addresses should still be handled in mailcow via SQL. That's what his code does right now, I think. It is easier to not split that logic too much.

etique57:
Hello gents,

Glad this seems to raise interest!

I'm in no way related with this gentleman's work, @bkleiner. Is there a way to compare/input his LDAP feature into the latest master of Mailcow, so I can easily reconciliate?

I'm a layman in Github, I think I would need to do the following:
- create a pull request from the latest master of Mailcow
- create a fork (?) of the LDAP-feature

But how can I somehow merge them?

Cheers!

p-artur:
Hi, 
Thx for your hard work!
Any updates in this thread?

andryyy:
@Programmierus Created a cool tool :)

p-artur:
> @Programmierus Created a cool tool :)
Oh! Thanks for that! Any plan to integrate this in your project?

andryyy:
No. It is an external tool to create mailboxes from external sources.

Programmierus:
@andryyy thanks for mentioning...

@p-artur it's entirely an external solution that is designed as a kind of extension...

p-artur:
@Programmierus thanks for this job!
@andryyy what if try to do this sync job for current user automitically after success login via ldap userdb/passdb in dovecot's postloginscript mechanism? postfix and sogo logging in via dovecot?

danielmorlock:
for the sake of completeness, the tool from @Programmierus is here: https://github.com/Programmierus/ldap-mailcow

p-artur:
@danielmorlock, @andryyy, @Programmierus thanks!

maverick85:
Hi all,

As I see it the programmierus solution is an external and unsupported solution (it says so in his page).
Any decent company that wants to keep the systems in sync and provide a centralised user management solution will resort to LDAP/Active Directory.
I have been following and using mallow for a couple of years, and we see proper/built-in LDAP support as a must-have feature.

I saw some comments about email and ldap passwords out of sync? well and now, being mailcow a totally independent system, aren't the passwords always different and never in sync anyway?

I think programmierus solution is quite interesting, but specially if from there stems a solution to make mailcow fully LDAP compliant that reads ldap users and manages mailboxes accordingly.
It would also be interesting to have an option to preserve emails/mailboxes even after removing an user.

Cheers

PR #757, opened 2017-11-18
DNS diagnostics: ignore spaces in DKIM record

mkuron:
OVH's DNS admin panel automatically inserts spaces after the semicolons in DKIM records. That's fine, but Mailcow's DNS diagnostics needs to remove them to perform a string comparison.

### DIFF
--- a/data/web/diagnostics.php
+++ b/data/web/diagnostics.php
@@ -194,18 +194,23 @@ foreach ($records as $record)
     }
   }
   
-  foreach ($currents as $current) {
-    $current['type'] == strtoupper($current['type']);
+  foreach ($currents as &$current) {
     if ($current['type'] != $record[1])
     {
       continue;
     }
     
-    elseif ($current['type'] == 'TXT' && strpos($record[0], '_dmarc.') === 0) {
+    elseif ($current['type'] == 'TXT' && strpos($current['txt'], 'v=DMARC1') === 0) {
+      $current['txt'] = str_replace(' ', '', $current['txt']);
       $state = state_optional . '<br />' . $current[$data_field[$current['type']]];
     }
     else if ($current['type'] == 'TXT' && strpos($current['txt'], 'v=spf1') === 0) {
       $state = state_optional . '<br />' . $current[$data_field[$current['type']]];
+    }
+    else if ($current['type'] == 'TXT' && strpos($current['txt'], 'v=DKIM1') === 0) {
+      $current['txt'] = str_replace(' ', '', $current['txt']);
+      if ($current[$data_field[$current['type']]] == $record[2])
+        $state = state_good;
     }
     else if ($current['type'] != 'TXT' && isset($data_field[$current['type']]) && $state != state_good) {
       $state = state_nomatch;
@@ -213,6 +218,7 @@ foreach ($records as $record)
         $state = state_good;
     }
   }
+  unset($current);
   
   if (isset($record[3]) && $record[3] == state_optional && ($state == state_missing || $state == state_nomatch)) {
     $state = state_optional;


andryyy:
The DNS diagnostic doesn't work correctly... First of all tls:// will not work for port 443 and https. You would need ssl://. When you don't have a dmarc record, it just prints the SPF for record for DMARC and leaves SPF empty. That's what I expected, the code is veeery hacky.

mkuron:
Regarding 443: are you sure? I believe it worked the last time I tried. We don‘t have a starttls for HTTPS, which is correct. ssl:// should auto-negotiate the encryption and tls:// should require a minimum of TLS 1.0.

Regarding SPF/DKiM/DMARC: thanks for finding this. The TXT record handling is indeed hacky. Too bad that a free-format text field like TXT has been abused for structured data like this... I‘m happy to provide a fix tomorrow.

andryyy:
Oh, wasn't supposed to be closed. O_O

andryyy:
> Regarding 443: are you sure? I believe it worked the last time I tried.

I also thought it was working last time I checked. Strange.

mkuron:
Ok, I‘ll investigate and fix it.

data/web/diagnostics.php
```diff
@@ -194,25 +194,31 @@
...
-    $current['type'] == strtoupper($current['type']);
+  foreach ($currents as &$current) {
     if ($current['type'] != $record[1])
     {
       continue;
     }
     
-    elseif ($current['type'] == 'TXT' && strpos($record[0], '_dmarc.') === 0) {
+    elseif ($current['type'] == 'TXT' && strpos($current['txt'], 'v=DMARC1') === 0) {
+      $current['txt'] = str_replace(' ', '', $current['txt']);
```
jhesketh, line 204: I realise this has already been merged, but wouldn't trim() be more appropriate?

mkuron: No, because the spaces can be anywhere in the string. `trim` would only remove them at the beginning and end.

andryyy: I think he was removing the spaces in between the records value.

andryyy:
Hm, I used a wildcard CNAME which the resolver followed and run the query on the linked domain, that's why I received a wrong value for DMARC.
"Wildcard" TXT tricked the DMARC check, too.
Meh, many "ifs" in this check. :-)

Moved it to a modal, will create a PR in a bit and link it.

mkuron:
Didn't think about wildcards... I have a fix though for the TXT issue, will post the PR in a few minutes. I can't reproduce the tls:// vs. ssl:// issue though.

data/web/diagnostics.php
jhesketh: Right, I would have thought spaces in the middle of the record would want to be preserved/checked against?

Additionally line 203 (just above) checks that the record starts with 'v=DMARC1' at pos 0, meaning if there are any leading spaces the record won't be recognised. Perhaps the trimming should happen earlier?

mkuron: Leading spaces are not valid in DMARC, so we don't care about them. The only place where they are optionally possible is after the semicolons. They carry no meaning though, so we can just remove all spaces.

jhesketh: Okay, sounds good.

andryyy:
@mkuron, can you check the new code in dev?

mkuron:
After the `foreach` ending on line 176, you need to `unset($current)`, otherwise the last entry of that array will get overwritten by the next `foreach`.

Also, you still need the changes from #853 to get correct TXT handling. You can then remove lines 212 and 206 from your code as my PR should handle those cases (and some others too).

Line numbers from d76d103c76f5c4f97e004b740ec8c6c99d44638a.

andryyy:
The missing unset is handled from line 200.
I did it due to a dmarc check or something. I actually cannot remember why I had to set it to false and move it there, but there was a reason. 😛 I removed it now and will test it later.

Why is POP3 SRV handled different from the other SRV records? Like this:
```
$current_records = dns_get_record('_pop3s._tcp.' . $domain, DNS_SRV);
if (count($current_records) == 0 || $current_records[0]['target'] != '') {
  if ($autodiscover_config['pop3']['port'] != '995') {
    $records[] = array(
      '_pop3s._tcp.' . $domain,
      'SRV',
      $autodiscover_config['pop3']['server'] . ' ' . $autodiscover_config['pop3']['port']
    );
  }
}
else {
  $records[] = array(
    '_pop3s._tcp.' . $domain,
    'SRV',
    '. 0'
  );
}
```

And why is $current unset after modifying it in this loop?
```
    if ($record[1] == 'SRV') {
      foreach ($currents as &$current) {
        if ($current['target'] == '') {
          $current['target'] = '.';
          $current['port'] = '0';
        }
        $current['data'] = $current['target'] . ' ' . $current['port'];
      }
      unset($current);
    }
```

After thinking about the TXT stuff... I think we should just display what is returned. So when it returns a SPF record because of a wildcard TXT, it is fine. You just see the real result. I removed the checks now.

mkuron:
> Why is POP3 SRV handled different from the other SRV records?

This is to support advertising POP3 as "not provided" by setting its host to `.`. The [autoconfig.php](https://github.com/mailcow/mailcow-dockerized/blob/26e2baa1311c3a586251e843284e1d0363c060eb/data/web/autoconfig.php#L51) also checks for that. We could also do that check for IMAP, but who wouldn't want to use IMAP?

> And why is $current unset after modifying it in this loop?

After a foreach-by-reference, you should always unset the loop variable. This only deletes the reference, not the array element. If you don't unset, you will overwrite the last array element if you ever set that variable again, like in the example below.
```
$arr = array(0,0,0);
foreach ($arr as &$val)
{
  $val += 1;
}
$val = "blah";
print_r($arr); // produces array(1,1,"blah")
```

andryyy:
Oh, I missed an s. :-) 

Thanks for clarifying!

PR #818, opened 2017-12-13
Update imapsync

jhesketh:
This PR contains two commits:

 - A fix for usage leaks via CVE-2013-4279
 - An update of imapsync

I'm not sure if the current imapsync in mailcow has local changes or not. If so, they will need to be reapplied. I have also not tested if this works and I am assuming imapsync hasn't broken backwards compatibility.


andryyy:
Hi, Thanks, I just updated it to the new version which disables that check by default. 👍 Will be pushed as new image with the next dev -> master merge.

jhesketh:
Ah cool. The diff between the versions is really large though. Any reason for this?

PR #846, opened 2017-12-30
Removed duplicate checkbox from mailbox add sync job modal

feldsam:


andryyy:
Hi, Are you sure this is still an issue? We fixed a duplicate checkbox a while back. "delete2" should remain.

Am I missing something here? 🤔

feldsam:
ahh, ok, I just cherry picked to dev...now I look into file and it was already fixed.

andryyy:
I see, thanks. 👍

PR #849, opened 2017-12-31
Sync job tables enhanced

feldsam:
Enhanced sync job tables - on user view there was to width overflowing table. On mailbox view I added excludes column


andryyy:
Nice, thanks!

PR #853, opened 2018-01-02
DNS diagnostics: show correct TXT records

mkuron:
Issue reported by @andryyy in #757.

I was unable to reproduce the TLSA issue for port 443 though. I may need more information about your setup to fix it.


mkuron:
Resolved the merge conflict with @andryyy's commits from yesterday. This pull request should eliminate all cases where, if one of the TXT records was missing, another one would be displayed in the "current state" column.

PR #859, opened 2018-01-03
Remove duplication of delete2 checkbox in addSyncModal

topelrapha:
I think i could be removed!


andryyy:
Hi, Thank you for this PR, but we already fixed it in dev. :-) Will merge very soon.

PR #870, opened 2018-01-06
New interface for manage filters more userfriendly way

feldsam:
Hello, I programed new "user filters" for easier filters managing, without writing own scripts. User/Admin can define rules and sieve scripts are automatically generating as active sieve postfilter

### DIFF
--- a/data/web/edit.php
+++ b/data/web/edit.php
@@ -768,7 +768,114 @@ if (isset($_SESSION['mailcow_cc_role'])) {
         <?php
         }
       }
-    elseif (isset($_GET['filter']) &&
+  }
+  if ($_SESSION['mailcow_cc_role'] == "admin"  || $_SESSION['mailcow_cc_role'] == "domainadmin" || $_SESSION['mailcow_cc_role'] == "user") {
+    if (isset($_GET['user_filter']) &&
+      is_numeric($_GET['user_filter'])) {
+        $id = $_GET["user_filter"];
+        $result = getMailboxUserFilter($id);
+        if (!empty($result)) {
+        ?>
+          <h4><?=$lang['edit']['user_filter'];?></h4>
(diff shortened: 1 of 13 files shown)


andryyy:
Hi, that's very nice!

The reason I have not implemented it is the existence of SOGos own filter managment. It writes JSON to a user pref table and creates these filters when you save them. It does not detect changes to those filters in sieve, you'd need to change the JSON code and have SOGo itself update them. So any change to the "sogo" sieve script is destroyed as long as it is not changed within SOGo itself.

feldsam:
oh, I did not observed it in SOGo, so probably waste of work. On the other hand, it is not bad have it in mailcow UI, just not connected with SOGo filters...

In future I have to ask first :) have mailcow some irc or slack?

hachre:
Yes! `irc.freenode.net`, channel `#mailcow`

andryyy:
We thought about making SOGo optional, so it would be a good feature then!

Arkaniod:
Making SOGo optional would be a great idea, explicit for an Backup-MX running with Mailcow, where SOGo isn't needed

feldsam:
What about this? Should I solve conflicts? or it is not planned to merge this?

discofris:
Having SOGo as an optional module, would be a fantastic feature!
After each restart I manually stop the SOGo docker container.

lavdnone:
Very useful. Maybe put it in to sieve_before ...
https://wiki1.dovecot.org/LDA/Sieve/Dovecot "Executing Multiple Scripts Sequentially"

feldsam:
@andryyy what about this? should I solve conflicts and you merge it?

andryyy:
As long as SOGo exists and is default, no. This will be confusing and breaking filters for SOGo users.

PR #874, opened 2018-01-08
Properly wait for MySQL to come up before starting SOGo

mkuron:
If MySQL was already up (or came up within about 30 seconds) when the SOGo container was started, everything would be fine (supervisord does 10 retries in 3-second intervals). But if MySQL took too long to start up, supervisord would give up on SOGo.

This was broken for a while now. The `mysqladmin ping` check was pretty useless previously -- if MySQL was not up, the script would just exit. Even though a `while` loop was used, it would be executed no more than once. I have now replaced the check with the same one used by the ACME container, which loops until the database comes up.


andryyy:
Thanks!

Issue #878, opened 2018-01-10
mailcow.mobileconfig: encoding problem with EmailAccountName


mritzmann:
I have testet the [mailcow.mobileconfig](https://mailcow.github.io/mailcow-dockerized-docs/client-apple/) profil for macOS. The installation works well. However, names with special characters are displayed incorrectly.

![screenshot from 2018-01-10 17-27-24](https://user-images.githubusercontent.com/4533780/34783330-975ad83e-f62b-11e7-8a9b-0990b40b4052.png)

![screenshot from 2018-01-10 17-29-16](https://user-images.githubusercontent.com/4533780/34783403-ccc824e0-f62b-11e7-843c-2c3f62e864c2.png)

Is there a known problem with this?

mkuron:
I‘m not surprised by this as I haven’t tested it — none of my users have umlauts in their names. A simple `utf8_encode()` around the name in mobileconfig.php should do. I‘ll fix it soon or you could post a pull request yourself.

mritzmann:
https://github.com/mailcow/mailcow-dockerized/blob/e4b91e06a393fa796f8875822398a3b258d2fe70/data/web/mobileconfig.php#L29-L31

You have alredy `utf8_encode()` in it. Without `utf8_encode` it works

```
if (!empty($MailboxData['name'])) {
  $displayname = $MailboxData['name'];
}
```

![string](https://user-images.githubusercontent.com/4533780/34819452-10a4ba9c-f6be-11e7-9875-314a2464116f.png)

I'll make a pull request.

PR #881, opened 2018-01-11
Fix umlauts in MacOS mobileconfig generator

mritzmann:
see: https://github.com/mailcow/mailcow-dockerized/issues/878


mkuron (approved):

ericgoe:
Hello, I have just written a mail which contains umlauts in the body and the umlauts were translated to question marks.
Why did you remove `utf8_encode()` in your commit `#1aac0b2`

andryyy:
That's not related. Which client, protocol etc.?

ericgoe:
It seems that the iOS gmail app was responsible for that not mailcow. Sorry.

PR #910, opened 2018-01-15
DNS Diagnostics enhanced

feldsam:


andryyy:
Thanks, will check it. 👍

andryyy:
I removed the "allow IP" stuff as "MX" or "A" are more dynamic and should be prefered.

feldsam:
ok, no problem

Issue #748, opened 2017-11-14
Enable sub-addressing


TiiFuchs:
Hi I'd like to use sub-addressing.
So I had a look into the postfix configuration.
In main.cf there was already an entry for recipient_delimiter.
In master.cf the maildrop configuration was missing some changes.

So I changed according to the docs http://www.postfix.org/MAILDROP_README.html#direct
I changed to `    maildrop  unix  -       n       n       -       -       pipe
      flags=ODRhu user=vmail argv=/path/to/maildrop 
      -d ${user}@${domain} ${extension} ${recipient} ${user} ${nexthop}`

Then I did a docker-compose down and docker-compose up -d and checked the config with docker exec. Everything was provisioned perfectly. But sadly it does not seem to work.
Did I miss anything? Does somebody already use this?

Thanks :)

TiiFuchs:
Oh I just saw in the logs (overlooked that in the first place)

`stored mail into mailbox 'INBOX/Subaddressing'` (I sent an mail to user+subaddressing@domain.tld)

But I can't access this mail in SOGo. There is no folder visible.

Anyway. I'd like to have this mail in the inbox as every other mail too.

andryyy:
You can click the folder symbol left to "INBOX". It expands. 👍

TiiFuchs:
Yeah, but this does not work when the folder did not exist earlier. You have to relogin for the folder to be visible in SOGo.

But that's not what I want to achieve. I want those mails in the INBOX not in any subfolder.

But I skimmed a few configs and stumbled upon a config named sieve_after in the dovecot folder.
There is some code that appears to do exactly what I don't want. So I will comment it out and test if that does the job.

mkuron:
You can log into the admin interface as your user and configure whether you want the subaddresses to go into subfolders or have the tag prepended to the subject.

TiiFuchs:
So I found the option to change the behaviour from "subfolder" to "subject" in the backend. But since I need a "do nothing" option, I'm a little bit screwed.

(Background: I use a mailbox for incoming mail in GitLab, and it need's subaddressing to identify where the mail belongs to. You can give it one folder (inbox, no subfolders) and the subject will be the title of the issue. So neither subfolder nor subject is an option)

[...]

PR #915, opened 2018-01-16
Adds 'do nothing' as default for sub-addressing

TiiFuchs:
You can still change your sub-addressing configuration on the user.php in the mailcow backend.
There are three options now: subject, subfolder and none.

Subject adds the tag in [] to the subject, as before.
Subfolder moves the mail in a Subfolder with the Tag as name, as before.
None does nothing of the above and just puts that mail in your inbox. You can see the tag in the To field of the mail.

### DIFF
--- a/data/conf/rspamd/local.d/multimap.conf
+++ b/data/conf/rspamd/local.d/multimap.conf
@@ -6,8 +6,14 @@ RCPT_MAILCOW_DOMAIN {
 
 RCPT_WANTS_SUBJECT_TAG {
   type = "rcpt";
-  filter = "email:addr"
+  filter = "email:addr";
   map = "redis://RCPT_WANTS_SUBJECT_TAG";
+}
+
+RCPT_WANTS_SUBFOLDER_TAG {
+  type = "rcpt";
+  filter = "email:addr";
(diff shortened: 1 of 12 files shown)


TiiFuchs:
As discussed in #748

andryyy:
That's very cool! Have you checked all functions?

TiiFuchs:
Jap, all three functions work fine and as expected. :)

andryyy:
Upgrading mailcows will all be reset to "do nothing", that's something we should have seen coming. :-(

TiiFuchs:
Ooh >.<

TiiFuchs:
Any mechanism for an update script that checks the value for any user and updates it on update? (At least for the users that haven‘t updated until now)

I would rather not change the default back since I think „do nothing“ is a sane default here.

andryyy:
Yes, I also think "do nothing" is the best default. 👍 We could add this to mailcow.email. 

I used to add the migration stuff to the docker entrypoints. Btw. .. need to clean that up.

Edit: Don't get me wrong, I really like your implementation! I just add this as a note for us to remember what happens with default values on updates. 😄

PR #923, opened 2018-01-16
Added possibility to configure the page title

develth:
Hi there,

i recognized that the Page Title wasn´t configurable, so i added it

Thanks & Cheers, 
Thomas


AikoAppeldorn:
Nice idea, but I think it's easier to configure the title using the Mailcow admin interface.

I've added this feature in: https://github.com/mailcow/mailcow-dockerized/pull/933

andryyy:
Thank you!
I will merge @DevTek314 s approach as it is consistent to what we already have. I hope that's fine. 👍

develth:
Better Solution Wins ;)
Thanks @DevTek314

PR #925, opened 2018-01-17
Added possibility to configure autodiscover names

develth:
Added feature to configure the autodiscover DisplayName & DisplayShortName via vars.local.inc.php


develth:
@DevTek314 @andryyy 
Maybe this is something we could implement like #933 ?

lavdnone (approved):
useful

mkuron:
I think this should just use the values from #933 and not have its own settings.

nnx-support:
Hi !

I was looking to do the exact same thing. Is there a chance for this PR to get merged? Should I open a new one?

@mkuron I noticed you say it should use the settings set in the WebUI but i think there is value in setting something different than the WebUI title. And also there is `displayName` and `displayShortName` while there is only one in the WebUI.

PR #929, opened 2018-01-17
Remove required from Add Domain - Description

Fvbor:
Die Description is not a required field for the creation of a domain.


andryyy:
The function must be fixed, too, to not delete the desc when the API call has no desc var set.

+ Please don't push to master.

Issue #931, opened 2018-01-18
Problems with PHP-FPM limits


feldsam:
Hello, today I experience problems with whole stack. Unresponsive Postfix and Nginx. I figure out, taht problem was caused by php-fpm hitting max children limit (5).

So I increase it to 10 and restart container. After a while problem was here again, with same message

    seems busy (you may need to increase pm.start_servers, or pm.min/max_spare_servers), spawning 32 children, there are 0 idle, and 9 total children
    server reached pm.max_children setting (10), consider raising it

Probably it is caused by debug.php page - I experience it three times after refreshing that page, so there is something, which takes long tim eto complete and use all available process.

Of course, I can try increase max children to 30 or 60, but we should solve root of problem.

I also propose to use more php-fpm pools (containers), at least two:
- one for basic mail server functionality - postfix, rspamd, etc.
- second one for web ui

feldsam:
I play a bit with pm values and create PR. This also increase max memory consuption

feldsam:
I investigate debug page a little bit, and see that all logs are downloaded via api immediately. That should be loaded only when user clicks in dropdown on some logs page.

Anyway, I still think, that pm.max_children 5 is small, especially when there are more calls from rspamd...

andryyy:
This should not happen and I cannot replicate it. It is - as I wrote - probably related to a reverse proxy setup. 

I am testing a separate worker for system stuff (dynamic) + web stuff (ondemand) right now. We can lazyload the tables, though.

andryyy:
Furthermore each child can handle a lot of requests (default should be around ~300?), it should never block 5 workers for 5 requests.

feldsam:
probably I miss-understood how it works, but I believe that if there are 5 slow requests, them all 5 children are in use. Am I wrong?

lavdnone:
Might also be worth putting auto recovery when dying in to `etc/php-fpm.d/www.conf` or `etc/php-fpm.d/system.conf`
helps to keep php up with sogo and activsync

     ; If this number of child processes exit with SIGSEGV or SIGBUS within the time
     ; interval set by emergency_restart_interval then FPM will restart. A value
     ; of '0' means 'Off'.
     ; Default Value: 0
     emergency_restart_threshold = 10
     
     ; Interval of time used by emergency_restart_interval to determine when
     ; a graceful restart will be initiated.  This can be useful to work around
     ; accidental corruptions in an accelerator's shared memory.
[...]

PR #932, opened 2018-01-18
[PHP-FPM] Updated pm config, increase values

feldsam:
Related to #931


andryyy:
Hi, first of all: thanks!
How about providing these options as env vars we could set in mailcow.conf? We should keep a default value (as is) and override them when set in mailcow.conf. Or we even move the file to data/conf...

feldsam:
What about mysql config, that also contains options like max-connections and max allowed packed and is not configurable via mailcow.conf

Look on related issue #931 please, there can be possible optimizations on debug page probably.

andryyy:
That's available from data/conf/mysql/my.cnf 👍 

I check the debug page.

feldsam:
ok, so I can do it similarly or what you suggest?

feldsam:
Closing this PR as I see it is already done better way in dev branch.

PR #934, opened 2018-01-18
[WEB] Make quarantine action column more wider

feldsam:
In same languages it was not enough


andryyy:
Thanks for checking that. 👍

PR #953, opened 2018-01-23
Expose Postfix's recipient_canonical_maps through web UI

mkuron:
This pull request makes `recipient_canonical_maps` configurable from the web UI. This option is similar to the recently-added BCC maps feature, but instead of adding recipients, it replaces the existing ones. In fact, I copied most of the BCC maps code (I couldn't reuse it directly because the logic is slightly different).

This can be useful to rewrite the recipients on outgoing messages, e.g. to replace _@oldcompany.de_ with _@newcompany.com_ or _@internaldomain.local_ with _@company.com_, for example because your users still have these old/internal addresses in their address book.


andryyy:
Very cool! Is it possible to base this on dev? :-)

mkuron:
Yes, we just need to merge master into dev first (they have diverged), then I can switch the base branch of this PR.

mkuron:
I did some more testing and the pull request is working fine. Please give it a try yourself, @andryyy, and then merge and increase the version number of the postfix container.

andryyy:
I will, thanks!!

andryyy:
Hm, you stopped translating it? And no domain check at all for domain admins?
I actually missed that for bcc, too (translating). :-( I will add the missing strings in a bit and add a domain column. It looks and works pretty good!

Aaaand I'm stupid. Forget the domain column.

mkuron:
>  And no domain check at all for domain admins?

Unfortunately these are global for the server. You can't have different rewrites for different users or domains. That's a Postfix limitation.

andryyy:
Yes, I understand. I was copying it for sender maps and stupidly assumed it was possible. Thanks! 👍

PR #972, opened 2018-01-28
Apple mobileconfig enhancements

feldsam:

### DIFF
--- a/data/web/inc/functions.inc.php
+++ b/data/web/inc/functions.inc.php
@@ -1183,4 +1183,19 @@ function get_logs($container, $lines = false) {
   }
   return false;
 }
+function getGUID() {
+  if (function_exists('com_create_guid')) {
+    return com_create_guid();
+  }
+
+  mt_srand((double)microtime()*10000);//optional for php 4.2.0 and up.
+  $charid = strtoupper(md5(uniqid(rand(), true)));
+  $hyphen = chr(45);// "-"
+
+  return substr($charid, 0, 8).$hyphen
+        .substr($charid, 8, 4).$hyphen
+        .substr($charid,12, 4).$hyphen
+        .substr($charid,16, 4).$hyphen
+        .substr($charid,20,12);
+}
 ?>
--- a/data/web/mobileconfig.php
+++ b/data/web/mobileconfig.php
@@ -12,158 +12,163 @@ if (!isset($_SESSION['mailcow_cc_role']) || $_SESSION['mailcow_cc_role'] != 'use
 error_reporting(0);
 
 header('Content-Type: application/x-apple-aspen-config');
-header('Content-Disposition: attachment; filename="Mailcow.mobileconfig"');
+header('Content-Disposition: attachment; filename="'.$UI_TEXTS['main_name'].'.mobileconfig"');
 
 $email = $_SESSION['mailcow_cc_username'];
 $domain = explode('@', $_SESSION['mailcow_cc_username'])[1];
-$identifier = implode('.', array_reverse(explode('.', $domain))) . '.iphoneprofile.mailcow';
+$identifier = implode('.', array_reverse(preg_split( '/(@|\.)/', $email))) . '.appleprofile.'.preg_replace('/[^a-zA-Z0-9]+/', '', $UI_TEXTS['main_name']);
 
 try {
   $stmt = $pdo->prepare("SELECT `name` FROM `mailbox` WHERE `username`= :username");
   $stmt->execute(array(':username' => $email));
   $MailboxData = $stmt->fetch(PDO::FETCH_ASSOC);
+  $displayname = empty($MailboxData['name']) ? $email : $MailboxData['name'];
 }
 catch(PDOException $e) {
-  die("Failed to determine name from SQL");
-}
-if (!empty($MailboxData['name'])) {
-  $displayname = $MailboxData['name'];
-}
-else {
   $displayname = $email;
+}
+
+if (isset($_GET['only_email'])) {
+  $onlyEmailAccount = true;
+  $description = 'IMAP';  
+} else {
+  $onlyEmailAccount = false;
+  $description = 'IMAP, CalDAV, CardDAV'; 
 }
 
 echo '<?xml version="1.0" encoding="UTF-8"?>' . "\n";
 ?>
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
-<dict>
-	<key>PayloadContent</key>
-	<array>
-		<dict>
-			<key>CalDAVAccountDescription</key>
-			<string><?php echo $domain; ?></string>
-			<key>CalDAVHostName</key>
-			<string><?php echo $autodiscover_config['caldav']['server']; ?></string>
-			<key>CalDAVPort</key>
-			<real><?php echo $autodiscover_config['caldav']['port']; ?></real>
-			<key>CalDAVPrincipalURL</key>
-			<string>/SOGo/dav/<?php echo $email; ?></string>
-			<key>CalDAVUseSSL</key>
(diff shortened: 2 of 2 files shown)


mkuron:
The changes look sensible. However, I don't think mounting the SSL keys directory into the PHP-FPM container is a good idea in terms of security (and it doesn't matter whether the private key is password-protected or not). The PHP interpreter is notorious for occasionally having security bugs that compromise files accessible to the web server, or there could be a path traversal issue lurking somewhere in Mailcow's PHP scripts, or something like that. **We should not take any chances in potentially exposing the web and mail server's SSL key.**

To make this secure, you should create a separate subdomain (e.g. `mobileconfig.$MAILCOW_HOSTNAME`), request a _separate_ certificate for that, store it into a separate directory (e.g. ./data/assets/ssl-signing/) and only expose that to the PHP container. You should then modify the DNS diagnostics page so that it also checks that `mobileconfig.$MAILCOW_HOSTNAME` is a CNAME for `$MAILCOW_HOSTNAME`.

feldsam:
Hi, thanks, can you participate too and help with that subdomain solution?

feldsam:
Hello, I see demand for this PR to be merged. @mkuron would you please help me with that separate cert? or I split this PR to two (general updates and signing)

mkuron:
@feldsam, I‘m sorry I didn’t have time for this yet. Splitting into two pull requests would be helpful so we can get the UUID changes etc. into the next release.

andryyy:
The signing should additionally be done using PHPs OpenSSL functions.

A random UUID sounds good.

feldsam:
Hi, I removed signing part from this PR, so ready to merge

feldsam:
@andryyy what you think by php openssl functions? We still need some key, or not?

andryyy:
Looks good, but don't just add a link to the mobileconfig php... I don't want people to complain about why their phone is now setup with IMAP instead of EAS. If you want this link to be available, set it up for both EAS and IMAP, please.
Also add a note that it is only meant for iOS devices.

andryyy:
Regarding OpenSSL functions: I won't ever merge proc_open or shell_exec stuff. I'd like to disable these functions alltogether sometime soon.

feldsam:
I see, ok, where I find EAS link? I don't know about EAS Apple autoconfig file.

proc_open - ok, I understand.

andryyy:
"com.apple.eas.account as the PayloadType" => https://developer.apple.com/library/content/featuredarticles/iPhoneConfigurationProfileRef/Introduction/Introduction.html

How about adding a _GET parameter like ?imap and ?eas

feldsam:
Ok, I do till monday resp. durring weekend

andryyy:
I can help this weekend. Or  try to. Time... :-(

mkuron:
I‘d just remove the link. We already link to the mobileconfig from the documentation. We don‘t have direct links to the setup of other clients in the web UI either.

For EAS, you don‘t really need a mobileconfig as it is simple enough to set up manually (through autodiscover).

feldsam:
Ok, I removed it

extremeshok:
What still needs to be done to get this merged ?
Once its done, we can work on adding the signing part using a separate signing domain (mobileconfig.domain)

data/web/inc/functions.inc.php
```diff
@@ -1183,4 +1183,19 @@ function get_logs($container, $lines = false) {
   }
   return false;
 }
-?>
+function getGUID() {
+  if (function_exists('com_create_guid')) {
+    return com_create_guid();
+  }
+
+  mt_srand((double)microtime()*10000);//optional for php 4.2.0 and up.
```
andryyy, line 1191: What is that good for?

feldsam: Seeds the Mersenne Twister Random Number Generator

feldsam: but for just this project we can use just com_create_guid()

data/web/inc/functions.inc.php
```diff
@@ -1183,19 +1183,8 @@ function get_logs($container, $lines = false) {
...
-  mt_srand((double)microtime()*10000);//optional for php 4.2.0 and up.
-  $charid = strtoupper(md5(uniqid(rand(), true)));
-  $hyphen = chr(45);// "-"
-
-  return substr($charid, 0, 8).$hyphen
-        .substr($charid, 8, 4).$hyphen
-        .substr($charid,12, 4).$hyphen
-        .substr($charid,16, 4).$hyphen
-        .substr($charid,20,12);
+function getGUID($id) {
```
andryyy: To prevent adding multiple configs for the same identity.

feldsam: ok

feldsam: but it should not be GUID named

feldsam: I think, that it is unnecessary change. According to docs `A globally unique identifier for the payload. The actual content is unimportant, but it must be globally unique. In macOS, you can use uuidgen to generate reasonable UUIDs.`

andryyy: Using the same UID overrides profiles, doesn't it?
edit: Or it prevents them from being added.

feldsam: you have duplicates in one profile now

feldsam: I personaly not tested replacing profiles, but if yes, them should be only on line 160 and not on all UUIDs

feldsam: I know, that if profile name is same, it not allow to install it.

feldsam: I tested it on iPad, and my version with GUID works fine

feldsam: I tested it now, it it don't allow to install same profile twice, no matter if guid si same or not. It just checks `PayloadIdentifier`

feldsam: So I propose to revert that changes and keep for ex. just that `com_create_guid` function

andryyy: Please respond in a single comment.....

andryyy: Okay, I see. So a random GUID is not a problem, but the PayloadIdentifier needs to be different for each profile. I have seen an issue which erroneously made me assume it was the GUID, sorry.

Can you revert my changes in your pr?

Just use `<?= xy; ?>` and 2 space indents.

feldsam:
I reverted your changes, so it is ok now, or do you want to change something? I can do that now

edit: I go to try use just com_create_guid
edit: Tested and found, that we don't have com_create_guid available in php, so I think that this can be final version. I tested various scenarios on my iPad and iPhone before contributing

edit: just one thing, I thinking about is ability to configure just IMAP, because not all customers wants to use calendar and contacts, especially when have multiple accounts

andryyy:
Hm, if we do this, another version with EAS would be super nice! :-)

feldsam:
no problem, but @mkuron have comment on it :) sooo I suppose to merge this PR and make new one for EAS and just IMAP config. To move things on.

andryyy:
Well, he is not wrong, configuring EAS is very easy with autodiscover. Could be added as optional feature though. Can you do this? Like bla.php?eas etc.?

Please do me a favor and change `<? echo ` to `<?= bla` etc. and use 2 space indents :-) I will be afk for a few hours now, we can merge it then.

feldsam:
Ok, I do

feldsam:
@andryyy reminding myself

andryyy:
@mkuron You think that's fine?

mkuron:
@andryyy, looks good to me.

vain90:
Hi,
this PR was only merged into the dev branch.
Are there any plans when this PR will be merged into the master branch?

andryyy:
Just forgot about it. :-(

strarsis:
So there is a PR for https://github.com/mailcow/mailcow-dockerized/issues/3681 ? Awesome, can this be merged into main branch?

PR #975, opened 2018-01-29
Index performance fsync never for lmtp  optimized

lavdnone:
Per https://wiki2.dovecot.org/MailLocation/LocalDisk#preview
lmtp  optimized is enough to save from power loss
works with no issues, tested over gluster mounts cluster
Ideally we should do separate dir for mail index as a setting as standard setup is to put those to SSD mounts to be separated from mail files. Index is also auto recoverable.

### DIFF
--- a/data/conf/dovecot/dovecot.conf
+++ b/data/conf/dovecot/dovecot.conf
@@ -7,7 +7,7 @@ log_path = syslog
 disable_plaintext_auth = yes
 # Uncomment on NFS share
 #mmap_disable = yes
-#mail_fsync = always
+mail_fsync = never
 #mail_nfs_index = yes
 #mail_nfs_storage = yes
 login_log_format_elements = "user=<%u> method=%m rip=%r lip=%l mpid=%e %c %k"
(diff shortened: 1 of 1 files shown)


andryyy:
"works with no issues" on power loss? How many times have you tested a power loss? :-O

We will not require users to setup a second volume on a SSD, sorry. This isn't "standard setup". It is an advanced setup on heavy-load machines, where you want a very fast index and a low-op storage to prevent disk failures.

lavdnone:
Good question. Certainly killed vms many times on tests - some times got delivery reties or fail over to different node with MX record. Guess just common sense and dovecot wiki: in plane mailcow case - if the only server will loose power there is no point to worry about one email file tiny chance to be lost (there will be other things to worry about), and if transfer didn't finish sender will repeat attempt.

Regarding separating index to SSD (or not SSD) - seems useful to make it configurable through mailcow config settings as other things. It is more sort of mature feature - maybe it will push starting guys to make better setups while figuring out what the setting is. By default disabled.

PR #995, opened 2018-02-02
SOGO & Rspamd interface: adding "expire" header to static files, allowing browser to be able to cache them

Alireza2n:
Hi,
I noticed there are no "Expire" headers for static content in Rspamd interface and SOGO.
In Rspamd, This causes RequireJS errors due to static files being loaded late (in case of a slow server or low bandwidth).
Also speeds up SOGO loading on slower links.
I have added a map to nginx allowing static content to be cached by browser.


andryyy:
Thanks!

PR #1008, opened 2018-02-07
Support for multiple accounts on device

extremeshok:
** I agree to the code of conduct and the contributory guidelines **

# This fixes the apple/ios/osx limitation where only a single profile and email account can be used.

Screenshots:  https://imgur.com/gallery/Ck41e?s=wa

This code is being used in our production systems, with more than 1000 accounts and 15 million emails.

All input variables are sanitized along with extensive error checking to prevent broken profiles.

PayloadUUID are dynamically generated to prevent UUID collisions.

Original functionality has been kept.

# Use cases
1.  Allows a link to be sent to a client/device which can be used to download their profile and configure their device without needing to login, ie via a sign-up email, 
2.  Client requires sales@companydomain.com and their personal account@companydomain.com.
3.  Client has 2 separate domains and wants to keep the accounts separate, sales@domain.com, sales@otherdomain.com

# Test cases and valid results
## multiple accounts using the same domain
````
mobileconfig.php?user=sales,admin,,,@samedomain.com   
valid=sales@samedomain.com & admin@samedomain.com
````
## multiple accounts with different domains
````
mobileconfig.php?user=sales@domain.com,admin@otherdomain.com,,@domain.com
valid=sales@domain.com & admin@otherdomain.com
````
## single account
````
mobileconfig.php?user=sales@domain.com
valid=sales@domain.com
````
## original usage
````
mobileconfig.php 
valid only if user is logged in, uses the users email address)
````

### DIFF
--- a/data/web/mobileconfig.php
+++ b/data/web/mobileconfig.php
@@ -4,34 +4,60 @@ require_once 'inc/prerequisites.inc.php';
 if (empty($mailcow_hostname)) {
   exit();
 }
-if (!isset($_SESSION['mailcow_cc_role']) || $_SESSION['mailcow_cc_role'] != 'user') {
-  header("Location: index.php");
-  die("This page is only available to logged-in users, not admins.");
-}
 
 error_reporting(0);
 
+// mobileconfig.php?user=sales,admin@samedomain.com   (valid=sales@samedomain.com & admin@samedomain.com)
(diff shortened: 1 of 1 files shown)


mkuron:
I prefer the solution from #972, it generates separate profiles (but with distinct IDs and UUIDs) for every address. Also, it’s a much simpler change.

Your solution requires the user to manually edit query strings in URLs, which is not exactly user friendly.

Issue #1011, opened 2018-02-08
imapsync failing for lots of folders


jhesketh:
When using imapsync with lots of folders imapsync_cron.pl fails to save the output as it is too long:

`DBD::mysql::st execute failed: Data too long for column 'returned_text' at row 1 at /usr/local/bin/imapsync_cron.pl line 102.`

This means the result isn't saved and the UI thinks the job is still running.

The migration is actually successful though and because imapsync_cron.pl doesn't check for is_running (just the timestamp) it'll reschedule it immediately (even if the time elapsed between runs hasn't occurred).

stevesbrain:
Correct me if I'm wrong, but wouldn't this be an imapsync specific issue (and therefore a problem for their repo, rather than mailcow)?

jhesketh:
No, it's to do with storing the result in mailcow's database. The output is captured in imapsync_cron.pl and stored in mysql. I have a fix incoming.

stevesbrain:
Ah, no worries - I'll leave you to it :)

PR #1012, opened 2018-02-08
Fix imapsync for large mailboxes

jhesketh:
Where there is a lot of folders in a mailbox the output may be larger
than TEXT will allow (64KiB). Instead use MEDIUMTEXT (16MiB) which
should be ample.

Additionally check that a sync isn't already running before starting
a new one.

Fixes: #1011


andryyy:
Thanks! Just added a new date to version to init the database.

I will add a new image very soon.

andryyy:
Unknown option: buffersize 8192000
Unknown option: skipheader 'x-*'
Unknown option: split1 3000
Unknown option: split2 3000

.......................................

PR #1015, opened 2018-02-09
Remove unused language keys from data/web/lang/lang.*.php

descala:
helper-scripts/check_translations.rb checks if keys present in the master
translation file are used in any php or js code.  Writes a sed command to
stdout to remove unused keys from all lang files.

This saves unnecessary work when translating.


PhoenixPeca:
Looks good to me.

andryyy:
And it fails.
grafik

Please push a fix. **Please test**!

descala:
Tested and pushed

PR #1037, opened 2018-02-13
Remove "empty" folders

klausenbusk:
There seems to be no reason for this empty folders.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,8 +3,7 @@ data/conf/sogo/sieve.creds
 data/conf/dovecot/dovecot-master.passwd
 mailcow.conf
 mailcow.conf_backup
-data/conf/nginx/listen*active
-data/conf/nginx/server_name.active
+data/conf/nginx/*.active
 data/conf/postfix/sql
 data/conf/dovecot/sql
 data/conf/nextcloud-*.bak
--- a/.gitignore
+++ b/.gitignore
@@ -3,8 +3,7 @@ data/conf/sogo/sieve.creds
 data/conf/dovecot/dovecot-master.passwd
 mailcow.conf
 mailcow.conf_backup
-data/conf/nginx/listen*active
-data/conf/nginx/server_name.active
+data/conf/nginx/*active
 data/conf/postfix/sql
 data/conf/dovecot/sql
 data/conf/nextcloud-*.bak
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,7 @@ data/conf/sogo/sieve.creds
 data/conf/dovecot/dovecot-master.passwd
 mailcow.conf
 mailcow.conf_backup
-data/conf/nginx/*active
+data/conf/nginx/*.active
 data/conf/postfix/sql
 data/conf/dovecot/sql
 data/conf/nextcloud-*.bak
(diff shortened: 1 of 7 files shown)


PhoenixPeca:
...Git cannot track empty folders (or any kind of folder) in the first place. Git only tracks files, the folders are incidental. Git only creates folders on disk in order to house files; a folder with no files in it does not exist in Git's world-view.

If you want to keep an "empty" folder in your Git repository, the convention is to add a hidden empty file to that directory, such as /java/.gitkeep

See this [Stack Overflow thread](https://stackoverflow.com/questions/25376342/does-git-delete-empty-folders)

PhoenixPeca:
Git is a content tracker. Hence, directories aren't indexed by it.

klausenbusk:
> Git is a content tracker. Hence, directories aren't indexed by it.

I'm aware of that, but why do we need this folders? I can't find any reference to them.

PhoenixPeca:
Yep, as of the moment, there are no crucial uses for these folders. However, future update might do. So, I think it's better to keep these folders in place. Just in case. Besides, these folders doesn't really take up huge disk space. 😉

klausenbusk:
> Yep, as of the moment, there is no crucial uses for these folders. However, future update might do. So, I think it's better to keep these folders in place. Just in case.

I disagree with that, from my point of view it is just clutter and confuse people (I would expect them to have some sort of purpose). I would just remove them, if we need them at some point in the future we can always readd them.

Just my 2 cent..

PhoenixPeca:
>...it is just clutter and confuse people...

That makes sense.

andryyy:
What the hell did you commit?! Have you even checked your commit?!

andryyy:
We will to merges on their own. Please don‘t commit changes like that. This can potentially break stuff.

These placeholders were part of mailcows structure you shouldn’t just change without asking for it in an issue. I agree we should use gitkeep, but removing them sucks. I disagree that they are confusing. We can point to the external Dockerfiles, if you want to.

klausenbusk:
> What the hell did you commit?! Have you even checked your commit?!
> These placeholders were part of mailcows structure you shouldn’t just change without asking for it in an issue.

Using a PR to ask or a issue should be the same. If not even the reviewer can see the issue, how I'm supposed to?

> I agree we should use gitkeep, but removing them sucks.

Why? What is the purpose of this "placeholders"?

andryyy:
You carried all commits from master to dev. That was not a problem today, but it can break things someday. The current git handling sucks for sure, and we will change that in the future. Sometimes changes to master are temp. fixes and handled differently in dev, which will then be merged to master. 

These are placeholders for non-mailcow Dockerfiles, though still (as of today) empty. You can happily delete them in your repository. As long as you aren't sure wether or not we will use them one day, you should leave the structure as-is. And that's why an "issue" should be preferred, just ask us... :-)

klausenbusk:
> You carried all commits from master to dev. 

I'm sorry about that, but the review should have caught that. I opened this PR against master, and it was changed to dev and merged before I had a chance to do anything. I apologise :)

> And that's why an "issue" should be preferred, just ask us... :-)

From my point of view asking in a issue or opening a PR and discuss as part of the review is near the same. But I will open a issue next time.

PR #1041, opened 2018-02-14
Fixes update.sh issues for Alpine

miclnl:
Alpine linux as a host works, but requires some tweaking on update.sh. But the SHA-checks on update.sh prevent any local changes.

* pip must be used for upgrading/installing docker-compose; the default upgrade method breaks this
* `cp -n` does not exist in busybox. using `false | cp -i` is a universally available alternative

For further info on docker on Alpine see https://wiki.alpinelinux.org/wiki/Docker

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -183,7 +183,7 @@ fi
 
 echo -e "\e[32mFetching new docker-compose version...\e[0m"
 sleep 2
-if [[ -f "/etc/alpine-release" ]]; then
+if [[ ! -z $(which pip) && $(pip list --local | grep -c docker-compose) == 1 ]]; then
   # Running on Alpine host, use pip to upgrade
   pip install docker-compose -Uq
 elif [[ $(curl -sL -w "%{http_code}" https://www.servercow.de/docker-compose/latest.php -o /dev/null) == "200" ]]; then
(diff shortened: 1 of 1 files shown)


andryyy:
Cool, thanks. But shouldn't we check if docker-compose was installed via pip?

miclnl:
That's a great idea! But I'm not sure how.
`pip list --local | grep -c docker-compose`  gives a '1' on my setup, but I'm not sure what it does when docker-compose isn't installed via pip, could anyone check?

andryyy:
```
if [[ ! -z $(which pip) && $(pip list --local | grep -c docker-compose) == 1 ]]; then
  echo "installed via pip"
fi
```

Should work. Your command gives me a 0. :-)

miclnl:
That'll work, I've edited the pull :)

extremeshok:
Some errors are not with the script, just the way busybox uses lite versions of the common utilities.

These are the actual fixes
```
#add support for cp -n and grep -P
apk add --upgrade coreutils
apk add --upgrade grep
```
+1 for
```
if [[ ! -z $(which pip) && $(pip list --local | grep -c docker-compose) == 1 ]]; then
  echo "installed via pip"
fi
```

miclnl:
My main goal was to add the pip exception. But I figured I could modify the other commands as well. As I see it this patch increases the compatibility without making compromises. 

I did overlook the `grep -P`, would you agree with:
`grep -o "image: mailcow.*" docker-compose.yml | cut -d ' ' -f2`

PR #1042, opened 2018-02-14
Fix for large Mailboxes to avoid timeouts

extremeshok:
** I agree to the code of conduct and the contributory guidelines **

Tested with 80+GB single inbox via imap

``` --nofoldersizes --skipsize --fast ``` file and folder sizes are only used for statistics, which are completely useless. Before the actual data is transfered the message and folder sizes need to be calculated on a 200 000+ message inbox, this will almost never complete.

``` --buffersize 8192000 ``` sets the io buffer to 8mb, the default buffer is 4kbyte, this is speeds up syncs.

```  --skipheader 'X-*' ``` X headers vary wildly from system to system and do not ensure message uniqueness, they are not needed.

```  --split1 3000 --split2 3000 ```    split the requests in several parts on the server, 3000 is the number of messages handled per request.

``` --fastio1 --fastio2 ``` use fastio


andryyy:
Nice! Thanks.

jhesketh:
Even if the headers are otherwise identical, wouldn't you still want to sync across a message that had different X- headers? In my opinion I would prefer to err on the side of caution and copy any messages I'm unsure about an let a user figure out any duplicates (or use their headers as they need etc).

andryyy:
Hm, that’s a good point

andryyy:
That was untested stuff and did break master.

PR #1043, opened 2018-02-14
Fix for large Mailboxes to avoid timeouts 2

extremeshok:
Missing ' (sorry my editor keeps correcting the '" to " )


andryyy:
Please don't push to master if it's not what you tested.

PR #1045, opened 2018-02-14
Enable maildir compression

extremeshok:
Currently the plugin is loaded, but actual compression is not enabled.

https://wiki.dovecot.org/Plugins/Zlib


andryyy:
That's up to each admin. A default compression method should also be discussed.

extremeshok:
Compression is transparent to the admin and users.

lz4 is the fastest on any modern day cpu. This saves storage and decreases I/O as compression is done in memory.

PR #1047, opened 2018-02-14
Support for alpine linux

extremeshok:
detects if cp and grep are the non BusyBox versions
No longer breaks docker-compose when installed via pip
minor fixes with missing if ; then ; fi for some statements, to improve code consistency

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -1,10 +1,25 @@
 #!/bin/bash
 
 for bin in curl docker-compose docker git awk sha1sum; do
-  if [[ -z $(which ${bin}) ]]; then echo "Cannot find ${bin}, exiting..."; exit 1; fi
+  if [[ -z $(which ${bin}) ]]; then
+    echo "Cannot find ${bin}, exiting..."
+    exit 1
+  fi
 done
 
-[[ ! -f mailcow.conf ]] && { echo "mailcow.conf is missing"; exit 1;}
+if [[ ! -f mailcow.conf ]] ; then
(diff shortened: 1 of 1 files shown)


nightah:
Can we add the `--no-cache` option to the apk command on these lines:
https://github.com/extremeshok/mailcow-dockerized/blob/3ec50d23ae40a3f8477c9a2cd104c0de3c28f972/update.sh#L16-L20

The `--no-cache` option allows you to not cache the index locally, which is recommended for keeping your containers small.

Similarly we would also want to change the references in #1048 too.

andryyy:
Why do you remove the "short ifs"? I use them quite a lot. The added space before ";" is not used at all. I don't see how that improves code consistency. :-/

Can you please just add the busybox stuff? I think you can remove the "head", too and just grep the help, it is not that much to read. :-)

extremeshok:
I generally prefer the longer format, for the following reasons:
It is more readable and less cryptic.
Code tracks easier on commits, you see just the line change and not the entire if statement change
It allows for more complex functionality to be added in the future.
People with less bash experience find it easier to understand the code.

--

> I think you can remove the "head", too and just grep the help, it is not that much to read. :-)
I do not understand this ?

andryyy:
I do it a lot in many sh files in mailcow, I don't really want to change that. :-/ Would be nice if you just change it back and only fix it for Alpine. Please also rebase on dev.

extremeshok:
Rebased on dev, and made the corrections
#1071 replaces this

Issue #1044, opened 2018-02-14
Strange DNS requests


levinuss:
Hey,
I discovered some strange DNS requests in a tcpdump today (see below). Because the the requested domain is dockerapi or redis and the DNS resolvers asked for the records are the google public dns servers (and do not use the google ones on my system) I figured the requests come from mailcow.
```
12:57:58.244804 IP6 xx:xx.56052 > google-public-dns-a.google.com.domain: 36872+ A? dockerapi. (27)
12:57:58.244865 IP6 xx:xx.56052 > google-public-dns-b.google.com.domain: 36872+ A? dockerapi. (27)
12:57:58.244913 IP6 xx:xx.56052 > google-public-dns-a.google.com.domain: 37224+ AAAA? dockerapi. (27)
12:57:58.244943 IP6 xx:xx.56052 > google-public-dns-b.google.com.domain: 37224+ AAAA? dockerapi. (27)
12:57:58.245616 IP6 google-public-dns-b.google.com.domain > xx:xx.56052: 36872 NXDomain 0/1/0 (102)
12:57:58.245634 IP6 google-public-dns-a.google.com.domain > xx:xx.56052: 36872 NXDomain 0/1/0 (102)
12:57:58.245700 IP6 google-public-dns-b.google.com.domain > xx:xx.56052: 37224 NXDomain 0/1/0 (102)
12:57:58.245713 IP6 google-public-dns-a.google.com.domain >xx:xx.56052: 37224 NXDomain 0/1/0 (102)
```
Is this mailcow? Is this the intended behavior?

levinuss:
I did a little bit of research myself and found out that the traffic comes from the watchdog container. I looked into the resolv.conf of the watchdog container and found `127.0.0.11` and the IPv6 addresses of the google public dns. The `127.0.0.11` seemed to work, only requests over v6 (to the google DNS) failed and were repeated because they generated a `NXDomain`.
I discovered that the `watchdog-mailcow` container had no DNS servers as parameters in the `docker-compose.yml`. So I decided to add one (copied it from another container). The Google DNS servers are gone and testing with `dig redis` and `dig dockerapi` worked fine.
I wrote a little PR to add the DNS parameter against master. I hope thats okay. See  #1052

andryyy:
[...]

PR #1052, opened 2018-02-15
Added dns option for watchdog

levinuss:
I added the dns option for watchdog because it would query the Google Public DNS if using IPv6 otherwise. I described the problem in #1044 
Because its a fix and I tested it I thought its okay to merge it into master.


andryyy:
Can you recreate that PR? I found another issue with watchdog failing to resolve dockerapi every nth time in a while. Must be due to Dockers internal round robin? Don't know. This PR would fix it, though we rely on Unbound then.

levinuss:
Of course

PR #1056, opened 2018-02-15
[Nginx] Reduce config duplication

klausenbusk:
It does not make sense having a seperate server block for both http
and https.
According to the nginx doc [1], using the same server block for both
should work.

[1] http://nginx.org/en/docs/http/configuring_https_servers.html#single_http_https_server

---

I'm using this configuration right now, and it seems to work fine.

### DIFF
--- a/data/conf/nginx/site.conf
+++ b/data/conf/nginx/site.conf
@@ -21,182 +21,6 @@ server {
   charset utf-8;
   override_charset on;
 
-  add_header Strict-Transport-Security "max-age=15768000; includeSubDomains";
-  add_header X-Content-Type-Options nosniff;
-  add_header X-XSS-Protection "1; mode=block";
-  add_header X-Robots-Tag none;
-  add_header X-Download-Options noopen;
-  add_header X-Frame-Options "SAMEORIGIN";
-  add_header X-Permitted-Cross-Domain-Policies none;
-
-  index index.php index.html;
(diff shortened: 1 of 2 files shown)


andryyy:
It makes sense and will break some setups without two server blocks, we tested this a while ago and run into some problems. Can't find the issue rn. I think it broke behind some reverse proxies?

klausenbusk:
> we tested this a while ago and run into some problems. Can't find the issue rn. I think it broke behind some reverse proxies?

Interesting, I found the commit a3e966696f6bcc8b871a76ed7811cc465171627d, but as the commit message lacks the "why" that doesn't really help.

Looking through the commits, the commit messages could be improved. Feel free to read: https://chris.beams.io/posts/git-commit/

klausenbusk:
> we tested this a while ago and run into some problems

I assume you are referring to this commit: c5054ae7eddec34ceb42863c2653504dd1ee6359, which was incorrectly implemented (you forgot to remove `ssl: on` and add `ssl` to https `listen`), so it effectively broke plain http.

andryyy:
I tested it with a working config later... working as in "I removed the crap". It still had some downsides I don't want to dig out right now.....

Please read the open issues about how we plan to change git handling sometime soon....

andryyy:
Hmm, let's give it a try.

PR #1070, opened 2018-02-20
BugFix: Outlook 2016 not using EAS

extremeshok:
Outlook 2016 autoconfig will not be EAS enabled, even though "$autodiscover_config['useEASforOutlook'] = 'yes';

Outlook 2016 gives the HTTP_USER_AGENT string of
```
Microsoft Office/16.0 (Windows NT 10.0; MAPI 16.0.9001; Pro)
```

Limiting the regex to only 15, causes the entire if statement to fail.

For future proofing, this has been set to any version string containing 15,16,17,18,19

This has been tested using the "Test Email AutoConfiguration for Outlook 2016"


andryyy:
Thanks

PR #1077, opened 2018-02-22
Update generate_config.sh

idenkov:
Creating the log files for clamav, without them it gets stuck in a loop and never update the DB.

### DIFF
--- a/generate_config.sh
+++ b/generate_config.sh
@@ -117,3 +117,8 @@ mkdir -p data/assets/ssl
 
 # copy but don't overwrite existing certificate
 cp -n data/assets/ssl-example/*.pem data/assets/ssl/
+
+# clamav is trying to write to log files and stucks in a loop
+mkdir -p /var/log/clamav/
+touch /var/log/clamav/clamd.log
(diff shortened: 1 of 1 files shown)


andryyy:
I don't understand, why do you need to create them on your host?

We create those files as links to pipes inside the container. :-)

tkorves:
Doesn't make sense - will not be added.

idenkov:
You are right, however it was late and the logs were saying this so I didn't put much thinking. And after creating it on the host errors stop showing, which is weird.
However the container don't have that /var/log/clamav folder, there is touch for these two files in the bootstrap.sh but that does nothing because the directory does not exist. Will submit another proper pull when I get a bit of time.

andryyy:
It does not care about what is on your host.

Please don't push stuff like that...

There is a relict from older versions which tries to chown files that are not necessary, I removed this line. It is a useless line that does nothing.

idenkov:
My previous comment is about the container not the host. What line do you talk about that you removed? Agree the pull was stupid, it was the last thing I did before going to bed.

PR #1083, opened 2018-02-24
Catalan language support

descala:
Catalan shows up as "Unknown language" in the language selection drop down because it is not included in usrz/bootstrap-languages.


andryyy:
Very nice, thank you!

PR #1084, opened 2018-02-24
Adds catalan language flag

descala:
A side efect is the IconDrawer flag images are replaced by the ones from
country-flag-icons by Wil Linssen, which in turn come from the Wikipedia

And the US flag is relaced by the UK flag for english


treed593:
How can we revert the English flag back to the US flag? @andryyy / @descala

Issue #1085, opened 2018-02-24
Inbound & Outbound Relaying


nupplaphil:
I'd like to use mailcow as a relay for inbound and outbound mails for
- antispam
- antivirus
- DKIM
- SPF
- DMARC

Therefore I will extendend the Backup MX function.

My first guess is to use the "relayhost" of an domain and add the following three states:
- Use for outbound (the current usage)
- Use for inbound (which I need for the transport)
- Both

Next I change some .cf - Files of postfix to force the "sender_dependent_default_transport_maps" for the inbound mails rather than the original MX-Record.

@andryyy is that the right direction for such a change?

nupplaphil:
See #1136 for further discussion

PR #1086, opened 2018-02-24
[Postfix][Web] Relay Inbound and Outbound mails per domain

nupplaphil:
As described in #1085 , I wrote an enhancement to relay inbound mails too.


data/web/lang/lang.en.php
```diff
@@ -341,6 +341,8 @@
 $lang['edit']['backup_mx_options'] = 'Backup MX options';
 $lang['edit']['relay_domain'] = 'Relay domain';
 $lang['edit']['relay_all'] = 'Relay all recipients';
+$lang['edit']['relay_sender'] = 'Relay all sending mails to the Relayhost';
+$lang['edit']['relay_recipient'] = 'Relay all reciving mails to the Relayhost';
```
paescuj: Should be "receiving" instead of "reciving"

data/web/lang/lang.de.php
```diff
@@ -340,6 +340,8 @@
 $lang['edit']['backup_mx_options'] = 'Backup MX Optionen:';
 $lang['edit']['relay_domain'] = 'Relay Domain';
 $lang['edit']['relay_all'] = 'Alle Empfänger-Adressen relayen';
+$lang['edit']['relay_sender'] = 'Alle gesendete Mails an den Relayhost relayen';
+$lang['edit']['relay_recipient'] = 'Alle empfangene Mails an den Relayhost relayen';
```
paescuj: I'd rather use:
- Alle gesendeten E-Mails an den Relayhost weiterleiten
- Alle empfangenen E-Mails an den Relayhost weiterleiten

nupplaphil: Sounds better, you're right.

data/web/lang/lang.en.php
nupplaphil: Arg, I hate typos ;-)

nupplaphil:
Oups, I' afraid that I found a mistake.

The query at "transport_maps" returns "smtp:" even when "relay_recipient" is '0' , so postfix tries to relay EVERY mail, becaus 'smtp:' simply means use the MX-Record.

I'm currently on the way to fix it!

data/Dockerfiles/postfix/postfix.sh
```diff
@@ -74,6 +75,48 @@ query = SELECT GROUP_CONCAT(transport SEPARATOR '') AS transport_maps
   AS transport_view;
 EOF
 
+cat <<EOF > /opt/postfix/conf/sql/mysql_recipient_dependent_default_transport_maps.cf
+user = ${DBUSER}
+password = ${DBPASS}
+hosts = mysql
+dbname = ${DBNAME}
+query = SELECT IF(EXISTS(SELECT 'relay' FROM domain
```
nupplaphil, line 83: This statement looks very ugly, but for me it was the easiest way to fix the query-issue.
Maybe somebody with more SQL-expierence can check it.

andryyy: There must be another solution for that. That'd be a much too complex query for _every_ mail.

andryyy: Wait, can't we move sender_dependent_bla to transport_map anyway? Should be possible.

andryyy (commented):
Please always only commit what you actually test on. This PR does not change init_db version, which results in never actually changing the database. You may always have a typo here you never catch. 👍 

Other than that it should always be based on dev (please don't just change the branch now, this will merge all of master to dev).

SamuelNitsche:
I also worked on a solution to relay inbound mails. You may have a look at [https://github.com/SamuelNitsche/mailcow-dockerized/tree/master](https://github.com/SamuelNitsche/mailcow-dockerized/tree/master)

data/Dockerfiles/postfix/postfix.sh
nupplaphil: The problem of sender_dependent_transport_maps.cf is, that it **always** responses "smtp:" and transport_map interprets that it should use the MX record of the domain, so I had to change it to respond "NULL" in case there is no transport_map.

nupplaphil:
@andryyy I changed the init_db.inc.php version, thanks fo the hint 
I will switch to "dev", so it means that this merge request can be closed, can't it? Damn ;-)

[Edit] Should I now make a Pull Request to "dev" or just checkout "dev", make changes and make a pull request to "master"? I'm sorry I'm not very experienced with github :-(

PR #1088, opened 2018-02-26
fix maxbytespersecond issue

Lucky88888:
Fixed $maxbytespersecond configurable option. This was still not working because it took the DB value of maxage. 
"First GitHub update for me, so please tell if I should have something different. ;) "


andryyy:
Thanks, that's a good catch... You'd need to update docker-compose.yml, but you cannot push images, so I will add it afterwards.

Issue #1046, opened 2018-02-14
Enable maildir compression


extremeshok:
Pull Request #1045 , was reject as there needs to be a discussion regarding compression.

discussion to follow---->

Currently the plugin is loaded, but actual compression is not enabled.

I suggested compression be enabled by default, which is transparent to the admin and users.

I also suggested lz4 algorithm be used, as the current dovecot version and docker libraries support this.

lz4 is also the fastest on any modern day cpu. 

The result of the pull would save storage and decrease I/O. 

I will also submit a custom written script to compress all existing non-compressed mails, if this becomes enabled by default.

catcun:
I am all in for this as it will help me a lot. But if I understand, I only need to change dovecot.conf and restart the container. Also, I would need to compress existing messages.
That said, I see no problem with it being enabled by default as it may be useful to many admins, but being so easy to setup perhaps a variable in mailcow.conf would do (the way Let's Encrypt is being handled.)

apintocr:
I also vote for compression by default.
However I believe there must be an option to disable it.

nightah:
Think the optional route is the best way to go, a lot of people could be using ZFS storage backends with compression enabled on their dataset, no point with double compression.

lavdnone:
lz4 default makes sense
@nightah  It would be funny if zfs translators would not figure out that files are already compressed and would try to compress them again. This would be an issue which we should file with zfs guys. Couldn't find the quick answer but I don't think thats the case. Real question might be who would do it better. And if you consciously using zfs compression you would figure out to change default setting in mailcow.config to not do compression.

nightah:
@lavdnone, I'm sure there's likely benchmarks out there which would give a rough idea.

The point I was trying to make was that whatever implementation is done, I believe there should be a switch of some sort to enable/disable (even if the default assumption is that it is enabled).
Administrators might have many cases outside of what I presented above as to why they may not want compression enabled give them the power and the option.

catcun:
Perhaps we don't even need the option in mailcow.conf, just document the possibility and let every admin decide.

mritzmann:
Can i restore my maildir with compression to another email software (or thunderbird with maildir) or is compression there a problem?

lavdnone:
when restoring dovecot will auto-detect compressed files by first bites with no issues, not sure about thunderbird

mritzmann:
[...]

PR #1090, opened 2018-02-26
Enable maildir compression

extremeshok:
Currently the plugin is loaded, but actual compression is not enabled.
https://wiki.dovecot.org/Plugins/Zlib

Notes: lz4 has about 2:1 compress-ratio on text
It will not double compress on zfs

Re-submitted for Inclusion as per discussion Enable maildir compression #1046


andryyy:
I removed zlib_save_level, thanks.

extremeshok:
> I removed zlib_save_level, thanks.
no problem, btw default is 6

PR #1094, opened 2018-02-27
Reorder navigation items

paescuj:
This PR reorders the navigation items for the configuration dropdown and the debugging tabs for the following reasons:
* The default page for admins is "Administration", therefore it should be the first item in the configuration dropdown
* It makes more sense to show the container overview in the debug page by default


andryyy:
Thanks, that's what I wanted to do like 20 times and always forgot... 👍

PR #1117, opened 2018-03-01
Fix pull request  #1015

descala:
I was assuming the section in $lang[section][i18n_string] was the name
of the file inside data/web/js

But this is not the case in debug.js


andryyy:
\o/ Thanks!

PR #1135, opened 2018-03-03
Add checks for total RAM of the server 

MAGICCC:
We had several users without enough RAM lately. This check will either disable or enable clamd where disabling has higher priority
At the moment the threshold is set to 1500MiB, which is 1572864kb

### DIFF
--- a/generate_config.sh
+++ b/generate_config.sh
@@ -40,7 +40,8 @@ else
 fi
 
 if [ $(awk '/MemTotal/ {print $2}' /proc/meminfo) -le "1572864" ]; then  #this is 1500Mib converted into kb which /proc/meminfo reports in
- read -r -p "The server you're using don't have enough RAM to run ClamAV in stable conditions! Do you want to disable the service? You can enable it in mailcow.conf later. [Y/n] " response
+  echo "Installed memory is less than 1500 MiB. It is recommended to disable ClamAV to prevent out-of-memory situations."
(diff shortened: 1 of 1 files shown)


generate_config.sh
```diff
@@ -30,7 +30,7 @@ fi
 if [[ -a /etc/timezone ]]; then
   TZ=$(cat /etc/timezone)
 elif  [[ -a /etc/localtime ]]; then
-   TZ=$(readlink /etc/localtime|sed -n 's|^.*zoneinfo/||p')
```
MAGICCC, line 33: fixed indentation fyi

generate_config.sh
```diff
@@ -39,6 +39,20 @@ else
...
+  case $response in
+    [nN][oO]|[nN])
+      SKIP_CLAMD=n
+      ;;
+    *)
+      SKIP_CLAMD=y
+    ;;
+  esac
+else
+ SKIP_CLAMD=y
```
jpramming: Shouldn't this have been `SKIP_CLAMD=n`?

MAGICCC: Yes you are right. I used a different approach on checking it and didn't change this one. Thanks!
Fixed in https://github.com/mailcow/mailcow-dockerized/pull/1135/commits/1741cc176ee9499c5ebbb1799b139cb81f900c1f

andryyy:
Just fixed the test, thanks. 👍

ntimo:
Why has this not been added to the master?

PR #1136, opened 2018-03-03
[Postfix][Web] Relay Inbound and Outbound mails per domain

nupplaphil:
New Pull Request for [mailcow-dev](/mailcow/mailcow-dockerized/tree/dev) instead of #1086


data/Dockerfiles/postfix/postfix.sh
```diff
@@ -74,6 +75,48 @@ query = SELECT GROUP_CONCAT(transport SEPARATOR '') AS transport_maps
   AS transport_view;
 EOF
 
+cat <<EOF > /opt/postfix/conf/sql/mysql_recipient_dependent_default_transport_maps.cf
+user = ${DBUSER}
+password = ${DBPASS}
+hosts = mysql
+dbname = ${DBNAME}
+query = SELECT IF(EXISTS(SELECT 'relay' FROM domain
```
nupplaphil, line 83: @andryyy 
The problem of sender_dependent_transport_maps.cf is, that it **always** responses "smtp:" and so transport_map interprets that it should use the MX record of the domain, so I had to change it to respond "NULL" in case there is no transport_map.

nupplaphil:
Better solution #1149 
-> Using a explicit "transport_map" table instead of re-using the "relayhost" for transport

PR #1148, opened 2018-03-08
[Web] remove encoding for resource data attribute to fix deletion

droid-sheep:
When creating a resource in the web administrating interface users can't delete it afterward.
This is related to the data-item attribute which encodes the item name.

`test@domain.com` is encoded in `test%40domain.com` and send in the ajax request to the server which can not handle it.

I tested the solution at my own server and it worked successfully!

I'm not shure if this would be the best solution but it is the easiest one.

Greetings,
droidsheep


andryyy:
Thanks, I missed that.

Issue #1068, opened 2018-02-19
Relayhost is to relay through? testing ability to send incoming emails to another server after spam-check


lavdnone:
I was testing ability to send incoming emails to another server after spam-check.
Enabling Relay, Relay all and specifying Relayhost for Domain attempts to follow simple backup MX scenario: it relays messages to first MX record.
Shouldn't it be relaying to/through Relayhost when it is specified?
https://sourceforge.net/p/postfixadmin/wiki/Relay_domains/

I just made unbound return MX record with host I want to relay to as quick workaround.

andryyy:
We use relayhosts sender dependent. I don't understand the link to Postfixadmin.

If you want to relay a domain, you'd need to mark it as backup mx. We could rename that to "relay domain" etc.

@nupplaphil is this what you try to archive? It could be enough to just create a relay_transport map in SQL. Like... returning "relay:" (the current default) for all except for those with relayhost defined. We just return the same stuff as we do for sender_dependent_default_transport then.

I wrote this a few hours ago and forgot to hit "Comment" 👍

lavdnone:
Ultimately I want to make relay for the certain domain to another host. Don't care who sends as long as they pass rspamd and domain is listed or is an alias:
Added to mailcow (host A) domain is being relayed to the host B.

It is either my understanding is different or there is a mix of terminology and outcome here:
1. When backup MX is enabled - postfix is trying to send messages received to the first MX of the domain. It is relaying Yes - but only to MX record;
2. relayhost is the host emails should be relayed to after they are received, but when backup MX is enabled it relays to/through first MX not through relayhost.

The link to postfixadmin shows that there is a different flag for the relay-relay scenario contrary to relay-to-MX scenario.

Is relayhost there to filter incoming emails to be allowed from the relayhost with additional auth?

nupplaphil:
Ah that's similar to what I try to achieve.

@andryyy it's not that simple. As lavdnone wrote, the relay has to be internal in case a Relayhost is set. But it should not get used for the outbound relaying, as the standard settings are.
So I used the transport_map in the main.cf of postfix, but just in case relaying for inbound Mails is set, I called it 'recipient_dependent_default_transport_maps' ;-)

@lavdnone maybe the enhancement at #1086 will help you too

lavdnone:
@nupplaphil thanks, yes #1086  
though I thought it is already working with relayhost and didn't want to make my mailcow differ even more
[...]

PR #1149, opened 2018-03-09
[WIP] Add inbound and outbound relaying via transport maps

SamuelNitsche:
I have also worked on a solution to implement inbound and outbound relaying via transport maps. I've implemented all the functions so far, but I haven't completely tested them yet. I have seen that [nupplaphil](https://github.com/nupplaphil) has also been working on a solution, but I think his implementation is much too complex for a mail server with a lot of email traffic. Please note that this pull request is a work in progress solution.


nupplaphil:
[Edit] Removed first post
As I see that @andryyy is working on this PR too, I think it's better to work on this instead of working on my PR. I will close my PR ;-)

SamuelNitsche:
As I already wrote, this pull request is not fully tested. Also there are few spots where I have to replace my placeholder texts. I absolutely didn‘t want to start a competition. I just wanted to show that there is another solution. :)

nupplaphil:
@SamuelNitsche Great idea to transport mails of whole domains and specific mailboxes too! I switched to your master and it seems to work for me now.

--> Next week I will try to use mailbox-specific transport_maps

[Edit] One suggestion: I have to make the nexthop like "smtp:IP:port". I think it's better to either describe how to use the nexthop field or to make different textboxes (maybe with a dropdown for smtp/smtps)

SamuelNitsche:
Good catch! I think adding different input fields is a good idea. Thanks!

[Edit] I have now implemented different input boxes, a basic description and some help texts. Also I have tested the whole system a bit more and for me it is working perfectly fine.

andryyy:
Hm, but shouldn't it be using relayhosts we added via UI?

nupplaphil:
Hm, you said at #1068 that forwarding isn't the same as relayhosts. I think we shouldn't cross-use these functions.

nupplaphil:
I'd love to use the dev/master branch again. @andryyy what do Samuel/we have to do to geht merged? :-)

SamuelNitsche:
We had another long and intensive internal discussion about the implementation of transport routes and now think that this can only be solved reasonably if we also solve issue #1068. Therefore I have updated the pull request and changed the way incoming mail routing works. The system has just passed all the internal tests. :-)
@andryyy The next two days I don't have not enough time to explain exactly what this update does. If you have any questions why and how we changed things, feel free to ask.

nupplaphil:
I'm sorry to repeat myself :-( , but are there any news so far?

data/Dockerfiles/postfix/postfix.sh
```diff
@@ -8,24 +8,6 @@ if [[ -z $(grep null /etc/aliases) ]]; then
...
-      WHERE relay_all_recipients=1
-        AND domain='%d'
-        AND backupmx=1
-  )
-  THEN '%s' ELSE (
-    SELECT goto FROM alias WHERE address='%s' AND active='1'
-  )
-  END AS result;
-EOF
-
```
SamuelNitsche, line 28: This query is no longer necessary, since it has been replaced by others.
(see main.cf -> mysql_relay_recipient_maps)

data/Dockerfiles/postfix/postfix.sh
```diff
@@ -43,6 +25,14 @@ query = SELECT IF(EXISTS(
...
 EOF
 
+cat <<EOF > /opt/postfix/conf/sql/mysql_transport_maps.cf
+user = ${DBUSER}
+password = ${DBPASS}
+hosts = mysql
+dbname = ${DBNAME}
+query = SELECT nexthop FROM transport_maps WHERE local_dest = '%s' AND active = '1';
+EOF
+
```
SamuelNitsche, line 35: Returns the transport routes for mailboxes.

data/Dockerfiles/postfix/postfix.sh
```diff
@@ -160,28 +150,50 @@ user = ${DBUSER}
...
-query = SELECT alias_domain from alias_domain WHERE alias_domain='%s' AND active='1'
-  UNION
-  SELECT domain FROM domain
-    WHERE domain='%s'
-      AND active = '1'
-      AND backupmx = '0'
+query = SELECT alias_domain from alias_domain
+        WHERE alias_domain='%s'
+          AND active='1';
 EOF
```
SamuelNitsche, line 157: This query has been simplified since the relay host implementation has changed.

data/Dockerfiles/postfix/postfix.sh
```diff
@@ -160,28 +150,50 @@ user = ${DBUSER}
...
+
+cat <<EOF > /opt/postfix/conf/sql/mysql_transport_domains_maps.cf
+user = ${DBUSER}
+password = ${DBPASS}
+hosts = mysql
+dbname = ${DBNAME}
+query = SELECT nexthop FROM domain
+        WHERE domain='%s'
+          AND active = '1';
+EOF
```
SamuelNitsche, line 186: Returns the transport routes for whole domains.

data/Dockerfiles/postfix/postfix.sh
```diff
@@ -160,28 +150,50 @@ user = ${DBUSER}
...
+cat <<EOF > /opt/postfix/conf/sql/mysql_transport_all_recipients_maps.cf
+user = ${DBUSER}
+password = ${DBPASS}
+hosts = mysql
+dbname = ${DBNAME}
+query = SELECT domain FROM domain
+        WHERE domain='%d'
+          AND nexthop <> 'lmtp:[dovecot]'
+          AND active = '1';
 EOF
```
SamuelNitsche, line 197: Returns all domains that should be forwarded to other mail servers via transport maps. Note that the 'relay_all_recipients' column has been replaced by defining a non default nexthop.

data/conf/postfix/main.cf
```diff
@@ -63,7 +64,7 @@ proxy_read_maps = proxy:mysql:/opt/postfix/conf/sql/mysql_virtual_sender_acl.cf,
   $smtpd_sender_login_maps
 queue_run_delay = 300s
 relay_domains = proxy:mysql:/opt/postfix/conf/sql/mysql_virtual_relay_domain_maps.cf
-relay_recipient_maps = proxy:mysql:/opt/postfix/conf/sql/mysql_relay_recipient_maps.cf
+relay_recipient_maps = $virtual_mailbox_maps $virtual_alias_maps proxy:mysql:/opt/postfix/conf/sql/mysql_transport_all_recipients_maps.cf proxy:mysql:/opt/postfix/conf/sql/mysql_transport_maps.cf
```
SamuelNitsche, line 67: See the comment above.

data/conf/postfix/main.cf
```diff
@@ -106,12 +107,13 @@ virtual_alias_maps = proxy:mysql:/opt/postfix/conf/sql/mysql_virtual_alias_maps.
   proxy:mysql:/opt/postfix/conf/sql/mysql_virtual_alias_domain_catchall_maps.cf
 virtual_gid_maps = static:5000
 virtual_mailbox_base = /var/vmail/
-virtual_mailbox_domains = proxy:mysql:/opt/postfix/conf/sql/mysql_virtual_domains_maps.cf
+virtual_alias_domains = proxy:mysql:/opt/postfix/conf/sql/mysql_virtual_domains_maps.cf
```
SamuelNitsche, line 110: This implementation is technically not wrong, but better solved in this way.
[Postfix virtual_mailbox_domains](http://www.postfix.org/postconf.5.html#virtual_mailbox_domains)
[Postfix virtual_alias_domains](http://www.postfix.org/postconf.5.html#virtual_alias_domains)

SamuelNitsche:
Sorry for the little spam. I added some comments to my changes to (hopefully) explain more detailed why and what changes were made.
@andryyy I'd appreciate it if you'd look at the changes and leave your opinion or wishes. Thanks!

nupplaphil:
@andryyy @SamuelNitsche

I'm sorry to ask again, but do you think you can merge it? :-( I don't do any upgrades since a month for mailcow because of this missing feature

andryyy:
This is something with very special changes that will break a lot of installations . I am also very sure some stuff is not 100% tested and will fail. Some configs values/parameters were chosen for a good reason and are now replaced to fix the special scenario this PR was made for.

I just don't have the time to check if everything still works as intended.

andryyy:
It will definitely destroy working backup mx setups, it does not migrate settings. I'm not sure about relayed users. It seems to just use transport_maps for everything now and kind of misuses relay rcpt maps. What about relayhosts? Do they still work? I didn't check all of it, this is still a very good change - don't get me wrong - but there is much more to it. :-(

nupplaphil:
Oh..

Sounds like a LOT of work. So was #1136 a more comfortable way for it with less side effects? I "just" reused the relayhost for transport_maps and didn't change anything else.

Or what about @SamuelNitsche first change before he completely changed the strategy for the transport_maps ? 

I don't want to stress, but for me it seems that this is a long run PR with a good chance that other changes will break this changes a lot more and make it more and more unlikely to get merged.

Or what do you think?

SamuelNitsche:
Since our ideas obviously differ, it probably makes more sense to close this pull request. It is probably best if you integrate this function as it fits into your existing concept.

PR #1207, opened 2018-03-27
Determine role for API-keys

Hecke29:
Currently the API-authentication via HTTP-header makes you an (super-)admin regardless of the "real" role of the corresponding user.

I changed this so that the superadmin-flag is checked. If it's set, then the role will be 'admin' else 'domainadmin'.
This also checks if the user is active as well as the api-key.


andryyy:
Thanks!

PR #1209, opened 2018-03-28
enable smtpd_tls_received_header

mritzmann:
see: http://www.postfix.org/postconf.5.html#smtpd_tls_received_header

> Request that the Postfix SMTP server produces Received: message headers that include information about the protocol and cipher used, as well as the remote SMTP client CommonName and client certificate issuer CommonName. 

bevor:

```
Received: from example.com (example.com [192.168.1.1])
```

after:

```
Received: from example.com (example.com [192.168.1.1])
	(using TLSv1.2 with cipher ECDHE-RSA-AES256-GCM-SHA384 (256/256 bits))
```

Find this a useful option. Gmail has this also enabled. I don't see any disadvantages. Could also be useful for debugging purposes.

### DIFF
--- a/data/conf/postfix/main.cf
+++ b/data/conf/postfix/main.cf
@@ -3,6 +3,7 @@ append_dot_mydomain = no
 smtpd_tls_cert_file = /etc/ssl/mail/cert.pem
 smtpd_tls_key_file = /etc/ssl/mail/key.pem
 smtpd_use_tls=yes
+smtpd_tls_received_header = yes
 smtpd_tls_session_cache_database = btree:${data_directory}/smtpd_scache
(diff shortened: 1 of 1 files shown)


andryyy:
Thanks!

Lucky88888:
great feature so hopefully soon to be merged with Master. I can confirm it works without issues. 

As add-on on this feature it would be great to see some statistics on how many incoming (and/or outgoing) e-mails are still being sent PLAIN TEXT. Is there a already present statistics command in Postfix for this?

mritzmann:
> it would be great to see some statistics

Not quite sure how that could be done. Maybe with a Sieve filter and a small counter in the debug area? But that will only work with incomming messages. But, someone else can deal with that...

PR #1229, opened 2018-04-02
Support custom server blocks

### LINKED ISSUE #1227: nginx: allow custom "server" blocks configuration
Currently the `include` for custom configuration in `site.conf` is only set inside a `server` block. It would be nice to allow including configuration (in a similar fashion as shown in the docs) outside the blocks (for, e.g. providing other containers on different domain names, or to customize the setup without breaking anything in `site.conf`).

lbeltrame:
Allows adding files in the `/data/conf/nginx/` folder, with names
`server.FOO.custom`. Each of thse should contain a full `server` block.

Fixes #1227.

### DIFF
--- a/data/conf/nginx/site.conf
+++ b/data/conf/nginx/site.conf
@@ -212,3 +212,5 @@ server {
 
   include /etc/nginx/conf.d/site.*.custom;
 }
+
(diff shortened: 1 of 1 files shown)


lbeltrame:
Unnecessary.

PR #1237, opened 2018-04-04
Optimize updating of docker-compose

Linus9000:
This pull request aims to prevent a download of docker-compose if the latest version is the same we have already installed.

While most servers connection should be fast enough to download docker-compose without any issues, I'm having some problems with the approach of first stopping the containers, then downloading docker-compose before restarting the containers. The approach itself is reasonable, considering the compose file might introduce changes requiring a recent docker-compose version, however waiting some minutes (assuming a bad connection) before restarting the containers bugs me. While it should be possible to download docker-compose before stopping the containers (assuming no change in docker-compose breaks `docker-compose down`, obviously) this propose leads to less wasted bandwith and was quite easy to implement.

Feel free to share your thoughts if you think this is a bad idea :)

### DIFF
--- a/data/conf/rspamd/lua/rspamd.local.lua
+++ b/data/conf/rspamd/lua/rspamd.local.lua
@@ -118,7 +118,7 @@ rspamd_config:register_symbol({
   type = 'postfilter',
   callback = function(task)
     local from = task:get_header('From')
-    if from and (from == 'monitoring-system@everycloudtech.us' or from == 'watchdog@localhost') then
+    if from and (string.find(from, 'monitoring-system@everycloudtech.us', 1, true) or from == 'watchdog@localhost') then
       task:set_flag('no_log')
       task:set_flag('no_stat')
     end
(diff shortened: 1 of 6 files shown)


stevesbrain:
I agree; my routes to download it often go fairly slow for the size of the file, and the fact we bring the servers down before downloading it increases the downtime. Not a big deal, but any downtime saved is a bonus in my books :)

chriscroome:
I have been using a [Ansible Playbook](https://git.coop/webarch/docker#install-or-upgrade-docker-compose) to upgrade Docker Compose, this results in it being owned by `root`. I manage Mailcow using a non-root account (that is a member of the `docker` group) and this means that the part of the script that tries to update `docker-compose` always fails due to the `mailcow` user not having write permissions, so my only suggestion would be that skipping the download if the `docker-compose` file isn't writable would be an additional check that could be added.

Linus9000:
Seems like a reasonable change, so I just added the check.

andryyy:
Thanks, but please push only to dev. Do not just change the base, that will probably break things.

Linus9000:
Sorry, didn't know your workflow here :) Changed target branch accordingly. If you prefer I can also create a new pull request.

andryyy:
No problem, but you did exactly that: you just changed the base. I cannot merge that.

PR #1238, opened 2018-04-05
Update update.sh

Linus9000:
Changes #1237 so that it's based on dev branch.


andryyy:
Thanks and sorry for the hassle... 👍

Issue #1245, opened 2018-04-06
Security concerns caused by the default password moohoo


herbetom:
I have some security concerns about the fact that a preset admin password is used. The user is not told to change it after installation with much insistence. And also if that would be the case somebody could get their before me.

Especially because after mailcow is started the first time their will be some time till LetsEncrypt has run succesfully and all is clear to go. So the user may starts Mailcow, takes a break because it is not responding directly. Comes after a few minutes back and changes the password. But until then many things could have happened. For example if somebody else (human or bot) is able to visit the page while the password is still moohoo he would be able to change things the user might never recognize but which can make some damage. This seems really unreasonable to me.

How difficult would it be to implement a random admin password which is shown in the console to the user?

Linus9000:
Implementing this took me around an hour, it will show the generated password after executing the generate-config.sh. I will gladly submit a pull request if it's wanted by the maintainers :)

herbetom:
I would really like that. You should not forget aubout `mailcow-reset-admin.sh` in the helper-scripts. The hash is their also stored as fixed hash as I have seen it. But that should be really simple easy to solve as in that file only a password must be generated randomly and then hashed and wrote to the DB like it is already the case.

I hope the maintainers are willing to change this, as it is a simple way to increase the security. Fixed default passwords are always a bad idea.

andryyy:
We had this once, I removed it due to the processing of init_db.inc.php and have not re-added it since then. Doing it in Bash is another method. But what happens when something fails? Leave it in mailcow.conf? What happens when you change the admin username?

Implementing the routine is not much work, but it needs to be fail-safe.

We could also implement a modal (or an alert) to warn about an unchanged password.

Linus9000:
Well, I hope my approach should cover the case that "something" fails in that it doesn't introduce that much more failure potential, it's almost entirely reusing code I discovered when finding out how to implement this. 

Feel free to discuss this, maybe something can be worked out :)

dannylee71:
Password is ok

herbetom:
I still think that it is not a good solution not to force the user to change the password in something diffrent then `moohoo`. If changing the password would be neccesary before accesing the config their would be at least a minimum of security that anyone messed sth. up. Although it is not a ideal solution it would be a little bit better then now.

PR #1247, opened 2018-04-07
Make the default admin password random

Linus9000:
This aims to solve #1245 by implementing a random admin password. Also, the helper script for resetting the admin will now generate a random password.

If there's potential for optimization, please let me know.

### DIFF
--- a/data/web/inc/init_db.inc.php
+++ b/data/web/inc/init_db.inc.php
@@ -1,5 +1,5 @@
 <?php
-function init_db_schema() {
+function init_db_schema($admin_default_pass) {
   try {
     global $pdo;
 
@@ -892,8 +892,9 @@ DELIMITER ;';
     $stmt = $pdo->query("SELECT NULL FROM `admin`"); 
     $num_results = count($stmt->fetchAll(PDO::FETCH_ASSOC));
     if ($num_results == 0) {
-    $stmt = $pdo->query("INSERT INTO `admin` (`username`, `password`, `superadmin`, `created`, `modified`, `active`)
-      VALUES ('admin', '{SSHA256}K8eVJ6YsZbQCfuJvSUbaQRLr0HPLz5rC9IAp0PAFl0tmNDBkMDc0NDAyOTAxN2Rk', 1, NOW(), NOW(), 1)");
+    $stmt = $pdo->prepare("INSERT INTO `admin` (`username`, `password`, `superadmin`, `created`, `modified`, `active`)
+      VALUES ('admin', :password, 1, NOW(), NOW(), 1)");
+    $stmt->execute(array(':password' => hash_password($admin_default_pass)));
     $stmt = $pdo->query("INSERT INTO `domain_admins` (`username`, `domain`, `created`, `active`)
         SELECT `username`, 'ALL', NOW(), 1 FROM `admin`
           WHERE superadmin='1' AND `username` NOT IN (SELECT `username` FROM `domain_admins`);");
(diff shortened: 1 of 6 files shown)


andryyy:
The variable shouldn’t exist in mailcow.conf when it is never reset to this value again. It just confuses users.

herbetom:
I think it is okay that it stands in `mailcow.conf` (at the moment moohoo is also a coment within that), and also it useful for an automated installation where mailcow is only a component of many and the password to login should be printed to the console after other things happenend. 

But I would change it so that in `mailcow-reset-admin.sh` the password is read from `mailcow.conf` and reset to the value in that file. This way the password is generated once per installation and the user is always able to run the script and use the default password again if he messed up. That sounds to me logically and like quite a good solution.

andryyy:
It is not a variable right now.

herbetom:
yes, but the variable is called DEFAULTPASS so it should be clear that it is not the current password. But i would change the Comment to look like this: 
```
# ------------------------------
# mailcow web ui configuration
# ------------------------------
# example.org is _not_ a valid hostname, use a fqdn here.
# Default admin user is "admin"
# Default password is the value of DEFAULTPASS. 
# INFO: You can set it back to "admin" and the value of DEFAULTPASS 
#       by running ./helper-scripts/mailcow-reset-admin.sh
MAILCOW_HOSTNAME=${MAILCOW_HOSTNAME}
DEFAULTPASS=moohoo
```

Of course how it is implemented by @Linus9000 probably wouldn't work because when updating from an older version it would come to an conflict because the mailcow.conf isn't updated by git (doesn't exist on GitHub so is their can't be easily done changes while updating). This has to be handeld. But that would be possible.

nupplaphil:
Hm.. I think it would be better to have an alert that you have to change the password rather than to use a variable with a random password in the mailcow.conf

And to you could add a "Please change the default admin password after you successfully logged in the first time" in https://mailcow.github.io/mailcow-dockerized-docs/install/

MAGICCC:
@nupplaphil Well the problem with that is still the same imho. Some random user detects a fresh mailcow installation and login with the current default password and changes it after seeing the popup.

To be honest. I wouldn't change anything. When an user can't wait some minutes until everything is started, then he may not use his own mailserver (regarding no time to check its status and update and such)
Don't get me wrong about password security and it's 'bad' to use everywhere the same password

Linus9000:
This PR aims specifically to prevent the scenario outlined in the second paragraph of the original issue #1245, which could not be mitigated by an alert of some sort. 

To be honest, on my first installation of mailcow after logging in I thought the password change is mandatory (I mean, it's red and shows right after login). Also I hope everyone using software like this on a public server knows to change default passwords.

The problem which lead me to writing the password into mailcow.conf is that I wanted to output it to the user, while making it available to the PHP scripts with only small changes. 
As the generate_config.sh is the only place where a user gets an output in the installing phase it made sense for me to generate it there. The logical consequence was then to write it into mailcow.conf, because all the logic to get values from there into the containers is already present.

Of course this approach may have undesirable consequences, confusing users is certainly one of them. I could change the mailcow-reset-admin.sh to use this password, thereby eliminating the confusion (if I understood your comment correctly).

andryyy:
Hm, we would need to update the docs, too.

Linus9000:
Just noticed I put the DEFAULTPASS-block just in between database config and database password, whoops..

Knight1:
Is there any intention to merge this?

PR #1251, opened 2018-04-10
BatchSyncJob.sh

reinistihovs:
I wrote a script. which is very handy to migrate multiple mailboxes from one server to another.
Script requires Dovecot master user and configured Domains and mailboxes on target server. 
Script is intended to be run on target Mailcow server.


dannylee71:
Dannylee71_check pull

dannylee71 (approved):

PR #1265, opened 2018-04-13
Fix initial compose project name

paescuj:
According to commit https://github.com/mailcow/mailcow-dockerized/commit/152ccba1d4a5d107019ca73ec11cc6155d8238cf, the initial compose project name should also be fixed


andryyy:
That’s indeed missing in master, thanks.

Issue #1286, opened 2018-04-18
NextCloud: Uploading large files via Syncer app doesn't work.


hachre:
Hi,

not sure if NextCloud is supported but in any case, I've noticed the following problem:

1. I installed NextCloud by running `bash helper-scripts/nextcloud.sh --install` and chose the `subfolder` method.
2. Then I logged in to the NextCloud interface with a normal mailbox login and everything seemed to work fine.
3. I then installed the official NextCloud sync client for Windows and attempted to upload a 5 GB file to the server.

Result:
1. The file upload randomly seems to cancel past around 3GB but not at the exact same value each time. It may be time-based instead of size-based. 
2. The sync client then waits for a short time and automatically retries uploading the file.
3. This goes on forever.

Observations:

I've logged in to the NextCloud interface with the admin user and looked at the file upload size setting there. It is set to 511MB and says it cannot be changed. However the docs for the sync client tell me that this setting doesn't apply to the sync client, since it will use small chunks to get around any limits anyway. This further makes me believe that there must be some time-based timeout that interrupts the transfer. 

Update: I've also found the following messages in the host machine's `dmesg` (there's hundreds of these)
```
[Wed Apr 18 12:06:44 2018] traps: php-fpm[17768] general protection ip:7f1483077927 sp:7ffdeab3f840 error:0 in ld-musl-x86_64.so.1[7f1483054000+88000]
[Wed Apr 18 12:06:46 2018] traps: php-fpm[17780] general protection ip:7f1483077927 sp:7ffdeab3f840 error:0 in ld-musl-x86_64.so.1[7f1483054000+88000]
[Wed Apr 18 12:06:48 2018] traps: php-fpm[17762] general protection ip:7f1483077927 sp:7ffdeab3f840 error:0 in ld-musl-x86_64.so.1[7f1483054000+88000]
[Wed Apr 18 12:06:56 2018] traps: php-fpm[17783] general protection ip:7f1483077927 sp:7ffdeab3f840 error:0 in ld-musl-x86_64.so.1[7f1483054000+88000]
```
The host is running Ubuntu 16.04 and is not running anything besides Mailcow. Docker is version 18.04.0-ce, build 3d479c0. My Mailcow installation is up-to-date as of 15 hours ago.

Any ideas?

hachre:
Another update:

[...]

PR #1288, opened 2018-04-19
partially fixes bug #1286

hachre:
this fixes the 120 second timeout problem mentioned in bug #1286


andryyy:
Hi,
Thanks, but please don't push to master. It is missing the fix for non-subfolder installations. I fixed it in dev and also removed the max body size there.

hachre:
Sorry and thanks :)

andryyy:
No problem... 👍 I also added the redis timeout value (not yet in dev).

PR #1294, opened 2018-04-19
Cleanup

flobeier:
Duplicate entries removed and remaining sorted alphabetically, indentation through tabs changed to spaces


hachre:
FYI: This will most likely get declined because it's against `master`. It should be against `dev`!

flobeier:
@hachre thanks for the heads-up, I'll do a rebase.

andryyy:
That will not work. This does merge master to dev, which is right now problematic. Can you please clone dev and push from there?

andryyy:
And thanks for the cleanup. :-)

PR #1302, opened 2018-04-21
Some quick UI edits

MAGICCC:
Basicly it moves the Rspamd settings into admin.php instead of debug.php

### DIFF
--- a/data/web/admin.php
+++ b/data/web/admin.php
@@ -1,8 +1,8 @@
 <?php
-require_once("inc/prerequisites.inc.php");
+require_once "inc/prerequisites.inc.php";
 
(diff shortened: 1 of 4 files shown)


andryyy:
The settings map is included for debugging. Moving menus or other UI changes will be closed without discussing them. :-(

MAGICCC:
So a PR can never ever include UI changes? That's a bummer because e.g. changing the admin password of Rspamd isn't a debug feature but an administration one...
Welp nothing can be done about improving little things then :/

MAGICCC:
Actually we can split the 'changing admin password' part and the 'settings map' part.
More in https://github.com/mailcow/mailcow-dockerized/issues/1304
Already commited btw https://github.com/MAGICCC/mailcow-dockerized/commit/e39b7c3218f989bf5121312432eeba9549f9ccd6

PR #1313, opened 2018-04-24
Cleanup

flobeier:
Duplicate entries removed and remaining sorted alphabetically, indentation through tabs changed to spaces. Duplicate of #1294 which was accidently closed by me while trying to fix it.

### DIFF
--- a/data/Dockerfiles/phpfpm/Dockerfile
+++ b/data/Dockerfiles/phpfpm/Dockerfile
@@ -1,84 +1,81 @@
 FROM php:7.2-fpm-alpine3.7
 LABEL maintainer "Andre Peters <andre.peters@servercow.de>"
 
-ENV REDIS_PECL 3.1.6
-ENV MEMCACHED_PECL 3.0.4
 ENV APCU_PECL 5.1.10
 ENV IMAGICK_PECL 3.4.3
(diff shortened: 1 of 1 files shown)


andryyy:
But this still has duplicates and is not sorted. 😱 :-)

andryyy:
Some pecl extensions should also be installed in a given order. I remember having trouble with this - not in mailcow though.

Still, it has duplicates and is not really sorted. 🤷‍♂️ 

Give me a few minutes to pipe it through sort and uniq.

flobeier:
@andryyy okay, I didn't know that there could be trouble with the pecl extensions if you change the order.

I'm sorry that you had to go through the file and fix it, which was my original intention to do. On top of being relatively inexperienced with git I was a little tired last night and must have commited the wrong file. One duplicate still slipped through (`libjpeg-turbo-dev`).
Thanks for the learning opportunity.

andryyy:
The image doesn't even build like this. Please never, ever push to this repository when you didn't test exactly what you push.

flobeier:
@andryyy yes, again, I'm sorry. I honestly don't know why I pushed the wrong content. I shouldn't have tried to quickly fix my earlier mistake before going to bed.

PR #1329, opened 2018-04-27
[Dovecot] Fix to allow the use of auth_default_realm without breaking…

### LINKED ISSUE #1325: dovecot default realm
Hi,

I'm currently configuring a mailcow box to replace an old sendmail install and I have a little question.

On our current sendmail setup our users can authenticate with username without the domain part, it defaults to our main domain.

So a user authenticating without domain in username, for example `mybuddy` will automaticall be transformed to `mybuddy@maindomain.com`.

This is achieved easily in dovecot.conf by adding 
```
auth_default_realm = maindomain.com
```
in `dovecot.conf` (well I added it to `extra.conf`)

It works like a charm, but it breaks imapsync jobs and probably something with sogo and sieve too.

The imapsync jobs seems to use a master dovecot user/pass (using the `*` separator on the username) and that is grabbed from a file `sieve.creds` in `/etc/sogo`.

Problem is that with `auth_default_realm` activated, dovecot will try to auth as `dovecot_master_user@maindomain.com` and this is not what is found in `dovecot-master.passwd`.
Authentification will then fail.

I thought about manually adding `dovecot_master_user@maindomain.com`  in that file but it guess will be overwritten at every dovecot container restart.

Any idea how I could cleanly add this feature so it's compliant with mailcow guidelines ?

I hope you'll understand the issue despite my bad english :)

sriccio:
… the master password used by imapsync and others

Fixes #1325

### DIFF
--- a/data/Dockerfiles/dovecot/docker-entrypoint.sh
+++ b/data/Dockerfiles/dovecot/docker-entrypoint.sh
@@ -97,8 +97,8 @@ if [[ $(stat -c %U /var/vmail/) != "vmail" ]] ; then chown -R vmail:vmail /var/v
 # Create random master for SOGo sieve features
 RAND_USER=$(cat /dev/urandom | tr -dc 'a-z0-9' | fold -w 16 | head -n 1)
 RAND_PASS=$(cat /dev/urandom | tr -dc 'a-z0-9' | fold -w 24 | head -n 1)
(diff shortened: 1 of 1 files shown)


hachre:
Please make a new pull request against the `dev` branch!

sriccio:
oops sorry, i'm not used to this. I'll PR again to dev

PR #1364, opened 2018-05-06
Fix issue where AUTO_INCREMENT != PK in mailbox table

### LINKED ISSUE #1363: Database initialisation failed
Brand new install after following the user guide, I get this when I hit the UI for the first (and subsequent times):

`Database initialisation failed: SQLSTATE[42000]: Syntax error or access violation: 1075 Incorrect table definition; there can be only one auto column and it must be defined as a key`

danielfaulknor:
Currently new builds fail because we're trying to use username as a primary key and have id auto incrementing.
I've presumed there should be an index on username considering it's the mailbox table.

Fixes #1363

### DIFF
--- a/data/web/inc/init_db.inc.php
+++ b/data/web/inc/init_db.inc.php
@@ -198,10 +198,11 @@ function init_db_schema() {
         ),
         "keys" => array(
           "primary" => array(
-            "" => array("username")
+            "" => array("id")
           ),
           "key" => array(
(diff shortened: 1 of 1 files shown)


andryyy:
This will not work on table updates because we cannot add an auto increment type to a non-primary key. This works on first init, yes.

Edit: I am working on fixing it right now.

PR #1380, opened 2018-05-12
WIP: Auto Delete Emails in Junk/Trash After 30 Days

PhoenixPeca:
Basically, this implementation allows automatic deletion of all emails in the Junk/Trash folder that has an age of 30 days or more. The deletion of emails is handled by a ~bash script~ PHP-CLI script in the "cron.daily" folder in the dovecot container. This is made possible with the [dovecot expire plugin](https://wiki2.dovecot.org/Plugins/Expire). I haven't implemented a feature that adds a settings to make it optional in the mailcow UI just yet.

~Additionally, in Gmail and Outlook, there's no settings (or i am not aware of a settings) that would allow their mailbox users to turn off the auto delete feature in their trash/junk folder. So i thought mailcow should also do the same.~ lemme know what you think? 😅😅😅😅

## UPDATE
I am working on making this a domain-wide policy so it becomes an optional feature. It is DISABLED BY DEFAULT of course.

I'll switch to the dev branch later. Just need it for testing in master branch for now.


ntimo:
Hello,
I think this feature would be really really great. But maybe a option within the mailcow admin panel would be nice, so that users can individually turn it off or on.

PhoenixPeca:
yup, that's what i'm also thinking. i'm working on it right now.

MAGICCC:
It's nice yes, but as @ntimo said not all (like me) deletes mails from such folders.
(and well merging into master isnt always the best idea tho - but it doesn't matter with this repo anyway)

PhoenixPeca:
Yep, @MAGICCC I am working on adding it as an optional domain-wide policy for domain-admins.

ntimo:
I actually would love it would not be domain based but mailbox based. Oh an @MAGICCC I think having this feature in the master branch would be nice. Because this would be really really helpful for those people how would like to use it (especially because you can then just run the update.sh in order to get this feature). I personally would use it for my own mailbox but not for the one of other users, because the decision to use this feature should be made by each Mailbox owner.

Oh and I think you might not want to change the repo url in the update.sh ;) This would be kind of bad if this branch would be merged with the master.

MAGICCC:
@ntimo I had several discussions about maintaining a git repo like this e.g. tagging/don't push to master (theres a dev branch for reasons) and such. But nobody listens...

andryyy:
You have also pushed some breaking changes to master. Plus we already discussed this matter.

Don't know why "not doing something" equals to "nobody listens".

Whatever: it still belongs to dev.

PhoenixPeca:
@andryyy, sorry. i really have no plan on merging this into master. Its just that the dev branch changes quite very fast that I find it hard to patch up with the changes and possible conflicts.. So, i chose master for now to get rid of possible issues while debugging.

I made this PR for other people to see my changes and to get some recommendations. Anyways, if this is done, i'll close this and open up a new PR that will be merged into dev branch. 

> You have also pushed some breaking changes to master.

No, I did not. I would never commit directly into master that would break the existing dev commits.

hachre:
> No, I did not. I would never commit directly into master that would break the existing dev commits.

@PhoenixPeca, I think @andryyy was referring to @MAGICCC in that statement.

PhoenixPeca:
Hi, Ohhh.. shit.. 😭 last night was really a mess, been figuring things out until 2AM. I did not see those comments ahead.. hahahah!! Sorry again.

PhoenixPeca:
As for @ntimo's comment...

> I actually would love it would not be domain based but mailbox based.

maybe i'll get that into consideration if the demand increases, but for now, I think it is best to have it this way.

> Oh and I think you might not want to change the repo url in the update.sh ;) 

Yep, I know that.. I was just testing things to see if everything goes right in update script if these feature were to be pushed into final release. Apparently, after the testing was done, I reverted that commit. Besides, that wouldn't be much of a problem really because running the update script would just replace those changes back to its original form.

PhoenixPeca:
## Update
After some tests, I can now confirm that the implementation of this proposal feature is successful:
image
image

PhoenixPeca:
I'll be opening a new PR for dev later. 😉

PR #1383, opened 2018-05-13
Dev: Auto Delete Emails in Junk/Trash After 30 Days

PhoenixPeca:
### This is a followup PR for #1380 

Basically, this implementation allows automatic deletion/expunging in the Junk/Trash folder of a mailbox that has an age of d>30 days . The deletion or expunging happens daily and is handled by a PHP-CLI script (that reads timestamps of the oldest email's from the database) in the "cron.daily" folder within the dovecot container. The mailbox-folder tracking is handled by the dovecot expire plugin which saves the timestamp of the oldest emails to the database.

I have also added it to web UI so that domain admins can enforce the policy if needed as seen in the screenshot below:
image

Note: This was implemented as a _domain-wide-policy_ which is disabled by default.

## [Update] Added debug log ui for auto expunge
image
image


PhoenixPeca:
i believe this is now ready for merge. 😄

andryyy:
Please revert this. There is too much we need to check and change before merging it.

PR #1409, opened 2018-05-22
Extract container's last IP address by default

mstate:
This takes care of the case where multiple IP addresses exist for a container and ensures that get_container_ip only returns one (the last one).

The issue with the existing script occurs when multiple IP addresses exist for a service.  For example, if nginx-mailcow is also on a network with a reverse proxy server.  In this case, the get_container_ip function would return a string with a space separated list of IP addresses.  Check functions calling get_container_ip would be unable to reach their respective services with this non-ip address string and would show these services as failed.


andryyy:
Thanks for contributing. 👍 

Wouldn't it be better to check if the containers IP is in the mailcow network? The last doesn't sound correct in all cases.

Like `if grep -q ${ipv4_net} <(echo ${container_ip}); then do something; fi`

Ah, please also push only to dev. Do not switch branches via UI, just checkout dev and push it there, please. :-)

PR #1425, opened 2018-05-25
[rspamd] fix redis multimaps in version 1.7.5

mkuron:
Rspamd 1.7.4 and higher (https://github.com/vstakhov/rspamd/commit/92cfb7f3b92813736ff949028366f989b5e5f8b1) changed the behavior of redis multimaps. Whereas previously it would only check whether a key is present and insert the defined symbol if it was, it now inserts the *value of the key* as a symbol.

This change broke the forwarding hosts functionality (instead of inserting a `WHITELISTED_FWD_HOST`, a useless symbol like `gmx` or `gmail` might be inserted) and may have also affected the tagged recipient functionality.

This pull request restores the old behavior by forcing the multimap module to only insert the symbol corresponding to the rule name.


andryyy:
Nice, thank you!

PR #1427, opened 2018-05-26
Update watchdog.sh

daschxx:
remove get_container_ip and set hostnames to be compatible with zwo networks (for example to use an external docker mailman or an docker nginx proxy)


andryyy:
You cannot just remove that thinking it has no meaning and push to master. O_O

PR #1434, opened 2018-05-28
hardening http headers

apoc4lyps:
add X-Frame-Options and Referrer-Policy to nginx config

### DIFF
--- a/data/conf/nginx/site.conf
+++ b/data/conf/nginx/site.conf
@@ -37,7 +37,7 @@ server {
   add_header X-Download-Options noopen;
   add_header X-Frame-Options "SAMEORIGIN" always;
   add_header X-Permitted-Cross-Domain-Policies none;
-  add_header Referrer-Policy no-referrer-when-downgrade;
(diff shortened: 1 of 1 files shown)


mritzmann:
I like the idea of improving the security headers.

Regarding `Referrer-Policy`:

In my opinion, there is no reason why the referer should be sent to external websites at all. Doesn't it make sense to put it directly on `same-origin`?

Knight1:
Indeed. If the SOGo Domain does not match the Mail domain it is hidden with same-origin. I personally would use strict-origin to prevent downgrade attacks.

mritzmann:
> I personally would use strict-origin to prevent downgrade attacks.

`strict-origin` also sends the reffer to external sites.

table

(Screenshot from [scotthelme.co.uk](https://scotthelme.co.uk/a-new-security-header-referrer-policy/))

----

Maybe `no-referrer` makes sense? Is the reffer being used for something useful?

Knight1:
Any intention to merge this?

PR #1483, opened 2018-06-16
add basic ldap support

etique57:
Not exactly sure of what I'm doing...


mkuron:
Discussion was started in https://github.com/mailcow/mailcow-dockerized/pull/743#issuecomment-388651331 a month ago. Maybe @bkleiner / @str8labs, the original author, can explain what they did in this code, what works and what doesn't. Then we can try to finish up their LDAP work.

neox8:
has anyone tried this build yet?

it-toppp:
LDAP worked?

pde-rent:
Why was this PR closed ? Anything wrong with the implementation ?

mkuron:
The original developers never followed up and while it looks like a good start, it was definitely incomplete. Mailcow has changed quite a bit over time so that this solution will likely not work anymore.

PR #1495, opened 2018-06-23
Adding Auth-User for imapsync

nupplaphil:
I'm currently migrating user mailboxes from exchange 2016 to mailcow.

Because I don't want to create separated *.pst files, I'm using the admin account as "authuser" and now I can sync every mailbox with mailcow :-).

This PR should enable it for everybody


andryyy:
Thanks. Can you add an explanation below the auth user field? Use `<p class="help-block">` for example. :-)

andryyy:
The description is not useful. People will not understand what this is for. This _will_ lead to questions like "When do I need to use this field? Should I leave it blank? Should I enter just anything?"

nupplaphil:
OK, I'll retry it ;-)

nupplaphil:
How about that?
If it's still not clear enough, please make a suggestion if it's possible

andryyy:
What about the password? Does it need to be the password for the user or the "auth user"? What is an "auth user account" anyway? Is it a master user? What is "username" used for, when we specify the auth user?

You see, this is frustrating for users. You know what you need it for, but not everyone will understand this.

What will be read when using an "auth user" to determine the correct mails to fetch? Will it read the "To:" header?

nupplaphil:
Hm .. You're right with the password! Because it has to be the password of the Login name - if set. Should I make a help-text to the password or describe it at the authuser anyway?

But it doesn't change anything in the behavior of imapsync. It's "just" to login if you don't own the user account. I don't think that I should explain what it doesn't change (determine correct mails to fetch, read "To:" , ...)

What about naming it "Auth User" with the description:
"The Auth(entication) User can optionally be used instead of the Username to login at the Usernames mailbox (if the Auth User has the right for it - f.e. administrator). If set, the Auth Users password has to be set at "Password".

nupplaphil:
One more try ;-) ..
[edit] And I thought that's a quick fix :-D

btw it's not that easy in ANY case. IMAP is different in exchange 2003/2007/2010/2013 and 2016. It took me hours to find the right imapsync-configuration for Exchange 2016 (even with the normal imapsync ;-) ).

nupplaphil:
For example I found out (with the help of https://github.com/imapsync/imapsync/issues/136#issuecomment-397020248 ) that Exchange is blocking the administrator account for IMAP at all (no login, no sync, ...) ;-)

nupplaphil:
From my part, I would be ready for merge :-) .. Just say if I have to change something

andryyy:
We added a custom parameter field, which should be used for that. 👍  Only a very few fields are hard-coded now. I hope you understand.

Issue #1491, opened 2018-06-20
ECC Certificates


carwynR:
Hello,

mailcow does not support ECC Certificates.

Postfix: `warning: TLS library problem: error:1417A0C1:SSL routines:tls_post_process_client_hello:no shared cipher:../ssl/statem/statem_srvr.c:1404:`

lismueller:
I don´t know how to fix this

andryyy:
Can you test again after update?

lismueller:
@andryyy I am using the latest mailcow, the issue persists

PR #1510, opened 2018-06-27
[Postfix] Default SMTP server security grade for EECDH key exchange

elcore:
Currently mailcow can´t negotiate a successful TLS Handshake with a P-384 certificate (sec384r1)

This PR modifies the Postfix configuration to "use the most preferred curve that is supported by both the client and the server", thus enabling a successful TLS Handshake.

http://www.postfix.org/postconf.5.html#smtpd_tls_eecdh_grade
http://www.postfix.org/FORWARD_SECRECY_README.html


elcore:
This might be related to #1491, I experienced the same issue.

andryyy:
And untested.

elcore:
> And untested.

Well, it fixed the issue for me.

andryyy:
How? It is available in Postfix 3.2. Check your logs. It should report that error.

elcore:
> How? It is available in Postfix 3.2. Check your logs. It should report that error.

@andryyy I have checked the logs and yes, it reports this error, yet my mailcow instance is working now.

elcore:
This config works too `smtpd_tls_eecdh_grade = ultra` (no errors in the logs)

elcore:
I have a proposal, we should update the TLS documentation until mailcow is using Postfix ≥ 3.2.

"You must replace `smtpd_tls_eecdh_grade = strong` with `smtpd_tls_eecdh_grade = ultra` in `data/conf/postfix/main.cf` if you are using a P-384 certificate."

elcore:
> I have a proposal, we should update the TLS documentation until mailcow is using Postfix ≥ 3.2.
>
>"You must replace smtpd_tls_eecdh_grade = strong with smtpd_tls_eecdh_grade = ultra in data/conf/postfix/main.cf if you are using a P-384 certificate."

@andryyy What do you think?

andryyy:
I will push a new image very soon. :-)

PR #1517, opened 2018-06-28
Redact user IP from outgoing mail header

MaxXor:
For privacy reasons the IP of the sender should be stripped from outgoing SMTP headers and replaced by localhost. The other ignored headers are to protect the user from leaking additional information from his mail client.


andryyy:
It is in the docs.

I will not add this by default, since we would be actively changing mail content.

This PR would also replace ALL "Received" headers, that's a very bad idea. :-( 

Thank you for your contribution anyway! 👍

MaxXor:
Ok I understand your point with changing the mail content, but at least the user IP should be stripped. This is a privacy leak where I doubt many users are even aware of it. Imho this should be enabled by default.

PR #1518, opened 2018-06-28
Redact user IP from outgoing mail header

MaxXor:
See https://github.com/mailcow/mailcow-dockerized/pull/1517#issuecomment-400939329

### DIFF
--- a/data/conf/postfix/header_anonymize_ip.pcre
+++ b/data/conf/postfix/header_anonymize_ip.pcre
@@ -1,0 +1,2 @@
+/^\s*Received:.*Authenticated sender:(.+)/
+    REPLACE Received: from localhost (localhost [127.0.0.1]) (Authenticated sender:$1
--- a/data/conf/postfix/main.cf
+++ b/data/conf/postfix/main.cf
@@ -127,3 +127,4 @@ smtp_sasl_auth_enable = yes
 smtp_sasl_password_maps = proxy:mysql:/opt/postfix/conf/sql/mysql_sasl_passwd_maps.cf
 smtp_sasl_security_options = 
 smtp_sasl_mechanism_filter = plain, login
+smtp_header_checks = pcre:/opt/postfix/conf/header_anonymize_ip.pcre


andryyy:
See my comment, please. I don't want this to be enabled by default. You browse websites every day exposing your IP address. This is not a leak in my opinion. I don't want to modify messages by default. After all you send from your personal email address, people will know who you are.

This also strips all "authenticated received" headers. Some people want to keep these headers from external sources. We could change the header check to only replace mailcows header, but even then I am not sure if I want to merge it, tbh.

Please also keep in mind, that we put it in the docs. It is very easy to implement for those who need/want it. :-)

MaxXor:
The people I exchange mails with are not the problem, it's the mail providers they are using. I'm sure most of the people don't even know how to check the mail headers. Mail providers like Google read the sender IP from the header and link it to other data that is generated by a user when surfing the web, for example.

We shouldn't make it too easy for Google to track every step of our life... :crossed_swords:

moba:
This is even an issue considering data protection law like GDPR. You are sharing personal identifiers with third parties, whereas there is no technical reason to do so. Just because "it's always been like that" is no argument. Gmail does not include Client IPs in the headers, and I've been running mailservers for more than 20 years that don't, with zero problems. It is in fact the other way round: In the few cases where people need the information, they should turn it ON (at their own risk).

mkuron:
We definitely need to keep the header on emails coming from outside (port 25), otherwise DKIM signatures will be broken. @maxxor, can you update your pull request and the documentation to limit the header replacement to port 587 and 465? In the current state, it's not even worth discussing, as @andryyy said.

MaxXor:
@mkuron I don't agree. The `Received` header is not included in the signing process to generate a DKIM signature thus replacing the authenticated sender IP address will not break the signature. It's usually these headers at most: `h=from:from:sender:reply-to:subject:date:message-id:to:cc:mime-version`
Some mail providers like yahoo use even less (`Date:From:Reply-To:Subject:References:From:Subject`)

mkuron:
@MaxXor, that's not true. A mailserver can decide to sign whichever headers it likes, and some do include the `Received` header. Just run `grep -r ':received[^-]\|[^-]received:' /var/lib/docker/volumes/mailcowdockerized_vmail-vol-1` and see. I have thousands of matches, and all of these would have had their DKIM signatures broken by this pull request. So either change it or close it.

MaxXor:
@mkuron It can be included yes, sorry for being unclear. I checked my mails with your provided regex and have 0 matches (out of 1500 mails). That's pretty weird.

andryyy:
Because your mails are compressed...

stevesbrain:
I'm confused, at least regarding originating IP on outgoing mails - I've sent test mail to Gmail, and to another Mailcow box on the same server, and my client IP isn't revealed on either. Is this abnormal? I'd always assumed Mailcow had just been stripping it out anyway.

EnriqCG:
> and my client IP isn't revealed on either

Yes indeed, the IP Address is sent and can be seen on the original message (gmail for instance). This message in particular was sent through Thunderbird.

`Received: from [192.168.1.3] (XX.XX.XX.XX.dynamic.ispreverse.tld [XX.XX.XX.XX]) (using TLSv1.2 with cipher ECDHE-RSA-AES128-GCM-SHA256 (128/128 bits)) (No client certificate requested) `

Note that not only the external IP is leaked, but the internal as well. This happens in most modern browsers too due to a bug in WebRTC.

andryyy:
External IP is leaked? The MTA will _always_ reveal its IP when it connects over TCP. The receiving (!) MTA will add the "Received From" header according to the delivering MTA, we cannot change that.

Replacing _all_ received headers is just wrong. It is non-sense. 

You can anonymize a users (!) IP by replacing the content after mailcow - as the receiving MTA - added the header to the message. That's what is described in the docs. mailcow will always see the users IP, that's how TCP works. You need to connect from somewhere. Postfix will also log what it replaced.

A proper regex to replace authenticated users from mailcow domains is fine. Replacing the IPs of all previous hops is not.

stevesbrain:
Yeah, sorry - let me clarify. When I sent, my actual client *public* address isn't revealed - only my Mailcow MTA. I was just expecting that my client IP was also going to be revealed - perhaps this is because I'm using EAS? Either way, I also disagree with stripping out headers in Mailcow by default (false sense of security at best)

mkuron:
> by replacing the content after mailcow - as the receiving MTA - added the header to the message

Of course, one needs to be careful to not replace a similar header that may have been by the originating mail server so that the DKIM signature is not broken. The currently documented method would in this case break DKIM signatures. One could easily limit the `smtp_header_checks` to ports 465 and 587 though.

> perhaps this is because I'm using EAS?

Correct. The first "Received" header then refers to the internal address of the SOGo container.

MaxXor:
@andryyy @mkuron  lol https://github.com/mailcow/mailcow-dockerized/commit/a844adde0f416357eefee125bbdf59185b17a3a6

andryyy:
I cannot stand the crying anymore. ;-) We still need to watch it and hope it does not break ARC.

andryyy:
PS: I hope you remember you replaced every Received header and argued it was fine. So much about "lol".

MaxXor:
It still worked as none of the mail servers (which I've checked) included these headers in their DKIM signature. soooo :stuck_out_tongue:

andryyy:
You are seriously still fine with replacing any Received header? Wow.

PR #1538, opened 2018-07-04
Fix infinite loop when parsing update.sh arguments

### LINKED ISSUE #1193: update.sh loops at commandline parsing
When running `update.sh --ours` on Ubuntu 16.04, the following code just loops, and the script does not continue. I figured this out when adding a `set -x`:

```bash
while (($#)); do
  case "${1}" in
  #...
   --ours)
      MERGE_STRATEGY=ours
;;
```

Output snippet:

```
+ (( 1 ))
+ case "${1}" in
+ MERGE_STRATEGY=ours
+ (( 1 ))
+ case "${1}" in
+ MERGE_STRATEGY=ours
+ (( 1 ))
+ case "${1}" in
+ MERGE_STRATEGY=ours
+ (( 1 ))
+ case "${1}" in
```

tsujamin:
update.sh would enter an infinite loop when arguments (e.g. --ours, --check). This is due to a missing `shift` statement that has now been added. This fixes #1193.

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -28,6 +28,7 @@ while (($#)); do
       MERGE_STRATEGY=ours
     ;;
   esac
+  shift
 done
 
 [[ ! -f mailcow.conf ]] && { echo "mailcow.conf is missing"; exit 1;}


andryyy:
Thanks!

PR #1562, opened 2018-07-13
Updates generate_config.sh making it working on macOS

### LINKED ISSUE #1556: MacOS: ./generate_config.sh: line 27: read: -i: invalid option
I'm trying to run the `generate_config.sh` on my Mac but it is failing as the following :

```
./generate_config.sh: line 27: read: -i: invalid option
read: usage: read [-ers] [-u fd] [-t timeout] [-p prompt] [-a array] [-n nchars] [-d delim] [name ...]
./generate_config.sh: line 39: read: -i: invalid option
read: usage: read [-ers] [-u fd] [-t timeout] [-p prompt] [-a array] [-n nchars] [-d delim] [name ...]
tr: Illegal byte sequence
tr: Illegal byte sequence
```

zedtux:
This PR is solving issue #1556 by updating the `generate_config.sh` file which wasn't working on macOS.

 * Changes the usage of the `read` command in order to not use the `-i` flag and check the answered value by the user after the `read` command.
 * Adds `LC_ALL=C` in front of the `tr` command to generate `DBPASS` and `DBROOT` variable values
 * Also update variable naming for `TZ` to `MAILCOW_TZ` avoiding collisions

Tested with the script on macOS 10.14 with empty and non empty answer for FQDN and timezone, both working fine.

Closes #1556

### DIFF
--- a/generate_config.sh
+++ b/generate_config.sh
@@ -11,7 +11,7 @@ if cp --help 2>&1 | grep -q -i "busybox"; then
   exit 1
 fi
 
-if [[ -f mailcow.conf ]]; then
+if [ -f mailcow.conf ]; then
   read -r -p "A config file exists and will be overwritten, are you sure you want to contine? [y/N] " response
   case $response in
     [yY][eE][sS]|[yY])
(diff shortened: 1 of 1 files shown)


zedtux:
Thank you @mkuron for changing the title of this PR (even if I don't really understand why you did this), but what about giving your input or even better, merging it ? 😄

andryyy:
This also removes the default example "mx.example.org" which I think is more useful than having generate_config.sh work on macOS. You should at least mention this in your PR. :-)

I will run a test and merge it.

zedtux:
Sorry @andryyy but you're wrong. I changed it to make it working on the Mac, but `mx.example.org` is still the default value.

andryyy:
It is not shown in any way. Before this PR it was set as visible default value. You know, what I mean? Same with TZ. I will change it a bit, so you see the default value and know you can confirm it by pressing enter.

grafik

I still don't know why we need this to run on a Mac. I hope no one will run a mail server on a Mac.

zedtux:
Are you, maybe, talking about [this commit](https://github.com/mailcow/mailcow-dockerized/commit/1e59816665b1de41db5be0d51d5bd8cb3d5772fe) which changed this?

Or do you mean `read` is showing the default value given to the `-i` flag in the terminal?
If it's the case, then I didn't know about it, and it should be changed, I agree.

> I still don't know why we need this to run on a Mac. I hope no one will run a mail server on a Mac.

Please ... read @mkuron's comment from #1556, you'll better understand the point of this PR. 😞

andryyy:
I mean the -i flag indeed. But that's not a big deal.

Changing the TZ var name is a good idea.

I know the point of this PR. But I still don't really see the need to run this on a Mac. Why would you not just run this script on the server? I will not respond to issues complaining about rejected mails sent from a macOS cow. :-P

When you copy the files to another location, remember not to break the symlink of mailcow.conf to .env.

zedtux:
I'm using `docker-machine` from my Mac. I would like to keep mailcow on my machine and push only the necessary to my server.
I know you couldn't like the way I want it, but I don't want to have to SSH my servers, remembering what server is doing what all the time (of course I know which server does what), and remembering where and how to perform this or that.

I have private Git repo with all my `docker-compose.yml` files (one for the front nginx proxy + letsencrypt companion, another one for my blog, and so on).

The idea behind is that in case my server crashes, I just build a new one, run a single command, and all is back in a minute (already used this with a projet, where the prod was down, after the server provider confirmed he can't do more to help me -- disk crashed --, the minute after the new server was up, my prod was back, thanks to Docker).

Anyway, having the scripts working on any platform isn't hurting that much, does it?

andryyy:
That's a valid use-case, yes. I mean, I don't care, you could run it on a Mac, who am I to forbid it? But I will not support it in any way.

I think it was a valid concern, that the -i flag was removed and we lost the "visible default", wasn't it?

andryyy:
Thanks for contributing. 👍  :-)

zedtux:
Thank you very much @andryyy and I fully understand your point.

Issue #1561, opened 2018-07-12
SoGo not working after Update


FloThinksPi:
After a Update SoGo cant connect to the DB anymore

I am on commit 1522b832477685766963bb7237fc883f71ecc0ec now

SOGO Logs
```
2018-07-12 21:32:34,081 CRIT Set uid to user 0
2018-07-12 21:32:34,089 INFO supervisord started with pid 1
2018-07-12 21:32:35,099 INFO spawned: 'syslog-ng' with pid 8
2018-07-12 21:32:35,102 INFO spawned: 'cron' with pid 9
2018-07-12 21:32:35,110 INFO spawned: 'bootstrap-sogo' with pid 10
2018-07-12 21:32:35,149 INFO spawned: 'sogo-webres' with pid 11
Waiting for database to come up...
Jul 12 21:32:35 847b7c2c1259 syslog-ng[8]: syslog-ng starting up; version='3.8.1'
2018-07-12 21:32:36,299 INFO success: syslog-ng entered RUNNING state, process has stayed up for > than 1 seconds (startsecs)
2018-07-12 21:32:36,299 INFO success: cron entered RUNNING state, process has stayed up for > than 1 seconds (startsecs)
2018-07-12 21:32:36,299 INFO success: bootstrap-sogo entered RUNNING state, process has stayed up for > than 1 seconds (startsecs)
2018-07-12 21:32:36,299 INFO success: sogo-webres entered RUNNING state, process has stayed up for > than 1 seconds (startsecs)
mysqld is alive
Waiting for database initialization...
Waiting for database initialization...
Waiting for database initialization...
Waiting for database initialization...
Waiting for database initialization...
Waiting for database initialization...
Waiting for database initialization...
Waiting for database initialization...
Jul 12 21:33:01 847b7c2c1259 CRON[39]: pam_unix(cron:session): session opened for user sogo by (uid=0)
Jul 12 21:33:01 847b7c2c1259 CRON[40]: pam_unix(cron:session): session opened for user sogo by (uid=0)
Jul 12 21:33:01 847b7c2c1259 CRON[42]: (sogo) CMD (  /usr/sbin/sogo-ealarms-notify 2>/dev/null)
Jul 12 21:33:01 847b7c2c1259 CRON[43]: (sogo) CMD (  /usr/sbin/sogo-tool expire-sessions 60)
Jul 12 21:33:01 847b7c2c1259 CRON[40]: pam_unix(cron:session): session closed for user sogo
[...]

PR #1566, opened 2018-07-13
Change Umask for update to ensure its 0022

FloThinksPi:
This change sets umask to 0022.
This works around issues like #1561

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -3,6 +3,8 @@
 #exit on error and pipefail
 set -o pipefail
 
+umask 0022
+
 for bin in curl docker-compose docker git awk sha1sum; do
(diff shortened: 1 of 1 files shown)


Knight1:
What issue exactly did you encounter?

FloThinksPi:
@Knight1 I use a default umask of 0027. However it seems as the updater needs a umask of 0022 to succeed. After updating with the script, the docker containers where not functioning anymore and did constantly reboot. (see #1561 for logs ) after adding a "umask 0022" at the beginning of the updater, the instance was working again. I could also reproduce this on another instance when i changed the umask. So making sure the umask in this script is 0022 is a solution to work around issues with more restricted umasks.

andryyy:
Shouldn't be a problem to set it for the subshell. It also inherits to subshells of the calling script (not that it is important as of today, but could become in the future).

Thanks.

PR #1604, opened 2018-07-24
Replace jqPlot with Google Charts

PhoenixPeca:
### Case:
Google Charts is lightweight, has better documentation, mobile ready and more pretty.
__________________________________

Normal view:
image

On-hover Tooltip:
image


andryyy:
Nice. You tested if reloading etc. still works?

PhoenixPeca:
Yes I did.. I just followed what you did in jqPlot, apparently both plugins works (more or less) the same way, just different designs.

andryyy:
Ok. I trust you with this. But don’t push something like this to master without at least discussing it.

ntimo:
Hey,
I actually think the flat chart from before looked better :(

EnriqCG:
It is obviously a matter of opinions, but as @PhoenixPeca mentioned, Google Charts is a more modern and more responsive library. 

This design fits better into the panel theme imo.

PR #1615, opened 2018-07-30
[Web] Improve Input Validation (and error messages)

PhoenixPeca:
The previous condition made it impossible to set a value of "**0**" to the numeric inputs. As you can see on the 2nd item on the list below:

> The following values are matched in the `empty` function
> "" (an empty string)
> **0** (0 as an integer)
> **0.0** (0 as a float)
> "**0**" (0 as a string)
> **NULL**
> **FALSE**
> **array()** (an empty array)

___________________________________________
When you enter "0" in the number field and submit, It shows `Changes to domain phoenixpeca.xyz have been saved` but the value hasn't changed, which I believe is a bug. This is a bug because setting the value as zero should result in an error that explains why mailcow retains the old value.
image
image

### DIFF
--- a/data/web/inc/functions.mailbox.inc.php
+++ b/data/web/inc/functions.mailbox.inc.php
@@ -1818,11 +1818,11 @@ function mailbox($_action, $_type, $_data = null, $attr = null) {
                 $backupmx             = (isset($_data['backupmx'])) ? intval($_data['backupmx']) : $is_now['backupmx_int'];
                 $relay_all_recipients = (isset($_data['relay_all_recipients'])) ? intval($_data['relay_all_recipients']) : $is_now['relay_all_recipients_int'];
                 $relayhost            = (isset($_data['relayhost'])) ? intval($_data['relayhost']) : $is_now['relayhost'];
-                $aliases              = (!empty($_data['aliases'])) ? $_data['aliases'] : $is_now['max_num_aliases_for_domain'];
-                $mailboxes            = (!empty($_data['mailboxes'])) ? $_data['mailboxes'] : $is_now['max_num_mboxes_for_domain'];
-                $maxquota             = (!empty($_data['maxquota'])) ? $_data['maxquota'] : ($is_now['max_quota_for_mbox'] / 1048576);
-                $quota                = (!empty($_data['quota'])) ? $_data['quota'] : ($is_now['max_quota_for_domain'] / 1048576);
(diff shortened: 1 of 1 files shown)


PhoenixPeca:
@andryyy I have tested these modifications, and it indeed fixes the issues.

PhoenixPeca:
I have discovered that this issue has been all over the input fields. I will create a new branch for fixing this issue now.

andryyy:
It should not be possible.

andryyy:
Please do not merge anything to master.

The whole logic behind these fields is fit to this behavior. The input validation happens for a reason as it is. If you change this, you'd need to change a lot more, some of which I don't want to be able to set to 0 (or empty).

PhoenixPeca:
Yes @andryyy I am testing it in a new branch. I will have you test it if it is ready.
The problem is, even the text fields have this behavior, which is why I am fixing it.

I'll let you see some GIF images of the previous behaviors and and how the new modifications fixes it.

PhoenixPeca:
@andryyy No, don't revert. The base branch is not on master.

PhoenixPeca:
@andryyy 
It exists on almost all inputs.
Here's a record of the issue.
[July 31, 2018](https://www.youtube.com/watch?v=P2KuYtvjAAs) YouTube

The text validation works well in mailcow, but the appropriate error messages aren't showing up whenever you enter something like `0` in text inputs.

this is because mailcow is treating numbers and strings with the `empty` function to output the appropriate error message. The problem comes when the string is a `0` because the function returns true.

Hence I am proposing a new approach which uses `$data != ''` as a condition.

PhoenixPeca:
So should I at least fix the issue where the input validation should result an error instead of giving a plain 'successful...' message whenever string "0" is entered? the branch that I have been working on seems to have been deleted.

andryyy:
You also changed that for stuff like aliases. That’s something I do not want to have set to 0. 

For text stuff in your video: For SOME values isset can be used. But often some minor changes like these will break the logic behind it... editing may fail depending on how values are posted to the API etc. (i.e. the way the UI posts to the API vs. external POST). The forms might need changes, api.js etc. 

I think posting 0 as a custom name for any description in mailcow can be ignored. But if you change it, you should make sure it works flawless with any POST to the API, no matter if it is an empty string, 0 or not set, from cURL or the UI. Do not change it in numeric fields. This will be problematic and raise new bugs.

PR #1663, opened 2018-08-13
add smtpd_helo_restrictions

Jasper-Ben:
Some clients (like Outlook) do not perform a valid helo by default.
Softening up the helo requirement for sasl authenticated clients seems reasonable.


mkuron:
This shouldn't be necessary if you submit your messages via port 587 or 465. They don't have HELO restrictions, see https://github.com/mailcow/mailcow-dockerized/blob/master/data/conf/postfix/master.cf . Port 25 must exclusively be used for server-to-server communication, and HELO checks are a commonly-used part of anti-spam strategies.

Jasper-Ben:
Ah, thanks for the info! Apparently the client was wrongly configured. I was told about the problem and wrongly assumed they had it configured correctly.
All good now.

PR #1683, opened 2018-08-17
Fixes idn_to_ascii(): INTL_IDNA_VARIANT_2003 is deprecated

Knight1:
Fixes: idn_to_ascii(): INTL_IDNA_VARIANT_2003 which is deprecated since 7.2 
See https://secure.php.net/manual/en/function.idn-to-ascii.php

### DIFF
--- a/data/web/inc/functions.inc.php
+++ b/data/web/inc/functions.inc.php
@@ -696,7 +696,7 @@ function is_valid_domain_name($domain_name) {
 	if (empty($domain_name)) {
 		return false;
 	}
-	$domain_name = idn_to_ascii($domain_name);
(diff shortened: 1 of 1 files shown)


andryyy:
I will need to run more checks and we will need to replace it everywhere.

andryyy:
This seems to be a matter of "is it fully compatible" instead of "is it deprecated in PHP". This _can potentially_ break compatibility.

I wouldn't want to merge it without being sure about that. It also does not cover all occurrences as said previously.

Knight1:
Still not merged, why?

andryyy:
This PR is missing most occurrences, it is useless. 

I also didn’t test the new format. I gave you all the reasons previously. I will close it, because it will not be merged as it is anyway.

PR #1684, opened 2018-08-17
Add ability to define if Mailcow is running behind Reverse Proxy 

Knight1:
This Setting defines if Mailcow is running behind a proxy Server. If we don't do this Plugins like IPFuck spoof all those Headers and Mailcow is logging random IP's generated by a Plugin :o 

https://chrome.google.com/webstore/detail/ipfuck/bjgmbpodpcgmnpfjmigcckcjfldcicnd?hl=de

### DIFF
--- a/data/web/inc/prerequisites.inc.php
+++ b/data/web/inc/prerequisites.inc.php
@@ -74,6 +74,7 @@ set_exception_handler('pdo_exception_handler');
 // TODO: Move function
 function get_remote_ip($anonymize = null) {
   global $ANONYMIZE_IPS;
+  global $PROXY_SERVER;
   if ($anonymize === null) { 
     $anonymize = $ANONYMIZE_IPS;
   }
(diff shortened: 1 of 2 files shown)


andryyy:
This PR would be fatal to merge to master. You cannot set this to false, not adding any docs and let everyone merge it.

It allows to set IPs from set_real_ip_from. Not sure, but you probably set your local reverse proxy to accept these headers from anywhere? mailcow would indeed accept this, as the last hop is a local network address.

It should not accept these headers from elsewhere.

andryyy:
But it does and it is a bug. I will update the Nginx config and PHP stuff. Thanks!

andryyy:
It is fixed. REMOTE_ADDR will always be set to what we tell Nginx to use as real ip (x-forwarded-for).

Thank you very much for this hint!

Knight1:
Thanks! But behind a proxy I guess that the IP is now the Proxy itself AFAIK.

andryyy:
I checked, Nginx sets the real ip to what it finds in x-forwarded-for - as long as it was set by a trusted source.

Knight1:
Jup I see it. Thanks!

Issue #1218, opened 2018-03-30
Please use OSCP Stapling for the acme Client


ghost:
Hey, I noticed that the build-in certbot function works just great!
But could you please add OSCP Stapling, so I dont have to use my own certificates? It just be pretty easy, because it exists an option (--must-staple).

andryyy:
I don’t really want to enable it. You can change acme-mailcow and other configuration files for your setup. 

I think at least Postfix still has some problems handling it properly. 

I also don’t like the idea to enable fancy options because they exist. :-) We should always make sure what problems may occur and wether or not it actually improves the setup.

ghost:
Okay, thank you for the quick reply. :)
Keep up the good work, mailcow is really, really nice

mkuron:
Dovecot said they might implement it in the future: https://www.dovecot.org/list/dovecot/2016-March/103403.html . Postfix said that they will never implement it: https://marc.info/?l=postfix-users&m=151119229905817 . So Mailcow will never be able to implement OCSP stapling.

As Let's Encrypt only issues 90-day certificates (and renews after 60), the need for stapling is not that big. In essence, stapling reduces the validity period of a compromised certificate to a few days. On average, a Let's Encrypt certificate will have 60 days of validity left at the time it is compromised, so you only shorten the validity of the compromised certificate by a factor of 20 through stapling. Compare that to long-lived certificates (2 years), where stapling gives you a factor of 250.

PR #1688, opened 2018-08-18
Enabling OCSP Stapling

Knight1:
This enables OCSP stapling so that nginx already submits the certificate revocation status (good or revoked) to the client. So the client does not need to do it and thus more anonymity.

https://www.digicert.com/ssl-support/nginx-enable-ocsp-stapling-on-server.htm
https://en.wikipedia.org/wiki/OCSP_stapling

### DIFF
--- a/data/conf/nginx/site.conf
+++ b/data/conf/nginx/site.conf
@@ -29,6 +29,8 @@ server {
   ssl_session_cache shared:SSL:50m;
   ssl_session_timeout 1d;
   ssl_session_tickets off;
+  ssl_stapling on;
+  ssl_stapling_verify on;
 
   add_header Strict-Transport-Security "max-age=15768000; includeSubDomains";
   add_header X-Content-Type-Options nosniff;


andryyy:
There are issues in which we talked about it. 

Closing this. Can’t look it up right now, I’m on mobile.

mkuron:
#1218 was about using OCSP must-staple in the ACME request, which creates certificates incompatible with Postfix/Dovecot. Just using stapling (not must-staple) as proposed in the current PR is actually worth considering as it slightly speeds up opening the TLS connection. I am not aware of any adverse effects this might have, but I have not tested it.

andryyy:
Thank you, I read this and fell asleep just after. :-)

You are right.

chriscroome:
I have used OCSP stapling in web servers with Let's Encrypt certificates and 99.9% of the time it is fine but whenever there is an issue with the Let's Encrypt infrastructure then it results in sites not being able to be loaded by web browsers, so I have disabled it, if you want I could dig up some threads on this from the Let's Encrypt issues on GitHub or their Discourse forum (this is [one memorable example](https://community.letsencrypt.org/t/ocsp-request-failed-with-following-message/50561/26)).

EnriqCG:
[This](https://letsencrypt.status.io/pages/history/55957a99e800baa4470002da/search/OCSP) is the collection of outages Let's Encrypt OCSP responder has suffered. Considering this outages last for a maximum of 24 hours, I don't know how this can break a server with proper OCSP stapling caching. Plus, the Let's Encrypt team is considering moving OCSP responders to another domain to prevent [this](https://letsencrypt.status.io/pages/incident/55957a99e800baa4470002da/5b5f5aa93a343f54d7982864) from happening again.

chriscroome:
>  I don't know how this can break a server with proper OCSP stapling caching

In my experience if you have OCSP stapling enabled on a server then web browsers that support it (all modern ones) won't let you access the sites when the Let's Encrypt infrastructure or the 3rd party services they depend on (their registrar or content delivery network) have issues &mdash; having a option to enable it in `mailcow.conf` would be good but I think it would be a mistake to enable it in a way that people can't switch it off.

Knight1:
Firefox always tries to check ocsp. Chrome does not for performance reasons. The hole PKI is a mess at this point. You can enable chrome checks but this is always soft-fail. If the ocsp is unavailable, slow, ... cert is okay. stapling just puts the answer the browser would do inside the TLS handshake. must-staple is a whole other level. you force the check to pass if not firefox is refusing the connection. But enabling stapling does not hurt anyone.

dragoangel:
Same thought what I wrote to telegram channel. Enable OCSP on nginx, but not generate must-staple flag on acme

PR #1691, opened 2018-08-19
Fix length of remote column in logs table

mkuron:
Otherwise a full non-anonymized IPv6 address won't fit and an error message is printed. This can only happen when `$ANONYMIZE_IPS = false` is set in vars.local.inc.php as otherwise the IPv6 addresses are truncated to a maximum of 21 characters.


andryyy:
Thanks!!

PR #1331, opened 2018-04-27
[Dovecot] Fix to allow the use of auth_default_realm without breaking…

sriccio:
… the master password used by imapsync and others

PR #1721, opened 2018-08-28
[dovecot] Enhancement to allow the use auth_default_realm

sriccio:
When using auth_default_realm in dovecot configuration to setup a default domain name (to allow users to login with only a username instead of the full e-mail address), it breaks the compatibility with imapsync and sogo/sieve vacation.

Adding a domain name to dovecot master user/pass fixes this.

This never made it do the master branch, so I'm trying a PR again :)

See: https://github.com/mailcow/mailcow-dockerized/pull/1331#issuecomment-386534307


mkuron (approved):

sriccio:
thank you! :)

andryyy:
Merci!

Issue #1728, opened 2018-08-30
[SOGo] Forwarding: Max e-mail addresses


jooP0:
Hi! I have set up a forwarding with 52 recipients in SOGo. This isn't working. Is there a limit?

andryyy:
Hi, don't you want to use an alias? I guess there actually is a limit for actions in Dovecot for sieve rules.

mritzmann:
> Is there a limit?

The limit seems to be 30.

https://github.com/mailcow/mailcow-dockerized/blob/775b69f63e2fe4b34463fb636f36e9f4353134a2/data/conf/dovecot/dovecot.conf#L281

PR #1733, opened 2018-08-31
disable sieve_max_redirects limit

jooP0:
see:  #1728

I don't think there are any disadvantages.

sieve_quota_max_scripts and sieve_quota_max_storage are also set to 0. Otherwise, an error message or error handling makes sense, otherwise you will not notice that you are exceeding a limit.


jooP0:
Otherwise, an error message or error handling makes sense, otherwise you will not notice that you are exceeding a limit.

andryyy:
I think that's something for the docs. Can you create a PR with an update for the docs?

I don't think I'd to this with a sieve forwarding rule. Create a BCC map entry with an alias as destination.

PR #1748, opened 2018-09-08
rspamd: never apply KEEP_SPAM to authenticated users

### LINKED ISSUE #1740: Mails not signed with DKIM key if sending host's IP is in "Weiterleitungs-Hosts (WebUI)"
looks a blit like #1091, but here the rebuild/update did not cure the issue. 

config: https://pastebin.com/cfbatu2T
ipv4tables: https://pastebin.com/AfFPuGii
ipv6tables: https://pastebin.com/SeT5jCiS
works for one domain:  https://pastebin.com/UmV4eggU
does not work for other domain: https://pastebin.com/6bHkvTqB

mkuron:
Fixes #1740

### DIFF
--- a/data/conf/rspamd/dynmaps/settings.php
+++ b/data/conf/rspamd/dynmaps/settings.php
@@ -100,6 +100,13 @@ settings {
       }
     }
   }
+  authenticated {
+    priority = 10;
+    authenticated = yes;
+    apply {
+      symbols_disabled = ["KEEP_SPAM"];
+    }
+  }
 <?php
 
 /*


andryyy:
Hm, I don’t think that’s a good solution. Other settings for authenticated users will not be matched anymore (or the other way round).

mkuron:
We don't have any other settings in the dynmap for authenticated users. Or are you referring to custom ones in the `settingsmap` table? I guess those are the user's own responsibility, but we could decrease the priority of the rule in this pull request so that a custom setting always takes precedence.

andryyy:
Yes, I have quite a lot in managed setups. :-)

Give me some hours, I will try to implement it in `data/conf/rspamd/lua/rspamd.local.lua`.

Issue #1723, opened 2018-08-29
specific hostnames for autoconfig / autodiscover


imp1sh:
We want the users to use specific hostnames for the mail client depending on what domain name they use for Email. That means we can migrate the users of that domain in the future more easily, without changing the hostnames in all the clients.

The corresponding DNS records are there already. When the client tries to autoconfigure in Thunderbird it fetches something like this:
https://autoconfig.domain.name/mail/config-v1.1.xml and extracts the hostname from that xml file, right?
The hostnames in that xml files are the default hostname though. Can we somehow specify other names?

What about other clients that do ActiveSync? Do they also fetch that xml or do they auto configure differently? Can we use specific names here instead of the default hostname?
Thank you

mkuron:
Duplicate of #528. This is not supported.

imp1sh:
Can you please consider this a feature request.

andryyy:
Si

RedFoxy:
Any update about that? I prefer to have domain specific dns for imap/pop/smtp too

Issue #1164, opened 2018-03-14
Autoconfig/autodiscover with multiple domains


alexandernst:
Is it possible to make mailcow reply with the autoconfig of the same domain that is requesting said config?

For example, I have `foo.com` and `bar.com` (the `PTR` of my server is `abc.com`).

Currently, whenever a user from any of the two domains tries to auto-configure a mail account, mailcow returns the domain `mail.abc.com`. Is it possible to make mailcow return `mail.foo.com` for the users of `foo.com` that are trying to auto-configure their accounts, and `mail.bar.com` for the users of `bar.com` that are trying to auto-configure their accounts?

mkuron:
Duplicate of #528

PR #1779, opened 2018-09-17
added additional san check for autoconfig hostname and AUTOCONFIG_MAP…

sourcefactory-at:
Quick and dirty rewrite for domain in autoconfig xml response.

With this change it's possible to define the domain which gets printed in autoconfig xml instead of the mailcow hostname.
If nothing is defined or the domain from $_SERVER['HTTP_HOST'] is not configured in ADDITIONAL_SAN the MAILCOW_HOSTNAME is used (as usual).

When the domain in $_SERVER['HTTP_HOST'] matches any domain in ADDITIONAL_SAN this gets used.

As I use mailcow for some domains with their subdomain "mx" this does not fully work.

For example:
mailcow hostname: example1.com
additional san: mx.example2.com
mail address: test@example2.com

example2.com is a website which runs on a different server.
When I add test@example2.com, thunderbird will try autoconfig.example2.com which points to the mailcow instance. 
With the normal san match it will print "autoconfig.example2.com" in the xml which is not correct.

For that case it's possible to add AUTOCONFIG_MAPPING to the .env file with the value "autoconfig.example2.com->mx.example2.com". This will override the value once more so that the correct domain "mx.example2.com" gets printed.

As the commit message says it's just quick and dirty but works for me with Thunderbird 52.

.env for this example:
```
MAILCOW_HOSTNAME=example1.com
ADDITIONAL_SAN=mx.example1.com,mx.example2.com,autoconfig.example2.com,autodiscover.example2.com,mx.example3.com,autoconfig.example3.com,autodiscover.example3.com
AUTOCONFIG_MAPPING=autoconfig.example2.com->mx.example2.com,autoconfig.example3.com->mx.example3.com
```

### DIFF
--- a/data/web/autoconfig.php
+++ b/data/web/autoconfig.php
@@ -8,7 +8,7 @@ $autodiscover_config = array_merge($default_autodiscover_config, $autodiscover_c
 
 error_reporting(0);
 
-if (empty($mailcow_hostname)) {
+if (empty($autoconfig_hostname)) {
   exit();
 }
 
@@ -27,7 +27,7 @@ header('Content-Type: application/xml');
 ?>
 <?= '<?xml version="1.0"?>'; ?>
 <clientConfig version="1.1">
-    <emailProvider id="<?=$mailcow_hostname; ?>">
+    <emailProvider id="<?=$autoconfig_hostname; ?>">
       <domain>%EMAILDOMAIN%</domain>
       <displayName>A mailcow mail server</displayName>
       <displayShortName>mail server</displayShortName>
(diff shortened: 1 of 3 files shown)


mkuron:
See #528. Hack/workaround for #1164, #1723.

ThomDietrich:
Hello @mkuron,
I am yet another interested user of this feature or any other solution to the autoconfig issue. Your comment included some references but no review. What's your opinion on this PR? It would be amazing to see this PR or a refined version of it merged soon. I'd be happy to assist.

@sourcefactory-at main thing about this PR that stands out as "hacky" is the "`->`" pseudo-syntax. Did you consider to choose a standardized syntax, like json?

PR #1794, opened 2018-09-25
grep -i "BusyBox"

JOduMonT:
Hi;
I was testing if mailcow will run on Alpine Linux than during my validation I found this 

grep -i "BusyBox"
because grep -q -i "busybox" return nothing

### DIFF
--- a/generate_config.sh
+++ b/generate_config.sh
@@ -2,11 +2,11 @@
 
 set -o pipefail
 
-if grep --help 2>&1 | grep -q -i "busybox"; then
+if grep --help 2>&1 | grep -i "BusyBox"; then
   echo "BusybBox grep detected, please install gnu grep, \"apk add --no-cache --upgrade grep\""
   exit 1
 fi
-if cp --help 2>&1 | grep -q -i "busybox"; then
+if cp --help 2>&1 | grep -i "BusyBox"; then
   echo "BusybBox cp detected, please install coreutils, \"apk add --no-cache --upgrade coreutils\""
   exit 1
 fi


andryyy:
Hi,
It returns nothing but quits with a non-zero exit code due to `-q` if the string was not found. It just suppresses unwanted output. It _should_ work as intended. Doesn't it for you?

JOduMonT:
maybe I mix two things but
`grep -i "busybox"` doesn't work
it's looks case sensitive even if -i is there

and now I'm confuse because on a brand new alpine 3.8
`-q -i` return this

> ~ # echo $0
> /bin/ash
> ~ # grep --help|grep -q -i "busybox"
> BusyBox v1.28.4 (2018-07-17 15:21:40 UTC) multi-call binary.
> 
> Usage: grep [-HhnlLoqvsriwFE] [-m N] [-A/B/C N] PATTERN/-e PATTERN.../-f FILE [FILE]...
> 
> Search for PATTERN in FILEs (or stdin)
> 
>         -H      Add 'filename:' prefix
>         -h      Do not add 'filename:' prefix
>         -n      Add 'line_no:' prefix
>         -l      Show only names of files that match
>         -L      Show only names of files that don't match
>         -c      Show only count of matching lines
>         -o      Show only the matching part of line
>         -q      Quiet. Return 0 if PATTERN is found, 1 otherwise
>         -v      Select non-matching lines
>         -s      Suppress open and read errors
>         -r      Recurse
>         -i      Ignore case
>         -w      Match whole words only
>         -x      Match whole lines only
>         -F      PATTERN is a literal (not regexp)
>         -E      PATTERN is an extended regexp
>         -m N    Match up to N times per file
>         -A N    Print N lines of trailing context
>         -B N    Print N lines of leading context
>         -C N    Same as '-A N -B N'
>         -e PTRN Pattern to match
>         -f FILE Read pattern from file

PR #1801, opened 2018-09-26
Update docker-compose.yml

mcmufffin:
update for latest dovecot mysql db error

after latest update dovecot stopped working

auth-worker(113): Error: mysql(/var/run/mysqld/mysqld.sock): Connect failed to database (mailcow): Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2 "No such file or directory") - waiting for 1 seconds before retry

this will add a volume for the socket


andryyy:
I fixed the image, this is part of another update...

PR #1823, opened 2018-10-02
google-style wildcard-emails

JoeRu:
This pull-request enables google-style wild-card emails. for example you can use 
`email+shopname@example.org` for one shop and
`email+othershop@example.org` for another shop - and all are going into the same account `email@example.org`.

I'm using this technique about 10 years - and in combination with sieve this generates a big plus on where are my mails stolen from and to shut them finaly without terminating an account.

One flaw is present - the seperator - in this case the '+' is not allowed any longer as a regular part of the username. 

The config has no impact on mails not using the wildcard-symbol.

### DIFF
--- a/data/Dockerfiles/postfix/postfix.sh
+++ b/data/Dockerfiles/postfix/postfix.sh
@@ -14,7 +14,7 @@ newaliases;
 cat <<EOF > /opt/postfix/conf/sql/mysql_relay_recipient_maps.cf
 user = ${DBUSER}
 password = ${DBPASS}
-hosts = mysql
+hosts = unix:/var/run/mysqld/mysqld.sock
 dbname = ${DBNAME}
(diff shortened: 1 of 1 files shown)


MAGICCC:
Isn't it the same as [Tagging](https://mailcow.github.io/mailcow-dockerized-docs/u_e-mailcow_ui-tagging/), except you can't select 'Put all mails in my inbox and not creating a new folder or prepend a tag'?

JoeRu:
pretty much; i didn't see it yet; thx. Also i use a '-' as a delimiter. But i'll check it out. 
If this works out of the box - i guess there is no need to add this up to the main branch - and i'll be happy with my '-'.

JoeRu:
yup... works as well. No need to merge this one...

PR #1825, opened 2018-10-03
Don't print 'Collecting garbage...' twice

MAGICCC:
When updating mailcow and it finished it printed 'Collecting garbage..,' twice since in docker_garbage() and before running the code we have the echo command.
Now it's printing the line before running docker_garbage()


andryyy:
Thanks!

PR #1838, opened 2018-10-04
Add missing close buttons to modal dialogs

heavygale:
There were a few close buttons (and the heading for "change password") missing in the modal dialogs, this pull requests will add them.


andryyy:
I didn't add them for some modals for a reason. They should not be closed, esp. some TFA stuff relies on events when the close button is used. :-/ We would need to check the modals first...

heavygale:
Ok, so there are events getting triggered when the user closes the dialogs by clicking into the transparent grey background, wich aren't triggered with the close button?

andryyy:
I'm not sure, I just remember I added this. :-( I'd need to check the code. It is quite possible that the event gets triggered with a click on the "X", too, and we don't need to change anything.

PR #1840, opened 2018-10-05
Add portainer and docker-compose.override.yml to .gitignore

Knight1:
Adding portainer and docker-compose.override.yml removes git merge errors while updating.


andryyy:
Thanks

PR #1847, opened 2018-10-05
[NGINX] always redirect to https

D3luxee:
The current nginx configuration does not redirect from http to https, even though the HSTS header will ensure the usage of https it is not recommended to send HSTS Headers via http initially.

Therefore we should actively redirect to https using a seperate server block for http.

Letsencrypt does follow the redirect and is working with snakeoil certificates.

Is there any reason for not redirecting to https?

Regards,
D3luxee

### DIFF
--- a/data/conf/nginx/site.conf
+++ b/data/conf/nginx/site.conf
@@ -17,6 +17,15 @@ map $sent_http_content_type $expires {
 }
 
 server {
+  include /etc/nginx/conf.d/listen_plain.active;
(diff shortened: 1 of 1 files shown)


andryyy:
How do you handle reverse proxy setups? I don't want to do double TLS here for no reason. There were other PRs iirc where we explained it. :-)

You also don't need a new server block for this to work.

I thought about a header we only accept from the RP (or from internal connections) like x-use-plain or something like this. But this needs to be announced and added to the docs.

PR #1850, opened 2018-10-06
Better update for docker-compose

Knight1:
No longer depend on docker-compose for the update. If it is missing we try to download and install it
No longer depend on external docker-compose version check
Checking sha256 against original file
Only download latest official release from github.com

Fixes: https://github.com/koalaman/shellcheck/wiki/SC2230

What do you think, anything wrong or missing?

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -3,7 +3,7 @@
 #exit on error and pipefail
 set -o pipefail
 
-for bin in curl docker-compose docker git awk sha1sum; do
+for bin in curl docker git awk sha1sum sha256sum; do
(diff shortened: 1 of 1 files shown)


andryyy:
Why would Docker compose not be installed when you already run mailcow? Plus don’t remove the Servercow links, as stated in the other PR.

Knight1:
I don't know for sure but this limits the possibility for problems. 

The other PR only mentions problems with future versions. in this case we can use my code with your domain or simply hardcode the version. You currently update it manually. We only put that hardcoded version into the project and thus we prevent some errors if your server is offline.

PR #1851, opened 2018-10-06
Fixes some bash lint problems within update.sh

Knight1:
Fixes:
https://github.com/koalaman/shellcheck/wiki/SC2068
https://github.com/koalaman/shellcheck/wiki/SC2230
https://github.com/koalaman/shellcheck/wiki/SC1117
https://github.com/koalaman/shellcheck/wiki/SC2053
https://github.com/koalaman/shellcheck/wiki/SC2206

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -4,7 +4,7 @@
 set -o pipefail
 
 for bin in curl docker-compose docker git awk sha1sum; do
-  if [[ -z $(which ${bin}) ]]; then echo "Cannot find ${bin}, exiting..."; exit 1; fi
+  if [[ -z $(command -v ${bin}) ]]; then echo "Cannot find ${bin}, exiting..."; exit 1; fi
 done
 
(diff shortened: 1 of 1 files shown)


andryyy:
You quote every single word variable, I don't understand why? mailcow options are _always_ single words, just as repository names etc. - it is not wrong so I'm fine if you keep them.

Stuff like `echo -e "\\e[32mCollecting garbage...\\e[0m"` is not related to SC1117. You add an extra \ for no reason.

command is indeed better, yes.

"for" loops can break when you loop with for bla in "$bla", it depends on the context. It works because it was never broken here.

I really don't like the idea of following guides blindly. You always need to understand what _exactly_ they are doing and why. Are you okay if we cherry-pick from your changes and merge?

Knight1:
"Are you okay if we cherry-pick from your changes and merge?" sure as long as my contributions are useful :)

PR #1864, opened 2018-10-08
Virtual Folders

JoeRu:
So this is more a feature add up - the 3 standard virtual folders last14, marked and unread are about to  safe time in mailboxes with many subfolders.


Braintelligence:
Is your "wildcard" feature any different than this: https://mailcow.github.io/mailcow-dockerized-docs/u_e-mailcow_ui-tagging/ ?

JoeRu:
It was (nearly the same) - so i removed this later on and merged today with the latest.

Braintelligence:
So what's left now is achieving virtual folders that gather the last 14 days, marked and unread mails?

JoeRu:
in short: yes. :)

Braintelligence:
OK, why not if it won't break anything or conflict with design principles :). Let's see what andryyy will say.

mkuron:
That's neat, but shouldn't be enabled by default. Most email clients do not deal gracefully with having the same message in multiple folders and e.g. download it multiple times for offline viewing. Also, some email clients (smartphones, Apple Mail) don't allow you to selectively subscribe to folders, so you couldn't even switch it off client-side.

I'll reopen the PR if you figure out how to make it optional. If you can't, then maybe this is better put into the documentation.

Also, your PR overwrites some unrelated things, which you probably did not intend.

PR #1888, opened 2018-10-13
Added .github folder to .gitignore file

ntimo:
Hey @andryyy I added the .github folder to the .gitignore this way the update script won't download it since its not need on the Mailcow host.


mkuron:
I don‘t think this even works. The updater script has nothing to do with .gitignore.

andryyy:
I'm also unsure... but will likely not hurt.

PR #1889, opened 2018-10-13
remove duplicated import time

Knight1:


andryyy:
I won't rebuild, as it would not change anything, but it is a good catch, thanks.

PR #1896, opened 2018-10-14
Fixed issue template for bugs

ntimo:
Sorry, now it should work :) I forgot to add the header that's why its not showing up in the Github selection menu.


andryyy:
As long as you don't push code, it is okay. The templates are not that important.

ntimo:
Yeah sorry, I didn't meant to break it. Will be more careful next time :)

PR #1899, opened 2018-10-15
Hide autodiscover records on DNS page for alias domains

mkuron:
This matches the change in a9eddae686d031514702f99cbdda9b95e6040580. The autodiscover and service records were never needed for alias domains.


andryyy:
Thanks!

PR #1949, opened 2018-10-25
[Postfix] Security: Prefer server-side ciphers

patschi:
Prefer server-side ciphers to prevent client-side cipher downgrade. Already enabled in Dovecot.

For more information: http://www.postfix.org/TLS_README.html

### DIFF
--- a/data/conf/postfix/main.cf
+++ b/data/conf/postfix/main.cf
@@ -99,6 +99,7 @@ lmtp_tls_protocols = !SSLv2, !SSLv2, !SSLv3
 smtpd_tls_mandatory_protocols = !SSLv2, !SSLv3
 smtpd_tls_protocols = !SSLv2, !SSLv3
 smtpd_tls_security_level = may
+tls_preempt_cipherlist = yes
 tls_ssl_options = NO_COMPRESSION
 smtpd_tls_mandatory_ciphers = high
 virtual_alias_maps = proxy:mysql:/opt/postfix/conf/sql/mysql_virtual_alias_maps.cf,


K2rool:
That option is already set on port 587 https://github.com/mailcow/mailcow-dockerized/blob/70a914a98cc8aa992fff0b06848746c0e50c12d5/data/conf/postfix/master.cf#L7-L11.

Not sure if just set on port 587 is enough I will let someone answer that.

patschi:
Yeah, that's true. But for example not for port 465 (SSL/TLS). IMHO it's generally a good idea to enable it by default.

mkuron:
This option should be enabled on port 465 since it is already on 587. it should probably not be enabled on port 25, which this pull request does, as it might prevent some remote mail servers from delivering TLS-encrypted messages to Mailcow.

mkuron:
Fixed in bf71f9b

patschi:
In my opinion this is the wrong way to go - `tls_preempt_cipherlist=yes` should be set by default for security purposes, specially in these days where secure transport is a very critical thing. Even between two mailservers - and 25 is the default port to receive/send mails.

I think no serious mailserver (and I don't consider Exchange 2003 as one, referring to [Postfix docs](http://www.postfix.org/postconf.5.html#tls_preempt_cipherlist)) will have any issues with that setting. [Heinlein](https://www.heinlein-support.de/blog/security/perfect-forward-secrecy-pfs-fur-postfix-und-dovecot/) explains how to globally enable this option, also [Thomas Leister](https://thomas-leister.de/mailserver-debian-stretch/) explicitly added instructions to enable this.

Feel free to do some tests on your own, using [testssl.sh](https://github.com/drwetter/testssl.sh). (only showing relevant "NOT OK"-parts)
```
 Testing cipher categories
 Weak 128 Bit ciphers (SEED, IDEA, RC[2,4])    offered (NOT ok)

 Testing server preferences
 Has server cipher order?     nope (NOT ok)
 Negotiated cipher            ECDHE-RSA-AES256-GCM-SHA384, 521 bit ECDH (P-521) (limited sense as client will pick)
 Negotiated cipher per proto  (limited sense as client will pick)
     ECDHE-RSA-AES256-SHA:          TLSv1, TLSv1.1
     ECDHE-RSA-AES256-GCM-SHA384:   TLSv1.2
 No further cipher order check has been done as order is determined by the client

 Testing vulnerabilities
 Secure Client-Initiated Renegotiation     VULNERABLE (NOT ok), potential DoS threat
```

Even mailservers by Google are using server-side ciphers for better security, feel free to test that yourself. And I think Google would have never enabled that, if that decision would bring dramatic problems - specially with the very, very huge and enormous user-base.

Some proofs:
- Google with mail.google.com: ```./testssl.sh -t=smtp gmail-smtp-in.l.google.com:25```
- Google with GSuite Business customers: ```./testssl.sh -t=smtp aspmx.l.google.com:25```
- Microsoft with Outlook.com: ```./testssl.sh -t=smtp outlook-com.olc.protection.outlook.com:25``` (Yep, even Microsoft!)
- Heinlein with Mailbox.org: ```./testssl.sh -t=smtp mx1.mailbox.org:25```
- Protonmail: ```./testssl.sh -t=smtp mail.protonmail.ch:25```

I would kindly ask to rethink that decision.

andryyy:
it might prevent some remote mail servers from delivering TLS-encrypted messages

Compatibility should be a bigger concern than having ultra-strong ciphers and settings, which _may_ not be supported by the remote server, which then results in a fallback to a plain session.

And I _really_ don't like the idea of "XY is doing it, too" or "XY wrote it here". :-)

I still understand that there a many valid reasons for server-side ciphers, so I'm definitely open for it.

patschi:
Of course compatibility should play a big role, but security does as well. And yes, blindly following the strategy "because others do" isn't always the ideal way.

But when couple of the largest mail providers in the world (and I think Google Mail is quite heavily used out there) have a good reason to enable server-side ciphers, I'm quite sure they know what they do. They have a lot of smart people working there, a lot of resources and security researchers. If this would be a bad decision, many people would be effected.

Furthermore I think that Heinlein is one of the most known experts when talking about the complex topic mailservers. So in my opinion, they know what they do.

I'm also happy to be corrected in my assumptions that probably server-side ciphers is a bad idea. but when the biggest players and hosters with billions users around the globe decide for enabling server-side ciphers it can't be that a bad idea.

Have you any proofs that this causes any noticable bad side-effects?

extremeshok:
@patschi my views are basically the same as yours.

andryyy:
Security does play a role indeed. That's why a plain text session would suck. We don't depend on an encrypted session, that's the problem.

I don't know how many or which connections would fall back to plain text because of this.

mkuron:
All relevant mail servers will probably support the TLS parameters suggested here, and I‘d trust Heinlein‘s judgement (Google is a different story, they might be doing fingerprinting to detect whether the client is new enough to deal with these parameters). We would probably only fall back to plain-text with some ancient mail servers that likely haven‘t received security patches recently either.

The oldest Linux distribution that is still in support that I can think of is CentOS/RHEL 6. If someone could spin up a Postfix on that and check whether it is able to negotiate an encrypted connection with Mailcow with that option set, I think it would be safe to use. Older mail servers often don‘t even deliver mail over TLS at all. Similar checks should also be done with the oldest supported Exchange Server and maybe some other widespread servers.

Could everyone who commented on this issue please check the Received headers of all mails in their inbox? That should show the cipher that was used to deliver a message, so we would easily be able to make a list of servers that still require weak ciphers. If none of us have received any important messages in the past two years that were delivered with weak ciphers, we don‘t need them anymore.

K2rool:
I just test sent a mail from Centos 6 vps I just created to test this setting tls_preempt_cipherlist to yes dose not break encrypted email from a Centos 6 host.

Server versions
<pre>
[root@uk /]# cat /etc/centos-release 
CentOS release 6.10 (Final)
[root@uk /]# postconf -d | grep "mail_version ="
mail_version = 2.6.6

</pre>

Log from Mailcow server <pre> Anonymous TLS connection established from unknown[185.****]: TLSv1.2 with cipher ECDHE-RSA-AES256-GCM-SHA384 (256/256 bits) </pre>

andryyy:
@mkuron Feel free to merge

mkuron:
So I've checked `Received` lines in my mailboxes. Almost all messages delivered to my server in recent years use TLSv1.2 (`using TLSv1.2 with cipher`). The only exceptions were TLSv1.0 with `ECDHE-RSA-AES256-SHA` (two small companies' mail servers), `DHE-RSA-CAMELLIA256-SHA` (an automated message) and `DHE-RSA-AES256-SHA` (two automated messages and one from AOL). All of these ciphers are still considered secure, so I think `tls_preempt_cipherlist=yes` would be safe.

PR #1958, opened 2018-10-27
Hide 'TLS policy maps'-Tab from Domainadmins

heavygale:
The tabpanel is already hidden, but the nav-tab was still visible.


andryyy:
Oh, good catch, thanks!

PR #1997, opened 2018-11-11
Add delimiter '-' for mail adress tagging

8ear:
Hi,
is it possible to add '-' as delimiter for mail adress such: mail_tag@example.com AND mail+tag@example.com?

Best regards


hachre:
Your example uses `_` but your patch uses `-`. Which one is the one you want?

andryyy:
`-` is too widely used, same for `_`. 

You can change that for your own installation, but there is no way we can merge this globally.

Adorfer:
Perhaps make it configurable.
But off course this has to be checked then 

- creation of new users (against delimiters) 
- modification of delimiters (against existing users)

Klexx:
> Perhaps make it configurable.

This could be useful since some webpages do not accept addresses using the '+' delimiter.

PR #2016, opened 2018-11-16
Make the update check in update.sh readonly

fabianonline:
Currently, the update check in update.sh (`update.sh --check`) runs `git fetch` to fetch the newest revision from Github before using it to check if there is a newer version of mailcow available.

While this is working, it does need write access to update the working copy. If you want to use that script with an automatic monitoring system like zabbix, you possibly don't want the user executing the script to be able to modify (any) files within mailcow's directory.

This PR replaces `git fetch` followed by `git rev-parse` by `git ls-remote`. This way, only read access is necessary for the script to determine if an update is available.

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -66,7 +66,7 @@ while (($#)); do
       if [ $? -ne 0 ]; then
         echo "A problem occurred while trying to fetch the latest revision from github."
         exit 99
-	  fi
+      fi
       if [[ -z $(git log HEAD --pretty=format:"%H" | grep "${LATEST_REV}") ]]; then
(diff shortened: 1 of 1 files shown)


fabianonline:
Any news on merging this? I still think it is a useful change and it doesn't touch important internal stuff...

hachre (approved):
I like this change, tested it and looks good to me. 
Only negative thing I noticed: there is a slight code formatting issue with that `fi` on line 69.

andryyy:
Sorry, totally forgot about it. :(

PR #2020, opened 2018-11-16
[Postfix] Disable weak 128 bit ciphers

patschi:
See #1949


andryyy:
No, better (almost) any encryption than none.

Server-side ciphers are fine, I will merge it, but disabling every not-so-secure cipher on an **optional** TLS channel is not okay.

I have seen soo many setups, where communication fell back to plain-text, just because some website recommended to use this and that cipher.

You can force super secure ciphers on mandatory TLS channels, that's a fairly good idea. We would in no way accept weak ciphers there then. 👍 
So if you want to adjust the PR for it, we can merge it.

Issue #1954, opened 2018-10-26
Can't Share to Any authenticated User in Sogo


isclever:
**Describe the bug**
When trying to share an imap folder or Calendar to "Any authenticated User" in Sogo the option does not appear.

**How to reproduce Reproduce**
Steps to reproduce the behavior please explain how to reproduce the bug:
1. Attempt to share the calendar
2. No "Any authenticated User" is available for selection.

**Expected behavior**
Authenticated User should be avaiable given this documentation: https://www.svenskadomaner.se/en/support/1920

**Screenshots**
![screenshot from 2018-10-26 13-54-41](https://user-images.githubusercontent.com/16140846/47584370-31aada80-d928-11e8-9004-9c18ebedaa68.png)

**System**
 - OS: CentOS 
 - Browser: Firefox

**Additional context**
This is likely a configuration option that I can't find to enable.

andryyy:
https://mailcow.github.io/mailcow-dockerized-docs/u_e-dovecot-any_acl/

isclever:
Thank you.

PR #2028, opened 2018-11-20
Allow setting ACL_ANYONE in the configuration

tisoft:
This makes sharing to "Any authenticated User" (#1954) configurable. Default is disabled as it is currently the case.


andryyy:
You even revert the patch, nice. :)

Let me run some tests, okay?

tisoft:
Yeah reverting the patch has the benefit of allowing an admin to enable and disable this at will. :)

Maybe the location of the "acl_anyone" configuration file is not the best, since that is outside the container and someone might think he can fiddle around with it. It will be overwritten at container restart anyway... So maybe something inside the container like /tmp is more appropriate.

Using ```acl_anyone = </usr/local/etc/dovecot/acl_anyone``` to read the value from a file was the only solution i found to dynamically change the dovecot config based on an environment variable without doing a sed replacement. Unfortunately %{env:ACL_ANYONE} replacement does not work here, since the string is not evaluated by the acl plugin. It is directly string compared. Also no newlines may be contained in the file. All learned the hard way. So if you think of simplifying this, I already tried some simpler methods and failed. :)

K2rool:
Should this line https://github.com/mailcow/mailcow-dockerized/blob/e10db26825fe0d6915d826511207dcdfac2ae4cd/docker-compose.yml#L150 not be like this <pre> - ACL_ANYONE=${ACL_ANYONE:-disallow} </pre> that way the default value is also set in docker-compose.yml not ever user will use update.sh some user with heavily modified install will be doing manual updates so unless the user adds it to mailconf.conf it will be undefined.

tisoft:
I have added the default to the compose file. Thanks for catching that!

andryyy:
Can you check the PR? Something seems wrong.

And sorry, totally forgot about it. Will merge it asap, but please check it first. :)

tisoft:
Updated to latest master

andryyy:
Niiiice, thank you!

PR #2032, opened 2018-11-22
Change name of .mobileconfig file

vain90:
Sets the filename of the mobileconfig as <mailaddr>.mobileconfig.
So the Admin or Domain admin can login as the user and generate the files for his users.
Useful for very inexperienced users.


mkuron:
The file extension was correct already, so I don't see the point of this pull request. You are only changing the file name, and that is kind of pointless if the UUIDs in the files stay the same. A while ago, there was #1008 that change to random UUIDs so you could have multiple accounts, but that never got merged to master unfortunately.

PR #2043, opened 2018-11-26
Avoid exposing MySQL Root password through the cmdline during a backup

curdbecker:
The current version of the backup script accidentally exposes the MySQL Root password through the process cmdline during a backup. 

This is a potential security risk to the MySQL database, since the cmdline of a process can even be accessed by non-root users regardless of whether the process is in a docker container or not. For instance, a simple execution of a small shell snippet like `while true; do ps aux | grep -E [m]ysqldump; sleep 1; done` could (most likely) directly expose the root password during a backup. 

The fix deals with the two possible possibilities of exposure as follows:
- The password is directly supplied in the cmdline of the `docker run` command and then also in the cmdline of the `sh` process executed inside the container.
 -> This can be dealt with by exporting the `DBROOT` variable itself to the container. 
- The password is then supplied to `mysqldump` for performing the actual backup again using the cmdline.
->`mysqldump` supports the use of defaults files to set various environment options. Such a file containing the password can be created dynamically before executing `mysqldump` and can be used instead afterwards. The creation of the file using `echo` itself is uncritical, since echo is a shell builtin and therefore won't show up as an own process.

### DIFF
--- a/helper-scripts/backup_and_restore.sh
+++ b/helper-scripts/backup_and_restore.sh
@@ -92,13 +92,16 @@ function backup() {
         debian:stretch-slim /bin/tar --warning='no-file-ignored' -Pcvpzf /backup/backup_postfix.tar.gz /postfix
       ;;&
     mysql|all)
+      export DBROOT
       SQLIMAGE=$(grep -iEo '(mysql|mariadb)\:.+' ${COMPOSE_FILE})
       docker run --rm \
         --network $(docker network ls -qf name=${CMPS_PRJ}_) \
         -v $(docker volume ls -qf name=${CMPS_PRJ}_mysql-vol-1):/var/lib/mysql/ \
         --entrypoint= \
+        -e DBROOT \
         -v ${BACKUP_LOCATION}/mailcow-${DATE}:/backup \
-        ${SQLIMAGE} /bin/sh -c "mysqldump -hmysql -uroot -p${DBROOT} --all-databases | gzip > /backup/backup_mysql.gz"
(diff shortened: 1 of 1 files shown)


andryyy:
Hm, it would still expose the password for a short time when it echos it to that file.

Oh, I am blind, you wrote:

    The creation of the file using echo itself is uncritical, since echo is a shell builtin and therefore won't show up as an own process.

Is this correct? Have never checked it. No way to still read it? It'd think it still shows as the commands parameter.

curdbecker:
No, it doesn’t. echo is a builtin for bash, so it won’t be shown as an independant process if I am not mistaken :)

andryyy:
It is shown as parameter, just checked it.

curdbecker:
Where did you test it exactly? I can't reproduce it.  Does just the `echo` part of the command show up? Or what does show up at all?

I changed the patch now that it uses `bash` instead of `sh`, because I am using also `bash` in my patched version of the backup script (support for borgbackup). I figured that `sh` could be using `/bin/echo` instead of the shell builtin command, but that's also apparently not the case. Neither for `bash`. Just check the output below for reference.

```
# docker run --rm -it mariadb:10.2 /bin/sh -c "type echo"
echo is a shell builtin
```
```
# docker run --rm -it mariadb:10.2 /bin/bash -c "type echo"
echo is a shell builtin
```

curdbecker:
Oh, and I just realised that I forgot the `-e` option for `echo` when switching the shell from `sh` to `bash`. I fixed it now. Sorry about that :)

PR #2056, opened 2018-12-01
Login page layout update.

mseed:
In order to solve a problem with confusing users with default login dialog I changed login page. Expandable area for login form and separate area for apps.


mseed:
Some previews: http://fastlinux.eu/_mailcow/mc01.png http://fastlinux.eu/_mailcow/mc02.png http://fastlinux.eu/_mailcow/mc03.png

Images have non default font (default font has problem with Czech special chars) and different buttons size which is not included in this pull request.

andryyy:
Hi, thanks, but I'm fine with the current mask. You can go ahead and adjust it for your installation any time.

Adorfer:
this would help a lot of confused users here

Issue #1341, opened 2018-04-30
Feature Request: DMARC report parser


normanu:
It would be nice to have a parser automaticly readout the incoming reports mailbox and work it into a report like at 

https://dmarcian.com/dmarc-xml/

andryyy:
We could indeed add something like that. Not much effort. We would also need to add it to our DNS check.

We could use something like dmarc-reports@MAILCOW_HOSTNAME. A MX record does not need to exist, the A/AAAA record will be used as fallback. We can then parse it through a script and, it is a valid XML report, write it to our database.

normanu:
I see http://dmarc.postmarkapp.com/ does it for free, but it might be better to keep it on the own server?

mkuron:
http://dmarc.postmarkapp.com/ only supports a single domain per account. https://www.dmarcanalyzer.com also has a free tier, but it is quite limited. Building something into Mailcow would be nice, but doesn't need to be a super high priority -- one can deploy DMARC just fine without monitoring these reports.

ntimo:
I just wanted to drop my opinion here too, I would love to have a dmarc analyzer build into mailcow. I currently use http://dmarc.postmarkapp.com/. But using a build in solution would be really great also in terms of privacy.

ledodev:
Solved that for my personal setup with these:
https://github.com/techsneeze/dmarcts-report-parser
https://github.com/techsneeze/dmarcts-report-viewer
Mailcow-official would be nice :)

ntimo:
Would it be possible to implement this into Mailcow? 
https://github.com/techsneeze/dmarcts-report-parser
https://github.com/techsneeze/dmarcts-report-viewer

SomeGeek:
Any updates on this @andryyy ? DMARC and other mail security features (like SRS) are increasingly important these days. It can be integrated individually, however, such important things should be part of the Mailcow core.

It somehow feels these things are not getting the priority it deserves.

andryyy:
A DMARC parser will gain you security? There are a million services to parse them with nice reports etc. DMARC already is part of mailcow (Rspamd).

I don’t understand how a parser deserves higher priority. :-/ It’s just a nice to have, I guess.

SomeGeek:
In my opinion DMARC doesn't make sense when you don't have insights in it. Sure, it can be done right now, but it should be a core feature.

andryyy:
[...]

PR #2065, opened 2018-12-03
Integrating DMARC reports

Clete2:
Integrates DMARC reports as per #1341.

Please provide feedback -- it is my first time contributing here!

You should:
- Create a postmaster mailbox
- Set up DMARC reports to go there
- Configure DMARC as shown in the config file, giving it the IMAP credentials
- Go to admin console and find the new link in the dropdown

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ data/assets/ssl/*
 .vscode/*
 data/web/.well-known/acme-challenge
 data/web/nextcloud/
+data/web/dmarcts-report-viewer/dmarcts-report-viewer-config.php
 data/conf/rspamd/local.d/*
 data/conf/rspamd/override.d/*
 !data/conf/nginx/dynmaps.conf
@@ -28,3 +29,4 @@ data/gogs/
 data/conf/sogo/plist_ldap
 .github/
 docker-compose.override.yml
+.DS_Store
--- a/data/Dockerfiles/dmarcts/Dockerfile
+++ b/data/Dockerfiles/dmarcts/Dockerfile
@@ -1,0 +1,29 @@
+FROM debian:stretch-slim
+LABEL maintainer "Clete Blackwell <clete@blackwell.email>"
+
+ARG DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update
+
+RUN apt-get install -y --no-install-recommends apt-utils
+
+RUN apt-get install -y --no-install-recommends \
+	libmail-imapclient-perl \
+	libmime-tools-perl \
+	libxml-simple-perl \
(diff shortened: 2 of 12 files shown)


ntimo:
This is awesome 😍 I really hope that @andryyy will approve your merge request this would be such a awesome feature.

ntimo:
Could it be that the dmarcts-report-viewer.php is missing? :)

Clete2:
Probably. I’ll sign on my laptop later and make sure I commit it.

mkuron:
This looks interesting, thanks for posting. Before we merge this, we should try to get a smoother and tighter integration though. The user should not need to manually create an IMAP inbox, that needs to somehow happen automatically. One shouldn’t use the postmaster mailbox as that is intended for a human admin; instead something like dmarc-reports@$MAILCOW_HOSTNAME should be the default.

Also, the DNS diagnostics page should show how to put the reporting address into DNS.

Futhermore, do we really need a separate container? There are PHP solutions available too (linked in #1341) that could run directly from the web interface.

andryyy:
Thanks, but I think a new container is too much for those reports. I’m not a supporter of this integration anyway, so it should really be as easy and smooth as possible.

Knight1:
If the PHP solutions are great I would prefer them. Also it is not really necessary because there are so many solutions out there. Also the setup should be automated inside the Admin page and we need a configurator for the DMARC system.

Clete2:
I’m personally not very fond of the solution I found because it involves storing credentials in the script. I’d rather make a custom solution but I thought that integrating existing scripts would be an easier route to at least get some kind of reporting done.

I’m ok with putting it in another container but have no idea which one to use as a cron job or event loop. Hence I made another container for it.

I didn’t see any PHP solutions linked in the issue. I just saw 3rd party cloud based ones and the Perl one I pulled in. Totally understand trying to keep with 1 language.

Clete2:
Also I’m not a php dev. While I can hack it together I’m not sure I’d be able to do it quickly.

Clete2:
Hey all,
I did some research. I recommend [parsedmarc](https://domainaware.github.io/parsedmarc/) as a self-hosted solution.

I run Splunk on my internal network at home. Parsedmarc is a Python utility that parses DMARC reports from an inbox and will ingest them into either Elasticache (for Graylog or Kibana) or Splunk.

For anyone reading this wanting to use it with Splunk, here's the command I use!

`parsedmarc -u <DMARCEmail> -p <MyPassword> -H <MyMailServer> --save-aggregate --save-forensic --hec https://<MySplunk>:8088 --hec-token <MyToken> --hec-index Default --hec-skip-certificate-verification --test`

PR #2067, opened 2018-12-05
Localization extension for admin.php

mseed:


andryyy:
Very nice, thank you very much!

PR #2074, opened 2018-12-06
Fix missing string in modal dialog. (Sync jobs)

mseed:


andryyy:
Thanks, go get'em missing translations! :D

PR #2102, opened 2018-12-18
[web] Duplicating DKIM key corrupts private key

patschi:
Missing base64_decode() corrupted private key when duplicating, as `$from_domain_dkim['privkey']` returns the public key base64-encoded.


andryyy:
Thanks, didn't check that. :-(

You tested it, I guess. :-)

patschi:
Yep, I did. I struggeled a few hours to get DKIM signing to work and came across this bug.

andryyy:
Thank you!

PR #2103, opened 2018-12-18
[Web] Add hint to DKIM key import for RSA PKCS#8

patschi:
Adding hint to explicitly provide a RSA Private key in the newer PKCS#8 format, as the webinterface denies the key with a cryptic error message otherwise: `Private key error: error:0EFFF06C:configuration file routines:CRYPTO_internal:no value`. To prevent frustrated users I'd add a simple notice which format is expected. PKCS#8 is also the default format when generating keys directly in the webinterface.

Some interesting resources:
https://stackoverflow.com/questions/20065304/differences-between-begin-rsa-private-key-and-begin-private-key
https://stackoverflow.com/questions/17733536/how-to-convert-a-private-key-to-an-rsa-private-key

This might be relevant/helpful for people generating their keys outside mailcow webinterface, e.g. using standard openssl command on Ubuntu. (or from previous setups, etc)

### DIFF
--- a/data/web/admin.php
+++ b/data/web/admin.php
@@ -333,7 +333,7 @@ $tfa_data = get_tfa();
             <input class="form-control input-sm" name="dkim_selector" value="dkim" required>
           </div>
           <div class="form-group">
-            <label for="private_key_file"><?=$lang['admin']['private_key'];?>:</label>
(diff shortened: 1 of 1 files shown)


andryyy:
Thank you!

PR #2125, opened 2018-12-25
Fix transport map authentication with multiple identical nexthops

mkuron:
@andryyy, previously Mailcow would authenticate as `user:passworduser:password` if two destinations with identical nexthops existed. Please push an updated Postfix image after merging.


andryyy:
Nice catch, thank you!

PR #2140, opened 2018-12-31
Use Horizontal Table Scrollbar Instead

PhoenixPeca:
I think it's better to have it this way:
image

than this:
image

### DIFF
--- a/data/web/css/mailbox.css
+++ b/data/web/css/mailbox.css
@@ -9,7 +9,7 @@ table.footable>tbody>tr.footable-empty>td {
   overflow: visible !important;
 }
 .table-responsive {
-  overflow: visible !important;
+  overflow: auto !important;
 }
 @media screen and (max-width: 767px) {
   .table-responsive {


ntimo:
Yes that would indeed be awesome :)

MAGICCC:
To be honest I am not a big fan of horizontal scrollbars. Instead can we decrease the size of the columns?

andryyy:
Would be better,  but it’s hard while keeping it responsive. Let’s merge this first. :)

Thanks

chriscroome:
The Action column is generally hidden for me on small screens on Firefox, for example I get this on my 12" laptop (unless I full screen Firefox, then I get the overflow):

[Mailcow screenshot](https://wiki.webarch.net/w/images/b/b2/Screenshot_2018-12-31_mailcow-UI.png)

As you can see in the above screenshot (I don't seen to be able to embed images, that is why I linked it) there is no indication that any columns are missing.

I agree a that having a scroll bar is better, however with long lists it still isn't ideal as the horizontal scroll bar is, initially, off the bottom of the screen.

PR #2167, opened 2019-01-10
[Clamav] Remove duplicate chown

amorfo77:
the permissions of file `/var/lib/clamav/whitelist.ign2` are done twice in `bootstrap.sh`


andryyy:
Thanks. I will not rebuild the image, okay? Will be included in the next version of the image. :)

PR #2192, opened 2019-01-17
Fix hard-coded SKIP_CLAMD

K2rool:
The option to disable ClamAv when less than 2.5 GiB memory dose not work as SKIP_CLAMD is hard-coded to no.


andryyy:
Thanks!

PR #2196, opened 2019-01-18
imapsync: Connect to database via socket

wedi:
Make the imapsync script connect to mysql via socket and not tcp like all other services do.

Reasoning: consistency. 🤓  And in my case the database does not listen on tcp; just on the socket.


andryyy:
Haha, yes, good catch!

PR #2200, opened 2019-01-19
[Dovecot] FTS: Exclude autoindex for Junk/Trash

patschi:
Indexing Junk and Trash folder are not really necessary, specially Junk.


andryyy:
I have customers who search Trash and Junk quite a lot. So I prefer to not exclude it.

PR #2212, opened 2019-01-22
Fix transport "destination" input validation to allow email addresses

wedi:
Emails are allowed in [transport](http://www.postfix.org/transport.5.html) and listed as example in the input box but they fail input validation.

### DIFF
--- a/data/web/inc/functions.transports.inc.php
+++ b/data/web/inc/functions.transports.inc.php
@@ -193,7 +193,10 @@ function transport($_action, $_data = null) {
       $username = str_replace(':', '\:', trim($_data['username']));
       $password = str_replace(':', '\:', trim($_data['password']));
       // ".domain" is a valid destination, "..domain" is not
-      if (empty($destination) || (is_valid_domain_name(preg_replace('/^' . preg_quote('.', '/') . '/', '', $destination)) === false && $destination != '*')) {
+      if (empty($destination)
+            || (is_valid_domain_name(preg_replace('/^' . preg_quote('.', '/') . '/', '', $destination)) === false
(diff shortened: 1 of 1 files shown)


andryyy:
I think we need to make sure that all validations still work when we change that. I need to check that. :) Not sure right now.

wedi:
Ah ok. You believe there might be more validations in place for this string?

andryyy:
More validations when editing existing and adding new items. The validation function is a mess.

wedi:
I saw some extensive regex here and there. Is there a minimum PHP version requirement in place? Starting at version 7 there is a `FILTER_VALIDATE_DOMAIN` flag as well.

andryyy:
Yes, but we need to remove some chars first, allow ".domain.tld" for example. But not always, it depends on other entries. Sometimes [domain.tld] is okay, sometimes domain.tld is not.

That's part of the mess I wrote about. :-)

The validation is tricky, the queries in Postfix follow a specific order, we need to check for some combinations.  We cannot allow a specific entry when this or that already exists. It is difficult. We cannot just use checks like "does X already exist?" or "is X a domain?".

wedi:
Are you going to look into this? You are used to the mailcow SQL foo. 🙃 Or would you be willing to merge this to allow email addresses now and add another issue for complete validation with duplicates and everything?

andryyy:
I will test this locally, ok? If it's fine (I think it will not break anything), I will merge it.

wedi:
Editing the transport does work.

andryyy:
It is something Postfix-internal. The order of the queries is the problem. Your PR is fine, I just need to make sure everything Postfix does is still okay. :)

wedi:
unstale :)

PR #2219, opened 2019-01-24
remove dns-search property

smueller18:
@andryyy the property `dns_search` does not make sense. Have a look at https://docs.docker.com/v17.09/engine/userguide/networking/configure-dns/ .
What was the reason for adding this property?


andryyy:
I probably copied 1:1 from dev and didn't check.

PR #2229, opened 2019-01-26
Forward mailcow-ui-domain to the docs

heavygale:
If the domain used for the mailcow UI differs from the hostname and the user tries to download the mobileconfig that's listet in the documentation, he will end up at the login promt. To prevent this, we need the docs to know the domain that's used to access the mailcow ui (as the mailcow-ui-session is only valid for this domain).
This PR adds the variable "ui" to the url and mailcow-docerized-docs can use this new variable when referring back to the mobileconfig file.


PhoenixPeca:
Hi pls check https://github.com/mailcow/mailcow-dockerized-docs/pull/131#issuecomment-457885243

PR #2245, opened 2019-01-29
U2F: Don't check for Firefox, check only if U2F API is implemented

TiiFuchs:
I don't quite get, why there is a check for "Firefox" in this...
If you use Safari and have installed this extension https://github.com/Safari-FIDO-U2F/Safari-FIDO-U2F you can use the U2F authentication quite fine.

I removed the check for Firefox and can now use U2F on macOS with Safari if I install the above mentioned extension.

### DIFF
--- a/data/web/js/u2f-api.js
+++ b/data/web/js/u2f-api.js
@@ -20,10 +20,9 @@
      * Modification:
      * Only continue load this library if window.u2f is not already supplied by the browser.
      */
-    var isFirefox = navigator.userAgent.indexOf('Firefox') !== -1 || navigator.userAgent.indexOf('Gecko/') !== -1;
     var browserImplementsU2f = !!((typeof root.u2f !== 'undefined') && root.u2f.register);
 
-    if (isFirefox && browserImplementsU2f) {
+    if (browserImplementsU2f) {
         root.u2f.isSupported = true;
         return;
     }


andryyy:
Have you tested your change in Firefox?

It was needed a while back in Firefox. Don't know if it is still the case. But please don't remove it _untested_.

andryyy:
https://github.com/bitwarden/web/commit/f7b60febe9e7a58b50c94cbcd49f50bd8b041c76

TiiFuchs:
Hold on... Looking into it...

andryyy:
Looks good in Windows/Firefox.

TiiFuchs:
Okay, one thing to clear up: For Firefox to work you have to enable u2f support in about:config like in https://support.yubico.com/support/solutions/articles/15000017511-enabling-u2f-support-in-mozilla-firefox described.

But if that is done U2F works in firefox as it works in every other browser. If the window.u2f api is there, it works. No check for userAgent needed. It sill does not make any sense to me why this should ever be needed. And if it's needed, then the && has to be wrong... Then it would only work in browsers with "Firefox" or "Gecko" in its user agent sting. (As we experienced it before.)

So yes, it still works in current versions of Chrome, Firefox (with enabled u2f) and Safari (with installed extension). 👍 

Edit: I meant "Yes I tested it in current versions of..."

andryyy:
Yes, the browser check seems weird.

TiiFuchs:
Thanks Andre 🙂 
Always a pleasure! ☺️

PR #2257, opened 2019-01-31
Adds missing values to alias INSERT statement.

guicapanema:
public_comment and private_comments were missing from the VALUES array in PDO prepare statement.


andryyy:
Thank you!

Issue #1661, opened 2018-08-12
TLS v1.3 when?


MAGICCC:
Since TLSv1.3 has a [RFC doc](https://tools.ietf.org/html/rfc8446) now, is there already an ETA to include it along with 1.2 in NGINX?

phenomax:
Following the release of OpenSSL 1.1.1 stable, it will become widely available. As of now, we have to wait or compile an own instance of NginX using openssl-1.1.1-pre.

andryyy:
Yes, we need to wait for OpenSSL 1.1.1 to be available in Alpine (since we use Nginx mainline on Alpine).

ghost:
Sounds great! Will TLS 1.3 be a "web only" thing, or are the messages itself also TLS 1.3 encrypted?
> https://github.com/openssl/openssl/milestone/9

MAGICCC:
Just a quick info.
OpenSSL 1.1.1 is out now. We have to wait until Alpine will release its build and then we have to wait for the Dockerteam to build a new NGinx version with it.
Until then we can read https://wiki.openssl.org/index.php/TLS1.3

MAGICCC:
yay https://pkgs.alpinelinux.org/package/edge/testing/x86_64/openssl1.1-dev
Now we have to wait for the latest push for nginx:mainline-alpine

andryyy:
Cool!

obendev:
This really is going faster than I expected. 😍

**Offtopic**
~@MAGICCC I know its a bit off topic but do you know when apache2 is planning to release TLS 1.3 support? I cant find any "good" information about it.~
~Update: They just released it. https://github.com/apache/httpd/blob/2.4.36/CHANGES~

Knight1:
I know how to build it. We build many containers but i guess it is not worth the effort?

andryyy:
It is not that we don't know how to build it.

Knight1:
That is not what i meant :o 

The problem is that OpenSSL and NGINX needs some time to compile. And Chrome and Firefox does not support it in stable only beta at this time. So do we want it or not?

andryyy:
No, I don't want an non-official image for this, that's what I meant.

MAGICCC:
We need to wait for Alpine 3.9. I am sure the Alpine team will put OpenSSL 1.1.1 in the 3.9 release

obendev:
[...]

PR #2258, opened 2019-02-01
Enable TLSv1.3 finally

Knight1:
With Alpine 3.9 https://pkgs.alpinelinux.org/package/v3.9/main/x86/openssl we got OpenSSL 1.1.1a. 
With https://github.com/docker-library/official-images/pull/5377 it was merged into the Nginx upstream image and thus Nginx was built with it. 

So with the next pull the server uses the new and latest Nginx.

### DIFF
--- a/data/conf/nginx/site.conf
+++ b/data/conf/nginx/site.conf
@@ -14,7 +14,7 @@ server {
 
   ssl_certificate /etc/ssl/mail/cert.pem;
   ssl_certificate_key /etc/ssl/mail/key.pem;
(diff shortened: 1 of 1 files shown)


Knight1:
fixes https://github.com/mailcow/mailcow-dockerized/issues/1661

ntimo:
@andryyy I think we can merge this, what do you think?

andryyy:
Not yet. Wait until I migrated acme-client to acme-tiny with Alpine 3.9 as base. :)

PR #2262, opened 2019-02-03
Update to Alpine 3.9

MAGICCC:
Update Dockercontainers to Alpine 3.9

### DIFF
--- a/data/Dockerfiles/acme/Dockerfile
+++ b/data/Dockerfiles/acme/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.8
+FROM alpine:3.9
 
 LABEL maintainer "Andre Peters <andre.peters@servercow.de>"
 
--- a/data/Dockerfiles/dockerapi/Dockerfile
+++ b/data/Dockerfiles/dockerapi/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.8
+FROM alpine:3.9
 LABEL maintainer "Andre Peters <andre.peters@servercow.de>"
 
 RUN apk add -U --no-cache python2 python-dev py-pip gcc musl-dev tzdata openssl-dev libffi-dev \
--- a/data/Dockerfiles/netfilter/Dockerfile
+++ b/data/Dockerfiles/netfilter/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.8
+FROM alpine:3.9
 LABEL maintainer "Andre Peters <andre.peters@servercow.de>"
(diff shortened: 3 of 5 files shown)


andryyy:
Needs much more testing and - more importantly - the images don't all build with 3.9 at the moment.

andryyy:
Damn, I think we should finally replace acme-client. 😄 HAProxy? Hm.

MAGICCC:
> 
> 
> Needs much more testing and - more importantly - the images don't all build with 3.9 at the moment.

Of course, just as reminder to update to a newer Alpine :p 



> 
> 
> Damn, I think we should finally replace acme-client. 😄 HAProxy? Hm.

Totally acme.sh :)

christianbur:
Why not traefik, with built-in acme client (lego)

andryyy:
Acme tiny it is. 

Full control over all steps. 

I do not want to autocreate sites for mail domains, additional domains etc. in Traefik and care about its kinks. This makes everything more complicted IN MAILCOW. Yes, it is a nice server bla, but not good for us. :)

Acme tiny gives us full control, that’s awesome. No overhead or automations we don’t need

ntimo:
You also have to keep in mind that Traefik stores it’s certs ins json file and we would need to extract the certs from this file in order to use them with Dovecot for example. That’s why I think Nginx with acme tiny is the ideal solution 🙌

christianbur:
Please keep in mind that there are also people who don't use the acme client of Mailcow (e.g. because only acme dns verification behind proxy is possible).

For this it would be cool if mailcow itself would recognize if a new Cert file would be in the directory and then restart the required containers on its own.

andryyy:
That's too much overhead when you can just send a reload/restart to all affected services. Almost all acme clients support post-hooks.

https://mailcow.github.io/mailcow-dockerized-docs/firststeps-rp/#optional-post-hook-script-for-non-mailcow-acme-clients

Or...

```
docker exec $(docker ps -qaf name=postfix-mailcow) postfix reload
docker exec $(docker ps -qaf name=nginx-mailcow) nginx -s reload
docker exec $(docker ps -qaf name=dovecot-mailcow) dovecot reload
```

I should add that to the docs.

christianbur:
Okay, so that's how I'm doing right now.

ntimo:
I think we can close this merge request since @andryyy already upgraded all Alpine images to 3.9

MAGICCC:
What a shame!

andryyy:
😄 
Had to test, sorry.

PR #2280, opened 2019-02-08
Set host timezone for unbound and solr

christianbur:
solr and unbound use UTC time


christianbur:
ipv6nat and memcached-mailcow also have UTC time, but these are external images and not so easy to modify.

andryyy:
Thanks! Will merge and push a new image in the next minutes.

PR #2281, opened 2019-02-08
Create nested directory for backup

aliceinwire:
/home/myhome/dir1/dir2/ is not a directory
create it now?

The expected result is that it create the folder also if is nested.


andryyy:
Mörci.

PR #2286, opened 2019-02-09
Improvement of quarantine notification

AikoAppeldorn:
Improved quarantine template for only 1 new message


andryyy:
Nice, thanks!

PR #2302, opened 2019-02-11
Fix: AutoDiscover issue with Outlook 2019

### LINKED ISSUE #2300: Potential AutoDiscover issue with Outlook 2019.
I am trying to add a mailcow account to Outlook 2019 in exchange mode. It fails, so I've started investigating the issue.

First thing I see that can be a potential issue:
Outlook sends Autodiscoverv1 as discovery protocol:

Logs from nginx:
```
94.103.x.x- - [11/Feb/2019:12:36:50 +0100] "GET /autodiscover/autodiscover.json?Email=my%40emailch&Protocol=Autodiscoverv1&RedirectCount=1 HTTP/1.1" 400 182 "-" "Microsoft Office/16.0 (Windows NT 10.0; Microsoft Outlook 16.0.11231; Pro)"
```

If i try this url from a web browser:

```
{"ErrorCode":"InvalidProtocol","ErrorMessage":"The given protocol value \u0027Autodiscoverv1\u0027 is invalid. Supported values are \u0027ActiveSync,AutodiscoverV1\u0027"}
```

You can see it suggest AutodiscoverV1 (with capital V)

So if I try again the same url just uppercasing the v1 in the protocol, it returns:

```
{"Protocol":"AutodiscoverV1","Url":"https://mail00.whatever.com/Autodiscover/Autodiscover.xml"}
```

Not sure that's why I can't add the account to Outlook 2019, but I think the procotol string should be case insensitive...

extremeshok:
Fixes #2300

### DIFF
--- a/data/web/autodiscover-json.php
+++ b/data/web/autodiscover-json.php
@@ -11,7 +11,7 @@ header('Content-type: application/json');
 if ($_GET['Protocol'] == 'ActiveSync') {
   echo '{"Protocol":"ActiveSync","Url":"' . $autodiscover_config['activesync']['url'] . '"}';
 }
-elseif ($_GET['Protocol'] == 'AutodiscoverV1') {
+elseif (($_GET['Protocol'] == 'AutodiscoverV1') || ($_GET['Protocol'] == 'Autodiscoverv1')) {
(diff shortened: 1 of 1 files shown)


andryyy:
Was about to push it. :) Thanks.

PR #2313, opened 2019-02-12
Fix mailbox garbage collector not working

OpenLarry:
I just realized that the hourly [`maildir_gc.sh`](https://github.com/mailcow/mailcow-dockerized/blob/master/data/Dockerfiles/dovecot/maildir_gc.sh) cron doesn't work.

The file gets cleared during the [envsubst command in `docker-entrypoint.sh`](https://github.com/mailcow/mailcow-dockerized/blob/60d74e3f17d8527b34975f7f8e7e9a820d0de628/data/Dockerfiles/dovecot/docker-entrypoint.sh#L192), because the one redirection operator opens the file for writing before the other starts reading from it.

This solution is inspired by https://serverfault.com/a/547331 :-)


andryyy:
Sigh. Thanks!

PR #2335, opened 2019-02-15
Add a "Virtual/All Mail" folder

jhesketh:
This doesn't appear in LIST and isn't subscribable, but it allows a user
to search across all folders at once (for example, with fts).

This should work as it is, but I'm not sure if it is overly helpful yet.

It would be useful to have an improvement to SOGo to search this folder. For example, what OX do: http://documentation.open-xchange.com/7.8.2/middleware/components/search/crossfolder_fts_in_mail.html

Additionally it appears that K-9 mail (on android), only searches the subject, to, and from fields on server-side search so its use is limited too.

Finally I'm not sure how other clients handle their server-side search. For me, thunderbird isn't using imap search which is something else I want to solve.

Having said all that, some users could find it useful to have an "All Mail" folder if we wanted to make it visible (perhaps via config)

### DIFF
--- a/data/Dockerfiles/dovecot/docker-entrypoint.sh
+++ b/data/Dockerfiles/dovecot/docker-entrypoint.sh
@@ -93,13 +93,13 @@ EOF
 echo -n ${ACL_ANYONE} > /usr/local/etc/dovecot/acl_anyone
 
 if [[ "${SKIP_SOLR}" =~ ^([yY][eE][sS]|[yY])+$ ]]; then
-echo -n 'quota acl zlib listescape mail_crypt mail_crypt_acl mail_log notify' > /usr/local/etc/dovecot/mail_plugins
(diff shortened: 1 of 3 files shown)


extremeshok:
If u want subfolder searching in SOGo ,the feature was added about 2-3years ago

Open sogo, mailview, on the left click the ... by your account name, select search
A blue bar will open, click the cog and select search "Search Subfolders"

sogo_search_sub_folders

extremeshok:
I do like the idea of an "All Mail" virtual folder, as gmail has this

Makes it easy for tracking down an email which could be filed in a subfolder or for special use cases.

mkuron:
Duplicate of #1864. see there why virtual folders can be problematic with some clients. Since you make them invisible, it‘s probably ok, but then I don‘t get how you would use them.

jhesketh:
Right, I was aware of the "search subfolder" option, this was more for other clients. (As a side note, and possibly a usability bug for sogo, it is a little hidden).

However the points in #1864 are very valid. I'm not sure there would be enough use for this but pushed it up just in case. At best it stops the client from having to recurse the directories itself to search (which some clients don't do, or would be a speed improvement in the case of sogo).

PR #2360, opened 2019-02-23
Allow admins to login as email user (without any password)

mhofer117:
**general changes**
- add config ALLOW_ADMIN_EMAIL_LOGIN, default=n
- add SOGoTrustProxyAuthentication = YES to SOGo config to allow authentication with x-webobjects-remote-user header
- add dovecot config to allow the sogo container imap access without a password
- add nginx auth_request configs for adding the x-webobjects-remote-user header
- create sogo-auth.php as auth_request backend + redirector for logins
- modify UI to show SOGo login button on email address list
grafik

**additional details**
- when clicking on the SOGo link, sogo-auth.php will check permissions and set a session variable 'sogo-sso-user' with the e-mail address. after that it will redirect to /SOGo/
- on every http request in SOGo, auth_request will call sogo-auth.php internally
- sogo-auth.php will then check if the 'sogo-sso-user' session variable exists and return the value accordingly so nginx can set the correct header.
- when accessing SOGo without this session variable, SOGo will display the normal login screen

**drawbacks**
- each SOGo request will cause an additional auth_request to the sogo-auth.php, on larger installations this might cause performance problems, this is not the case if you have the feature disabled with ALLOW_ADMIN_EMAIL_LOGIN=n
- SOGo will not display a logout link, to login normally one has to logout from the mailcow UI so the session is destroyed

**security considerations**
- according to my tests, the dovecot passdb argument allow_real_nets should ensure that only the SOGo container can access imap without a password:
  - when SOGo web UI makes imap requests, dovecot will see the SOGo container IP
  - when SOGo proxies Active-Sync requests, the real client IP should be forwarded and dovecot requires a valid password
  - the following (existing) nginx configs should ensure that the real client IP cannot be spoofed from outside of local networks:
```
  set_real_ip_from 10.0.0.0/8;
  set_real_ip_from 172.16.0.0/12;
  set_real_ip_from 192.168.0.0/16;
  set_real_ip_from fc00::/7;
  real_ip_header X-Forwarded-For;
```
- in sogo-auth.php i copied existing code to ensure the admin user is allowed to login as the destination email address, as far as i could see this is correct, but should be double checked.

### DIFF
--- a/data/web/sogo-auth.php
+++ b/data/web/sogo-auth.php
@@ -1,10 +1,9 @@
 <?php
-require_once $_SERVER['DOCUMENT_ROOT'] . '/inc/prerequisites.inc.php';
 
 /**
- * currently disabled: we could add auth_request to ningx sogo_eas.template
- * but this seems to be not required with the postfix allow_real_nets option
- */
+* currently disabled: we could add auth_request to ningx sogo_eas.template
+* but this seems to be not required with the postfix allow_real_nets option
+*/
 /*
 if (substr($_SERVER['HTTP_X_ORIGINAL_URI'], 0, 28) === "/Microsoft-Server-ActiveSync") {
   require_once $_SERVER['DOCUMENT_ROOT'] . '/inc/prerequisites.inc.php';
@@ -38,6 +37,7 @@ if (!$ALLOW_ADMIN_EMAIL_LOGIN) {
   exit;
 }
 elseif (isset($_GET['login'])) {
+  require_once $_SERVER['DOCUMENT_ROOT'] . '/inc/prerequisites.inc.php';
   if (isset($_SESSION['mailcow_cc_role']) && $_SESSION['acl']['login_as'] == "1") {
     $login = html_entity_decode(rawurldecode($_GET["login"]));
     if (filter_var($login, FILTER_VALIDATE_EMAIL)) {
@@ -53,9 +53,10 @@ elseif (isset($_GET['login'])) {
 }
 else {
   // this is an nginx auth_request call, we check for an existing sogo-sso-user session variable
+  session_start();
   $username = "";
   if (isset($_SESSION[$session_variable]) && filter_var($_SESSION[$session_variable], FILTER_VALIDATE_EMAIL)) {
-    $username = $_SESSION[$session_variable];
+      $username = $_SESSION[$session_variable];
   }
   // if username is empty, SOGo will display the normal login form
   header("X-Username: $username");
--- a/data/web/sogo-auth.php
+++ b/data/web/sogo-auth.php
@@ -1,4 +1,5 @@
 <?php
+require_once $_SERVER['DOCUMENT_ROOT'] . '/inc/prerequisites.inc.php';
 
 /**
  * currently disabled: we could add auth_request to ningx sogo_eas.template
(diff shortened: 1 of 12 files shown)


mkuron:
Thanks, this looks interesting. Before merging, we need to check carefully though whether this has any negative security side effects.

data/web/sogo-auth.php
```diff
@@ -1,62 +1,63 @@
 <?php
+require_once $_SERVER['DOCUMENT_ROOT'] . '/inc/prerequisites.inc.php';
```
mhofer117: I specifically added it only on requests with $_GET['login'], so it is not executed on auth_request calls, but I don't know how much cpu time the prerequisites consume, i'll leave the decision to you :)

andryyy:
It's a cool approach!

But should be checked for security indeed.

mhofer117:
Yes I agree, would be nice if a few people can inspect specifically regarding security issues I might have not thought about. Thank You for considering this addition.

data/web/sogo-auth.php
mhofer117: maybe it was not the best idea in the first place to use the same file for redirection and auth_request

andryyy: I should ask before changing your code... sorry.

sriccio:
This is a very useful feature. I'm hot to try it on our mailcow box. Any easy way to push these changes to our running containers ? We need to rebuild them ? I'm not yet 100% used docker :)

mhofer117:
you may do it with git:
```
# git remote set-url origin https://github.com/mhofer117/mailcow-dockerized.git
# git fetch
# git checkout allow-admin-email-login
# ./update.sh
```

then edit your `mailcow.conf` and set:
```
ALLOW_ADMIN_EMAIL_LOGIN=y
```
and run `docker-compose up --build -d`
this will take a while if you haven't built the images from source before


and to switch back:
```
# git remote set-url origin https://github.com/mailcow/mailcow-dockerized.git
# git fetch
# git checkout master
# ./update.sh
# docker-compose pull
# docker-compose up -d
```

if you have any conflicts, try git `reset --hard`

keep in mind that you might have to resolve some issues manually later if you switch back and forth
no guarantee whatsoever, do at your own risk on a live system

sriccio:
Hey @mhofer117, 

Thanks a lot for the howto. I'll check it out and test it on a mailcow install we use daily but that is not critical (and having snapshots).

I'll let you know if we see something wrong using the feature :)

andryyy:
I can add any account of the server as new imap account (edit: without password) to SOGo and see its content.

You'd need to set SOGoMailAuxiliaryUserAccountsEnabled to NO. But I don't think we should change that.

andryyy:
Don't know. That's dangerous. I'm not sure where or if it is possible to force SOGo to use another IMAP account. With our current config, we would allow everyone to add any foreign mailbox of that server. If we remove SOGoMailAuxiliaryUserAccountsEnabled, we disable the possibility to add other accounts. If we disable that option, it removes already added accounts.

We could refuse to start when its value is not NO and/or add another parameter "insecure_sogo_login" bla to make an admin confirm the described behavior.

andryyy:
As a dirty hack, I disallowed editAccount for other accounts than 0.

SOGo now always queries IMAP via IPv4.

https://github.com/mailcow/mailcow-dockerized/commit/57312ad605b1e0b2a996b07cde6280299ae8025c

Keep in mind that this also disables _any other_ account from being added.

mhofer117:
Very good find! This is indeed dangerous and i didn't know about this.

I would suggest force disabling SOGoMailAuxiliaryUserAccountsEnabled if admin login is enabled and add an appropriate warning to the new setting.

Anything else would be too dangerous in multi-user environments.

When I have time again I'll try to find a workaround, but it seems to be a limitation of SOGo for now as the imap trust in dovecot is the suggested way in the docs, even though they don't warn about auxiliary email accounts.

andryyy:
Force-disable is probably the way to go. Not sure about my Nginx args hack. What do you think @mkuron?

mhofer117:
After digging in the sogo source, i found a proper workaround with the x-webobjects-auth-type header and a normal basic auth header + a random master password.
I think this removes most of the attack surface from this PR.
Will test a bit more this evening and create a new PR.

sriccio:
I've tested the feature a bit, seems to work great.

May I suggest that when you click on the sogo link on the e-mail addresses list, it opens it on an unique named tab/window instead of creating a new one at each click ?

This is to prevent admin thinking they can log into multiple accounts in multiple tabs. At the moment, if you impersonate to a mailbox then goes back to the e-mail list and impersonate to another mailbox, it will invalidate the session in the previously opened tab.

Maybe there is a way to have multiple tabs keeping each tab session alive ?

Cheers

mhofer117:
Is it possible to use / re-use a named tab?
As far as i know you can only do that with popup windows.

I think there would be a way to have multiple tabs / sessions at the same time because all authenticated requests to SOGo contain the usernamen in the request url (/SOGo/so/user@domain/...), which we could check and set the headers accordingly.

Maybe I'll give this a shot later, but for me this feature was mainly to allow access to a single mailbox for debugging / support purpose.

sriccio:
Hello. Yes for me too it's for debugging and support. But sometimes you are doing support involving multiple mailboxes, so I thought it would be nice that it supports it :)

PR #2368, opened 2019-02-26
Security Fix/Extension for #2360 - Allow Admin Login

mhofer117:
As mentioned in https://github.com/mailcow/mailcow-dockerized/pull/2360, this should fix most security concerns

There is no documentation mentioning it, but after finding [SOGoProxyAuthenticator.passwordInContext](https://github.com/inverse-inc/sogo/blob/master/SoObjects/SOGo/SOGoProxyAuthenticator.m#L108) I finally found a solution to pass a password to SOGo with proxy authentication. While this is not documented, the code is there since a while so i guess it will not be removed / changed soon.

1. We have to set the following headers which tells sogo to extract the password from the basic auth header:
```
x-webobjects-auth-type: Basic
x-webobjects-remote-user: username
Authorization: base64(username:masterpass)
```
The headers are injected by the nginx reverse proxy and not visible to the user.

2. Because we don't have the users passwords (they are encrypted) we have to generate a random dovecot master password which is valid for all users (but only from the sogo container) and re-generated on each dovecot restart
```
passdb {
  driver = static
  args = allow_real_nets=172.22.1.248/32 password={plain}xyz
}
```

This solves the security issue mentioned here: https://github.com/mailcow/mailcow-dockerized/pull/2360#issuecomment-466812291 and generally removes a lot of the attack surface as there i no empty dovecot master password anymore.

*edit: Additionally, SOGoEncryptionKey is set so stored remote calendar passwords are encrypted during proxy-auth

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 rebuild-images.sh
 data/conf/sogo/sieve.creds
+data/conf/phpfpm/sogo-sso/sogo-sso.pass
 data/conf/dovecot/dovecot-master.passwd
 data/conf/dovecot/dovecot-master.userdb
 mailcow.conf
--- a/data/Dockerfiles/dovecot/docker-entrypoint.sh
+++ b/data/Dockerfiles/dovecot/docker-entrypoint.sh
@@ -118,17 +118,6 @@ default_pass_scheme = SSHA256
 password_query = SELECT password FROM mailbox WHERE active = '1' AND username = '%u' AND domain IN (SELECT domain FROM domain WHERE domain='%d' AND active='1') AND JSON_EXTRACT(attributes, '$.force_pw_update') NOT LIKE '%%1%%'
 EOF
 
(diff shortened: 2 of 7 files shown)


andryyy:
👍 Sounds good, let's test! :)

andryyy:
CalDAV/CardDAV is broken now and always asks for permission.

It will probably be enough to disable the extra params for /SOGo/dav/

andryyy:
Should be fixed with 38911034c3858d3dd55c46673c010fb06f1e4d29 - but I guess there are not-so-ugly implementations for this fix. :-D Quick-and-dirty.

mhofer117:
We might also need to prevent forwarding the special headers on this directive, because according to
http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_pass_request_headers
they are forwarded by default
if you want i can make another PR tomorrow

andryyy:
That would be really, really nice!

PR #2373, opened 2019-02-27
always check basic auth against user database for EAS and SOGo if ALLOW_ADMIN_EMAIL_LOGIN is enabled

mhofer117:
This took a bit more time than expected @andryyy :

After some more testing with your feedback about DAV, i found out that if SOGoTrustProxyAuthentication is enabled, SOGo will always trust the username in the authorization header and not validate the password..!
This is true for CardDAV, CalDAV and even EAS calendar / contact sync...

So this forces us to always validate the credentials in the Authorization header (if set) at the proxy level.

I implemented this by re-adding the code i previously removed and cleaned everything a bit up.
The additional nginx directive for /SOGo/dav could also be removed again because we have to add auth_request to all SOGo calls anyway.

After validating the Authorization header, it is forwarded correctly to SOGo so there should be no unnecessary credential popups in the email clients.

One thing I'd like you to confirm:
- Does check_login($username, $password) === 'user' properly ensure a valid email login?

### DIFF
--- a/data/conf/nginx/site.conf
+++ b/data/conf/nginx/site.conf
@@ -142,7 +142,19 @@ server {
     try_files /autoconfig.php =404;
   }
 
+  # auth_request endpoint if ALLOW_ADMIN_EMAIL_LOGIN is set
+  location /sogo-auth-verify {
+    internal;
+    proxy_set_header  X-Original-URI $request_uri;
+    proxy_set_header  X-Real-IP $remote_addr;
+    proxy_set_header  Host $http_host;
+    proxy_set_header  Content-Length "";
+    proxy_pass        http://127.0.0.1:80/sogo-auth;
+    proxy_pass_request_body off;
+  }
+
   location ^~ /Microsoft-Server-ActiveSync {
+    include /etc/nginx/conf.d/sogo_proxy_auth.active;
     include /etc/nginx/conf.d/sogo_eas.active;
     proxy_connect_timeout 4000;
     proxy_next_upstream timeout error;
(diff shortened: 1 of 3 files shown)


mhofer117:
Oh and one more thing I did not have the time to check:
We have to make sure the fail2ban implementation is catching authentication errors returned by auth_request
I wlll do that tomorrow.

andryyy:
Hi,

Wow, thank you so much for this - again!

    Does check_login($username, $password) === 'user' properly ensure a valid email login?

Yes, it does.

I can update the f2b implemenation.

mhofer117:
That would be great :)

If f2b can block brute force to sogo-auth.php I'd consider this feature to be safe.

If no other issues pop up, it can probably me released then.

andryyy:
I think it looks good. :o

andryyy:
@mkuron - merge?

mkuron:
Sorry, I didn't have time to review the admin-login branch yet. I still consider this a rather dangerous thing because it can introduce security issues on multiple levels.

mhofer117:
What type of security issues are you still concerned about?

Here's the state after the recent updates with help of @andryyy :
- Dovecot receives a random master password which is valid from the SOGo container only and made available to the php-fpm container for admin logins
- In SOGo, SOGoTrustProxyAuthentication is set to YES
- Any request to SOGo will then trust the username in the Authorization _or_ x-webobjects-remote-user header

This means we have to make user/password checks before forwarding these headers to SOGo, thus:
- Any request to SOGo or EAS has nginx auth_request enabled with the new sogo-auth.php
- sogo-auth.php will verify basic authentication if the Authorization header is set (this is true for DAV and EAS calls), otherwise
- sogo-auth.php will set a session variable for the desired mailbox if the mailcow user is permitted to access said mailbox (?user=xy calls), otherwise
- sogo-auth.php will check this session variable on accessing /SOGo and either set the required headers for /SOGo or set them empty so /SOGo will display the normal login page
- failed authentications are watched by netfilter

So both of the trusted proxy headers are always generated by sogo-auth.php and only if credentials are sufficient.

Would be nice if you find the time to review the change and ask if there is something unclear :)

PR #2377, opened 2019-03-01
[Feature] Added Policyd

zkryakgul:
[What is policyd?](https://wiki.policyd.org/)


andryyy:
Thanks, that’s very nice, but why? We have Rspamd which does not only cover everything Policyd does but is even much more flexible.

zkryakgul:
Can we do spesific user based limitaions on rspamd? For example; block some users mail to gmail or allow some users only send mail on local domains.

andryyy:
Yes. There are multiple ways, I think. There already is a map containing mailcow domains you could use to identify local mail. Specific policies could be forced with a settings map containing references to that map. Like “if it’s coming from this and going there, reject”.

extremeshok:
policyd has been superseded by rspamd. Please do NOT add this.

Rather customize rspamd to add the missing features or develop a custom module for it https://rspamd.com/doc/modules/

zkryakgul:
Im little confused because im pretty new on rspamd; for example: how can i do like this;
```
from: testuser@mailcow.email
to: *@gmail.com
verdict: reject
```

or 

```
from: testuser@mailcow.email
to: !not_local_domain_group
verdict: reject
```

andryyy:
grafik

```
authenticated = yes;
priority = 10;
rcpt = "/.*@gmail\.com/";
apply "default" {
  GMAIL_OUTGOING_BL = 9999.0;
}
symbols [
  "GMAIL_OUTGOING_BL"
]
```

PR #2385, opened 2019-03-02
allow multiple concurrent admin logins

mhofer117:
small change to allow multiple admin login sessions and even a normal SOGo login at the same time, see https://github.com/mailcow/mailcow-dockerized/pull/2360#issuecomment-468547605


sriccio:
Thanks Marcel,

I'll test this out ASAP.

Cheers

PR #2388, opened 2019-03-02
Admin login to SOGo

mkuron:
From #2373. i still need to review whether this introduces any security concerns before we merge it.


andryyy:
Thanks. :)

extremeshok:
Any progress on this ?

andryyy:
I think this will be a merger. But let's give @mkuron time to check!

mkuron:
Sorry, I‘ve been too busy lately. Hopefully this weekend.

extremeshok:
I was not sure if we were waiting for something to be coded or bug to be fixed, was going to see if I could do anything to help.

sriccio:
Just a little note for @mhofer117, I've tested the multiple concurrent admin logins (commit https://github.com/mailcow/mailcow-dockerized/pull/2388/commits/6a7b4387ebd54180decc3e3ea6b2c0c9bf44fc09) and it works perfectly :)

andryyy:
It works. The only concern is security at the moment.

mkuron:
I don't understand how this works. The master password is stored in `$_SESSION`. Then nginx makes an internal request to the script to retrieve it. That would only work if nginx somehow copied the session cookie into that internal request?

Also, what does `SOGoEncryptionKey` do? What does SOGo encrypt with it and where is the encrypted text stored?

mhofer117:
Hi @mkuron, thanks for taking the time to review

Yes, the master password is stored in the $_SESSION so we don't need to read the file on every auth_request call.
Yes, nginx auth_request actually has the same context as the normal request, even though it's executed internally. This includes the php session cookie and means php will automatically find the users session on session_start() with this cookie.

[SOGoEncryptionKey](https://sogo.nu/files/docs/SOGoInstallationGuide.html) is used to encrypt remote calendar passwords in case of proxy auth. I'm not sure why this would be important but it's mentioned in the SOGo docs so I figured it might be a sensible option.

and thanks @sriccio for testing

extremeshok:
Could this be merged with the default being disabled.

A notice could be added that this could introduce a security risk, this would allow far more user testing and battle hardening.

mhofer117:
Actually this is already the case.
On update the configuration option will be added with value "n" which disables the feature completely

mkuron:
I think this is ok. Please resolve the merge conflict, update the image versions and merge. We should definitely keep it disabled by default, but it's a nice feature for those who need it. I have triple-checked that it does not have any side effects if is disabled and am reasonably certain it is secure when enabled.

Also, this infrastructure could be used to make the SOGo link in the top bar when a user logs in take them directly to SOGo without having to log in again.

extremeshok:
> Also, this infrastructure could be used to make the SOGo link in the top bar when a user logs in take them directly to SOGo without having to log in again.


Now that would be useful for users ;P

extremeshok:
Can this be merged ?

andryyy:
Let's just create some docs first. :)

mhofer117:
Let me know what type of documentation you'd like to have and if I can help.

andryyy:
Just anything would be fine. :) Perhaps inside Debugging & Troubleshooting, not sure. Just a small doc about what it is and what it does.

I will merge it afterwards.

mhofer117:
Hey @andryyy I finally had enough time to add some documentation. Hope it's what you expected.

andryyy:
@mhofer117 Absolutely, thank you!!

extremeshok:
Thanks to everyone who made this possible

mhofer117:
Thanks for merging :)

andryyy:
Thank YOU! :)

andryyy:
Can you check the open issues about this?

mhofer117:
Yes you can ping me if there are issues. As far as I can see there was only one with the proxy and you aolved it, right?

andryyy:
Indeed, thank you. :)

PR #2393, opened 2019-03-04
Fix BREACH (CVE-2013-3587)

aliceinwire:


andryyy:
I'm still not sure about this. I know it exists, but is a pita to find useful information to check for vulnerability.

We should probably merge this. Mh. @mkuron ?

Braintelligence:
Is there an alternative compression algorithm that can be used safely?

mkuron:
I‘ve also looked at this but couldn’t make up my mind. HTTP compression still seems to be very common. It is only exploitable in cross-origin requests as far as I can see, so Mailcow‘s attack surface is quite limited.

evilstiefel:
Disabling gzip completely is a bad idea. A good mitigration is to declare all cookies as same-site only cookies as part of the Set-Cookie header

`Set-Cookie: key=value; SameSite=strict`

Alternatively, disabling compression for cross-site requests is an option, but that would be more involved than just adding a flag to cookies when setting them.

The SameSite-flag, by the way, is now supported by almost all major browsers:

[CanIUse Reference](https://caniuse.com/#search=samesite)

PR #2405, opened 2019-03-07
Moved date definition to allow providing custom values

Arteneko:
I had the need for a custom backup script to wrap the mailcow one, but I needed control over the backup directory name.

This allows the user to define a custom `DATE` environment variable to replace the default `DATE` value.


hachre (approved):
looks good to me, maybe the env variable should  have a mailcow-specific name though?

Arteneko:
I think so, but that would require a bit more change.

If the change occurs, I think that, instead of setting the date, we should be able to set the entire directory name though.

PR #2413, opened 2019-03-09
Fix SOGo default access comment 

fm:
While setting up a dev environment of mailcow, I came across a stray comment that needed some fixing. This is a very simple PR, but thought it would be helpful to fix!


andryyy:
Oops, thank you!

PR #2417, opened 2019-03-10
fix solr query ngram

### LINKED ISSUE #2412: Solr, Dovecot FTS partial word match searching
I'm not getting as much search results as required, only full words will be found, but not partials of a longer word, e.g. if I search for snow, I don't get snowball, even if I search for snowba, I don't get snowball from the fulltext engine of dovecot, no matter if I use SOGo or Rainloop to do an IMAP search

I did do some research and it seems I found a possible way to do it, I'd welcome a way to activate partial word match keyword searching in Solr, as described in this article here:
https://hostedapachesolr.com/support/partial-word-match-solr-edgengramfilterfactory

**Additional context**
Unfortunately I have not found a way to reconfigure Solr, as it looks like it's all done in the entrypoint.sh script and I didn't want to mess around with the script, maybe you could offer some help, how I could activate partial word searching

Thank you ever so much

soulsymphonies:
fixes https://github.com/mailcow/mailcow-dockerized/issues/2412

### DIFF
--- a/data/conf/solr/solr-schema-7.7.0.xml
+++ b/data/conf/solr/solr-schema-7.7.0.xml
@@ -18,7 +18,6 @@
     </analyzer>
     <analyzer type="query">
       <tokenizer class="solr.StandardTokenizerFactory"/>
-      <filter class="solr.EdgeNGramFilterFactory" minGramSize="3" maxGramSize="20"/>
       <filter class="solr.SynonymGraphFilterFactory" expand="true" ignoreCase="true" synonyms="synonyms.txt"/>
       <filter class="solr.FlattenGraphFilterFactory"/>
       <filter class="solr.StopFilterFactory" words="stopwords.txt" ignoreCase="true"/>


andryyy:
Thank you, I just wanted to be sure I don't destroy it (yet again). 😄

soulsymphonies:
it's alright, I love to make contributions anyway :)

andryyy:
Doesn't really work. It shows no results at all for some test installations of mine now.

I have seen this happening before when I removed Ngram from query.

Please check this or we need to revert.

//Edit: This probably happens after first commits occur.

andryyy:
If it's not fixed with the next change, I will drop Solr. So tired of such a ram-bloating non-working crap. If it is too hard to run a stupid index with Solr, we need to go for something else. I don't care about it at this point.

soulsymphonies:
I will test it on another installation, for my main installation it still works with these settings, kind of strange.

Edit: Okay, somehow it stopped working, maybe I can find out why this is.

soulsymphonies:
I haven't much time though at the moment, probably it would be better to temporarily put in the line again and I will do some research and testing in the next days, if that is desired

andryyy:
I already hate Solr.

But it still works fine for you? Can you test a compose down and up -d? Still possible it is problematic I always recreate the core at the very first container start.

soulsymphonies:
I can't say anything negative about solr, maybe it was just my mistake, I'm sorry for the inconvenience then...

My further testing:
- after docker-compose down and up -d it's all gone, no search results at all
-  If I revert the change and put the line 

`<filter class="solr.EdgeNGramFilterFactory" minGramSize="3" maxGramSize="20"/>`

back in and do a rescan, now I get search results again, but this time without all the extra search results of partials.

```
<analyzer type="query">
      <tokenizer class="solr.StandardTokenizerFactory"/>
      <filter class="solr.EdgeNGramFilterFactory" minGramSize="3" maxGramSize="20"/>
      <filter class="solr.SynonymGraphFilterFactory" expand="true" ignoreCase="true" synonyms="synonyms.txt"/>
```

I'm completely confused now, maybe there was just something wrong with my solr indexes the last time I did the investigation, maybe that's the reason I did get these extra results 

I'm sorry, probably it would be best to revert to the previous state and put back the line into the query section and to monitor it for the next few days.

soulsymphonies:
sorry for having messed things up, I'v created a PR https://github.com/mailcow/mailcow-dockerized/pull/2431/ that reverts the change

PR #2421, opened 2019-03-10
Enable MariaDB SSL

ntimo:
Hello together,

I went a head and enabled SSL for MariaDB by doing the following:
- Test if the certificates exists if not generate it
- Test if the certificates is expired if yes delete it and regenerate it
- Mount the entrypoint inside of the docker container
- Updated the configuration of MariaDB to use the certificates

### DIFF
--- a/data/assets/mysql/docker-entrypoint.sh
+++ b/data/assets/mysql/docker-entrypoint.sh
@@ -6,6 +6,17 @@ if [ ! -f /var/lib/mysql/sql.key ]; then
   openssl req -x509 -sha256 -newkey rsa:2048 -keyout /var/lib/mysql/sql.key -out /var/lib/mysql/sql.crt -days 3650 -nodes -subj '/CN=mysql'
 fi
 
+if [ -f /var/lib/mysql/sql.key ]; then
+  if openssl x509 -checkend 86400 -noout -in /var/lib/mysql/sql.crt
+  then
+    echo "Certificate is good for another day!"
+  else
+    rm /var/lib/mysql/sql.key
+    rm /var/lib/mysql/sql.crt
+    openssl req -x509 -sha256 -newkey rsa:2048 -keyout /var/lib/mysql/sql.key -out /var/lib/mysql/sql.crt -days 3650 -nodes -subj '/CN=mysql'
+  fi
+fi
+
 # if command starts with an option, prepend mysqld
 if [ "${1:0:1}" = '-' ]; then
 	set -- mysqld "$@"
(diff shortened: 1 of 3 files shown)


mkuron:
Since we are using MariaDB over a unix socket, we don‘t really need SSL. Also, it‘s useless if the certificate isn‘t being validated.
Further issues:
- why does the entrypoint now need to be mounted but didn‘t need to be before?
- you only check for certificate validity when the container is restarted, but the certificate could expire while the container is running

ntimo:
@mkuron 
Yes that's true, but you could also use the MariaDB server from Mailcow, to host other databases for example if you are using Gitea. The entrypoint needs to be mounted because I customised the entrypoint sh to create the certificate this is not part of the [normal entrypoint ](https://github.com/docker-library/mariadb/blob/master/docker-entrypoint.sh). 

I think checking if the certificate has expired via the entrypoint at startup is okay since the certificate has a validity of 9.993155 years.

mkuron:
Then you should probably renew the certificate if it‘s got less than five years left. It‘s unlikely that anyone will use Mailcow for that long without updates/reinstall/etc., so just to be safe. After renewal, the clients won‘t trust the certificate anymore, so you might just drop the renewal mechanism entirely because it‘ll need manual work anyway.

Regarding the entrypoint: could you call the original entrypoint from your custom one? I see that it has already diverged a bit from the official entrypoint.

ntimo:
@mkuron I will update the entrypoint to be the same as the original one. I just updated the one @andryyy added some time last year. 
And I will also update the check so that the cert gets renewed if its past 5 years.

mkuron:
> I will update the entrypoint to be the same as the original one.

That does not solve the problem. They will diverge again.

andryyy:
I stopped using a modified entry point because it seemed wrong/dangerous to diverge from the original script.

There are images with this functionality available. But unfortunately not for MariaDB, iirc. The official MySQL does this, I think.

Don’t know if it’s necessary though. It would probably be if we used TCP.

mkuron:
Even TCP on localhost doesn't need SSL, so it's only useful if you're using Mailcow's database for some external service. And for that, you might as well run another MariaDB instance (it doesn't take that much memory) or put stunnel in front of Mailcow's to add the encryption externally.

ntimo:
Yes that is true tcp on localhost does not need ssl. If doing it this way is not an option, we can close this PR.

extremeshok:
Im against this, 
Custom entry scripts and  mariadb docker do not mix. If u need to run a sql database for other uses, fire up another docker container for it.

There is zero benefit to this, as the mariadb ssl connection would never be used by mailcow containers.

mitchellurgero:
I would have to agree with @extremeshok. I do not see any good side to this. Plus The MariaDB container for mailcow should really only be used for mailcow (or apps that it has support for like Nextcloud/gitea) but not for an external service. 

If we are voting on anything, I vote no for this.

PR #2422, opened 2019-03-11
Upgraded MariaDB to version 10.3.x

ntimo:


MAGICCC:
@andryyy We could use `image: mariadb:10` tho

ntimo:
@MAGICCC that would probably be the ideal idea :)

mkuron:
I think there might have been situations in the past where modifications to Mailcow or its schema were required for compatibility with new minor versions of MariaDB. That is why it is hard-coded to a specific 10.x release.

ntimo:
@mkuron Then I would probably set the version to 10.3 :)

andryyy:
This requires a lot of testing. I usually upgrade the versions once I found time to test upgrades on many different setups.
Last time I tried it broke some of my cows.

ntimo:
I tested it on my backup mx server and it seemed to work fine. But I will test it a bit more with fresh installations today evening. And let you know if it works or not :)

andryyy:
Fresh installations are not interesting.

Edit: Just tested on a single VM and already run into event scheduler errors. This is much to dangerous for a PR.

ntimo:
I am going to close it.

andryyy:
I actually found a bug in the entrypoint script by checking this... thanks!

extremeshok:
Headsup 10.2 to 10.3 requires manual intervention. ie running of mysql_upgrade

https://mariadb.com/kb/en/library/upgrading-from-mariadb-102-to-mariadb-103/

feldsam:
What is benefit to have newer mysql server? I think there is no one, if we don't run in some bug with mysql.

extremeshok:
Too many to mention
https://mariadb.com/kb/en/library/changes-improvements-in-mariadb-103/

mkuron:
> What is benefit to have newer mysql server?

Older versions eventually go out of support. So we have to keep up to date. There is no urgent need to go to 10.3 immediately, but we‘ll do it at some point in the near future.

PR #2424, opened 2019-03-11
Fix rejected mails not being quarantized properly if they are tagged

OpenLarry:
Mails sent to mailbox+something@example.com that got rejected by rspamd do not show up in the quarantine, because the [alias resolution in pipe.php](https://github.com/OpenLarry/mailcow-dockerized/blob/40a826a3471db61349eb64f58dd35bdddd0ae19e/data/conf/rspamd/meta_exporter/pipe.php#L122-L134) doesn't find any mailboxes.

This PR solves the issue by removing the tag from the local part of the address before resolving aliases.

### DIFF
--- a/data/conf/rspamd/meta_exporter/pipe.php
+++ b/data/conf/rspamd/meta_exporter/pipe.php
@@ -84,6 +84,9 @@ $rcpt_final_mailboxes = array();
 
 // Loop through all rcpts
 foreach (json_decode($rcpts, true) as $rcpt) {
+  // Remove tag
+  $rcpt = preg_replace('/^(.*?)\+.*(@.*)$/', '$1$2', $rcpt);
+  
   // Break rcpt into local part and domain part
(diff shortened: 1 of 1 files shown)


andryyy:
Good catch.
lgtm, what do you think, @mkuron ?

It will also work on user+tag+bla@domain.tld. I cannot think of a case where this check fails.

mkuron:
The first `.*`needs to be replaced with `[\+]*`, otherwise multiple `+` may run into an issue with greedy vs. lazy regex evaluation where not all tags are captured.

OpenLarry:
You probably mean `[^+]*` instead of `[\+]*`?

But since the first `.*?` is evaluated lazy, the regex should catch the first occurring + fine:
https://regex101.com/r/X6uls0/2

But I can change the PR if you want to make it more clear.

mkuron:
Ah, missed the `?`, that takes care of it.

Issue #2333, opened 2019-02-15
FTS is case-sensitive


jhesketh:
Full text search is case-sensitive. Generally searches are expected to be case insensitive, so I would consider this a bug. I can't find an option to change it though.

extremeshok:
correction, might be a bug, as the lowercase modifiers are specified in the 
function solr_config() {

jhesketh:
Right, this is a feature-request then :-).

Would you mind sharing your solr-schema.xml?

Also I couldn't see where mailcow is placing the schema. Does it not need one?

extremeshok:
@jhesketh https://github.com/mailcow/mailcow-dockerized/blob/master/data/Dockerfiles/solr/docker-entrypoint.sh is where its generated

This is the dovecot recommended config
https://wiki2.dovecot.org/Plugins/FTS/Solr?action=AttachFile&do=view&target=solr-7.x-schema.xml

jhesketh:
Ah neat, thanks (I was grepping for schema.xml!).

At a skim it appears to be attempting to set things to be case-insensitive, so perhaps this is a bug. (I've only skimmed it very quickly though as it's very late here and I've run out of time sorry)

extremeshok:
Looks the same to me, hence my corrected comment ;P

To help with testing, was your search for the subject, sender or body ?

jhesketh:
I tested subject and body. I was fiddling with other things (search across all-folders) so it wasn't very thorough and probably needs confirming.

extremeshok:
confirmed, bug on my side too. 
@andryyy maybe we should just include the schema.xml and not configure via the json curl ?
I know the json is better for portability to new versions of solr

andryyy:
We just need to configure it. :) Shouldn't be too hard.

andryyy:
text_en_splitting seems to be a default field type. I will try it.

Seems to work. But it also looks like a rescan is needed.

andryyy:
Was a bit more tricky... will push the change this evening.

extremeshok:
Assume we need to flush all the solr data and let it reindex.

andryyy:
Try without first. It should also work with fts rescan.

andryyy:
A complete flush of data is not needed. Worst case is a rescan.

jhesketh:
Thanks for the fix. It seems to be working now :-)
I had to do a rescan first though (`docker-compose exec dovecot-mailcow doveadm fts rescan -A`)

[...]

PR #2431, opened 2019-03-12
revert solr config change

soulsymphonies:
reverts https://github.com/mailcow/mailcow-dockerized/pull/2417

no more search results shown after removing EdgeNGramFilterFactory from query part of solr scheme,
this reverts the change and puts EdgeNGramFilterFactory back

### DIFF
--- a/data/conf/solr/solr-schema-7.7.0.xml
+++ b/data/conf/solr/solr-schema-7.7.0.xml
@@ -18,6 +18,7 @@
     </analyzer>
     <analyzer type="query">
       <tokenizer class="solr.StandardTokenizerFactory"/>
+      <filter class="solr.EdgeNGramFilterFactory" minGramSize="3" maxGramSize="20"/>
       <filter class="solr.SynonymGraphFilterFactory" expand="true" ignoreCase="true" synonyms="synonyms.txt"/>
       <filter class="solr.FlattenGraphFilterFactory"/>
       <filter class="solr.StopFilterFactory" words="stopwords.txt" ignoreCase="true"/>


andryyy:
But it’s strange. Does this only happen on down and up? Then it could also be an issue with the entrypoint.

soulsymphonies:
I have to monitor it over the next days, it's running now with the change reverted on my installation
so the EdgeNGramFilterFactory in my installation is back in the query part of the solr schema

extremeshok:
The revert needs to be done, as the removal of ngram killed search

hachre:
I've updated to the latest commit and am reindexing currently. Will try this out in about 30 minutes.

_Edit:_ It's taking longer than expected to reindex. Still waiting!

andryyy:
Did it happen to you without a down and up in the meantime? @extremeshok

hachre:
I tried it out now and can report: After freshly indexing it worked perfectly. Both this as well as the upper and lower-case searching discussed in #2333 

However: After running `docker-compose down && docker-compose up -d`, searching stopped working entirely, just like @extremeshok reported. All searches return 0 results.

soulsymphonies:
I have tested in the meantime **with revert in place, so EdgeNGramFilterFactory present in the query section again**, it seems that it works so far even after a docker-compose down && docker-compose up -d

but I have to monitor if the strange search result behavior will occur again, where I searched for vitamin and it brought results like vital, vitality, etc.

hachre:
I've noticed this being displayed after `docker-compose down && docker-compose up -d`, while the Solr volume actually still uses the usual 1.5 GB that it should for my server (attained via `du -hs /var/lib/docker/volumes/mailcowdockerized_solr-vol-1`).

image

soulsymphonies:
okay, tested again, even with the revert, after  `docker-compose down && docker-compose up -d`
I get 0 search results, 

first I though it would work, but these were just cached results from SOGo


Edit:
grafik

It seems that Solr is loosing all its indices after a container restart
normally I have around 6000 mails that are indexed

Braintelligence:
@soulsymphonies So you need to reindex after each container restart but then it works?

soulsymphonies:
yes no matter if I use EdgeNGramFilterFactory in query or if I remove EdgeNGramFilterFactory from query, after a reindex I can search

after a container restart the index is gone and a reindex needs to be done...

andryyy:
Found the bug, give me a few minutes, please.

soulsymphonies:
but if I add add the EdgeNGramFilterFactory in query, I get the strange search behavior, where vitamin returns vital, vitality and all words that start with vit

so actually my first PR fixed the strange search behavior...

Edt: a revert of the config change would bring back the strange search behavior

soulsymphonies:
I close the pull request now, for a revert of the config change would bring back the strange search behavior again

the bug seems to be completely independent of the config change anyway,

andryyy:
My whole image was broken, it was not your fault.

Solr is still shitty, but it I was shitty, too.

Edit: Ok, Solr is not shitty. I am just too stupid.

extremeshok:
@andryyy I appreciate your effort with this.

andryyy:
Thank you! :) I hope it is finally fixed.

I forward it to my wife for 💄 and such. 😄

soulsymphonies:
@andryyy: I, too, appreciate your efforts, your service makes a difference in so many lives, no matter if you know it or not, I consider you a jewel 🥇 

And mastering so many different technologies and services and then offering it to others, that means a lot

How did you fix it then?

Braintelligence:
Is there any change on the GitHub repo that reflects what had to be done to get it working?

andryyy:
There were multiple issues. The biggest issue was, that I forgot to change a path in docker-compose.yml, so the data was never persistent (sigh).

We also need to revert to bootstrapping Solr when building it. Solr wants to delete the data directory when creating the core. This can only happen at build time, because you cannot delete a mounted directory. This worked previously, because the volume path was broken, as I mentioned before.

I'm not sure how relevant it is to use the correct solrconfig.xml and schema.xml when bootstrapping the core, but we now do it anyways (copying them to the default configset).
Before, we just copied the files after core creation.

soulsymphonies:
just wanted to say thank you again, everything regarding search and search indexing works now like a breeze 👍

andryyy:
Thank you, too! :)

hachre:
With the latest changes (I suspect the ngram change), my index has grown from ~1.5 GB to ~ 7 GB. Is this expected? Did anyone else experience this?

soulsymphonies:
the index definitely grows and it depends on how many emails you have and how much text they have in them.

The EdgeEngramFilterFactory splits the words in partials in this case with a minimum of 3 letters and 20 letters, it then adds those partials to the indexes, so you can now search not only for complete words but for partial parts of them, however this special filter only creates partials at the beginning of the words.

"I truly love life"

first is tokenized in "I" "truly" "love" "life"

and then partials are built

"I" "tru" "trul" "truly" "lov" "love" "lif" "life"

which obviously results in a bigger index but a far better search experience.

NGramFilterFactory would store partials in the middle of the word, too, like
"truly" -> "tru" "rul" "uly" "trul" "ruly" "truly"
which would make the index even bigger, but is not really practically as it would literally blow the index :)

I hope this little explanation helps

hachre:
Thank you!

PR #2432, opened 2019-03-12
Quarantine - Enhanced JS + Show btn fix event binding

feldsam:
There was problem with show content btn, when you use table search. I use event delegation, by binding event on body, so there is no need to rebind event on table changes.

I make some enhancements in code also.


andryyy:
Thanks!

feldsam:
oh, this was quick :). I thanks for this software!

PR #2433, opened 2019-03-13
Readded Travis CI

ntimo:
Hello together,
I readded Travis CI support. Every container is being build in its own stage this should improve build times. I also added a  check to check if the container already exists in the Docker Hub if not the container gets build and pushed. This works by graping the container tag from the compose file and checking it against the Docker Hub.


.travis.yml
```diff
@@ -1,16 +1,95 @@
...
-  - secure: fWzZisT6nGDNL4lf6tXB07eFG2drgBakHxzdF/NFVvzuP861RFR6omuL+ED0PgXrEHDJBxaBLv52je8irmUXrAH1CNr7T8DWiZo/h5h609Uzr+38T1NnIu4krL0Wo6/CDwlLKnzqTq9yBIZLQSHVJmo8AOpo1JPIi2ajodqj9ZfmAxDQTQl+G6zvQjtqIkYHsHY7A44Rto0f14ykn7w2S82Jn6Ry89VNI5V1WEO3sMpM/XekNP/HokNcRIuntL/0+kuLvTJ5akGoTjBQxSnSW95opzPeGky74HRU2obExJYqKvF0VfVJRNAqejwjIiFIbbjqV0Sk5391kFuhuBErQQDM1bOHGdxZ41HsJH29qNWIl7C33Yl10qERoqecgsJ1N/bS2ZEmWqm/zQh5GClCXPvYmzEqMYsMGM3vjbKdjDlc1Wh2w/eFclsXN9LSXh1mc35rtj46frcT6e5Kof87AIfC9hTgDvk9kAsyjaHMkSHSZthbZXCIcsD8qriNm5UqfFBYD79mPIP1S2YMQ2jscCsjHOZgYVrcm0kzDF21J1w6H0Lo7d1jw37LYlegBdtLQ9gYgqY2D5m+nxWuVoD5FZmpR+5JGtK+ootyLFF8aiFoHXd4op1JCxRLjgkmnZKXzw3kTQSpE7oa7CgzchtQmK2nqcqla1b5Qk7ilVcjooo=
+    - DOCKER_COMPOSE_VERSION=1.23.2
```
mkuron: You don't use this variable anywhere

mkuron (changes requested):

andryyy:
Michaels version worked quite well. We only need a reliable check if an image exists.

.travis.yml
ntimo: Well these where in the original Travis CI configuration file. I am going to remove them, since they aren't used anywhere 👍.

I removed the global variables completely since they are not used. I wanted to use the DOCKER_COMPOSE_VERSION var to set a static version of docker-compose. But then thought that would be too much and removed it. Sorry I forgot to remove those :) Thanks for the heads up.

ntimo:
@andryyy the check is super reliable, it greps the image version from docker-compose file and then checks the Docker Hub API to see if the image tag exists. And if not the image gets build and pushed.

PR #2443, opened 2019-03-16
Add ECDSA support

### LINKED ISSUE #2426: Add ECDSA certificates
Since [dovecot](https://wiki.dovecot.org/SSL/DovecotConfiguration#Multiple_SSL_certificates), [postfix](http://www.postfix.org/TLS_README.html), [nginx](http://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_certificate) and openssl all support dual-stack ECDSA and RSA certificates for quite some time, I would love to see mailcow also generating ECC certificates.

ECC certificates use shorter keys and require less computing power providing the same security as RSA. Because RSA certificates can be served as well, there should be no compatibility issues.

Acme-tiny seems to be capable of handling that, so the renewal script would have to be extended and dovecot's, postfix' and nginx' config files slightly changed.

I would like to support mailcow by creating a pull request for that, if you are interested. :-)

There is only one question that comes into my mind, which is still open:
* Since some users are probably using their own certificates instead of letsencrypt ones, ECDSA certificates should be optional and there must be the possibility to disable this functionality.
* Enabling / Disabling the feature requires the config files to be rewritten / adjusted.
* For dovecot and nginx that seems to be quite easy, as there already exists some logic that replaces env variables on start up. The necessary tls config lines could be placed into a separate file that gets included into the main config.
* But postfix doesn't have such an import mechanism as far as I can see and editing the main.cf seems like a bad idea to me. Is there another solution for that?

What do you think of the idea? :-)

OpenLarry:
I added the ECDSA certificate support like we have discussed in #2426 .

I tried to modify the bash script as less as possible to reduce the time for code review/testing and I tried to use the same coding style like before. However, I decided to move some code into separate functions to avoid code duplication.

A few notes:
* I did not add a separate option (at least not yet) to disable ECDSA certificates when using the integrated Letsencrypt client. I don't think that it makes sense to disable ECDSA, because there should be no drawback when it stays enabled for everyone?
* If SKIP_LETSENCRYPT=y (or the installed certificate isn't issued by Letsencrypt) and there is no custom ECDSA cert, the script will add a symlink from the ECDSA certificate to the RSA certificate.
* ECDSA and RSA certificates are checked and renewed independently from each other, so it's likely that they have different validity periods after migration.
* I've combined [this if-statement](https://github.com/mailcow/mailcow-dockerized/blob/d449984a66d8e6f5b362988581963229337c619b/data/Dockerfiles/acme/docker-entrypoint.sh#L143) with the statement above. I think it's a bug to continue the script if there isn't a certificate, but we have an old one that has non-matching hashes. (However it should happen rarely - or even never?)

I've tested migration (using custom and Letsencrypt certificates) and fresh installations. I've checked all services using [testssl.sh](https://github.com/drwetter/testssl.sh) and it reports two certificates or one certificate depending on whether an ECDSA certificate exists or not. Looks good to me, so I'm looking forward to feedback! :-)

Of course I can add a few lines to the documentation as well.

Closes #2426


andryyy:
I have not looked into it at all. What about TLSA?
Any changes when updating?

I

OpenLarry:
Damn... yes.

Didn't think about that yet because I don't use DNSSEC.
A second certificate also requires a second TLSA record for each offered service.
The automatic generation of DNS records in Mailcow's UI could probably be extended easily to show the required records for both certificates.

But that makes migration from an RSA-only setup indeed more complex for users that are already using DNSSEC.

Any ideas how to solve that easily?
* Check if TLSA records are present and disable the ECDSA certificate if it's not listed?
* Show a warning during the update process and let the user decide (and add an option to disable the ECDSA certificates in mailcow.conf)?
* Combination of both?

andryyy:
And that’s where hacky starts and delivery may fail.

mkuron:
That means we need a configuration option where users explicitly have to turn on ECDSA. I was hoping to avoid that (we have enough such options already), but there is no way around it. I guess we could turn it on by default on new installs though (from generate_config.sh).

docker-compose.yml
```diff
@@ -308,6 +308,7 @@ services:
         - DBUSER=${DBUSER}
         - DBPASS=${DBPASS}
         - SKIP_LETS_ENCRYPT=${SKIP_LETS_ENCRYPT:-n}
+        - SKIP_ECDSA_CERT=${SKIP_ECDSA_CERT:-n}
```
mkuron: This has to default to `y` since not everybody uses update.sh.

mkuron (changes requested):

docker-compose.yml
OpenLarry: You're right.

andryyy:
Can we get docs for this? :)

OpenLarry:
Sure! :-) If the PR is ready to get merged and the functionality is more or less complete, then I'll add something about it to the documentation.

andryyy:
Looks complete. :) Do you think this is ready to be merged?

A few words in the docs are enough.

OpenLarry:
I've added a few words about it to documentation and I updated the chapter slightly. I hope it's all right. :-)

I tried to test my code with every scenario that came into my mind. So I hope it works. 😃

OpenLarry:
Is anything still missing here?
Do you need help? :-)

patschi:
@andryyy Anything new in terms of merging this?

Knight1:
I wonder what happens with all the Software when they load the same certificate twice.

OpenLarry:
I've already checked the services implementation and documentation [here](https://github.com/mailcow/mailcow-dockerized/issues/2426#issuecomment-472438292).

Since it's documented OpenSSL functionality that loading the same certificate twice actually does nothing and postfix/dovecot/nginx don't care about the certificate handling/parsing itself, it should (and does!) work flawlessly in my opinion. :-)

patschi:
What's the exact state with this PR, @OpenLarry? Now as #2509 was merged in the past, I'd guess that most of the code is outdated and not compatible anymore. Any input from @mhofer117 for this?

mhofer117:
I had this included in #2509 initially because I thought ECDSA would be merged first. Then removed it so #2509 could be merged. However my changes are still ready here: https://github.com/OpenLarry/mailcow-dockerized/pull/1

Personally I have no need for ECDSA certs so i'm not maintaining it, but that PR should solve almost all conflicts

OpenLarry:
I've already merged the mentioned PR by mhofer117 locally and it seems to work. But I wasn't able to test everything yet.

I will try to do that in the next two days and update this PR afterwards. :-)

OpenLarry:
Sorry for the late reply, had a lot of stuff to do during the last days...

I've merged the PR by @mhofer117 containing the ECDSA adjustments for the SNI part and it looks good to me. But I haven't tested the SNI part yet, because I have no need for that. Maybe someone else wants to try that out and verify that everything works properly?

MAGICCC:
Just saw your rebase on master, possible to change the base branch to staging and rerebase? :p

OpenLarry:
Sure! I will do that in the coming days and also clean up this branch. :-)

SecT0uch:
Using ECDSA with last acme.sh deploy hook. Really need that to be merged. 

Any news ?

christianbur:
@SecT0uch 
Do you mean [this](https://github.com/acmesh-official/acme.sh/blob/master/deploy/mailcow.sh)? 
I added the ECDSA support over a year ago.

SecT0uch:
@christianbur 
No, I mean mailcow recognizing `ecdsa-cert.pem` and `ecdsa-key.pem` intruduced with [that commit](https://github.com/acmesh-official/acme.sh/commit/2bc627970ed38fbccd7316e7e4a868eb33cff94c)

I had a previous version of the deploy script that was working well.

Edit: I have mailcow up-to-date and don't want to play with symlinks in docker

OpenLarry:
I've squashed all messy commits together and rebased it on the current staging branch. Seems to work (and is already working for me for years).

andryyy:
Yes, keep in mind there are like 47838374 different setups and thinks will get dirty. We will need to be prepared. Thank you, Aaron! :)

SecT0uch:
What's the status on that PR ?

christianbur:
@SecT0uch 
Why did you break my acme.sh deploy plugin? -> https://github.com/acmesh-official/acme.sh/pull/4170

If you want to store ECDSA certificates in the (RSA) cert files (cert.pem, key.pem), this would have been possible simply with symlinks. With your change in acme.sh, you destroyed my plugin and also my mail setup, because DANE/TLSA was broken.

@SecT0uch  -> Please undo the change as soon as possible. 


Create symlink in directory ./mailcow-dockerized/data/assets/ssl
```
ln -s ecdsa-cert.pem cert.pem
ln -s ecdsa-key.pem key.pem
```

```
$ ./mailcow-dockerized/data/assets/ssl# ls -al 
lrwxrwxrwx 1 root root 10 Aug 24 12:11 cert.pem -> ecdsa-cert.pem
lrwxrwx 1 root root 9 Aug 24 12:11 key.pem -> ecdsa-key.pem
-rw-r--r-- 1 root root 5934 Aug 24 11:54 ecdsa-cert.pem
-rw-r--r-- 1 root root 3247 Aug 24 11:54 ecdsa-key.pem
```

SecT0uch:
Because you are including custom changes in a public repository that breaks a default mailcow installation.

You broke my installation in the first place with your changes on the acme plugin.
Since clearly this commit won't be merged, I don't see the point in breaking everyone else's installation.

If you want to keep using the custom changes you made on mailcow, you should fork the acme plugin and make your changes on your side.

christianbur:
The current Mailcow standard is as follows 

- cert.pem, key.pem -> Mailcow rsa cert files
- ecdsa-cert.pem, ecdsa-key.pem -> Mailcow ecdsa cert files

With your change to the acem.sh mailcow deploy plugin, ecdsa certs are now also written to the standard cert files (cert.pem, key.pem). If you want to run Mailcow with ecdsa certs only, this is ok, but it is not the default, so please use a symlink and fix the Mailcow deploy plugins again.

christianbur:
further discussion about acme.sh deploy plugin see here https://github.com/mailcow/mailcow-dockerized/issues/4564

DerLinkman:
Hello everyone,

as i have taken the position to maintain the repository and i'm not 100% sure what your PR does i would kindly ask you to describe what the benefits of ECDSA are and if it will break existing installations.

I've saw, that you've changed also the dovecot.conf for alt_certs. How does dovecot/postfix behave if these certs are not generated?

apio-sys:
I have proposed a PR for this here: https://github.com/mailcow/mailcow-dockerized/pull/7250 that is is much simpler with less changes. It's just ECDSA for new installs by default. Existing installs needs to choose to migrate if they want to (little procedure provided in the PR that should be made into a short doc if the PR gets approved, at least into staging...

PR #2457, opened 2019-03-20
[netfilter] Set F2B default values with environment variables

wedi:
This commit makes it possible to set default values for fail2ban without editing code files.
I wanted to avoid overloading mailcow.conf and still denote they are available so I added them to the Dockerfile.


mkuron:
You can set them in the admin UI already. I don't think it's a good idea to have the same thing twice in different places and with unclear precedence.

wedi:
That's why I have hidden it. It's docker so it is about automation. I'd like to be able to set mailcow up without having to touch the UI and I'd rather not use `sed` to change a source file to set some default values. I don't insist on environment variables. It's the docker way to set things but I will happily redo it differently as long as I don't have to use sed. :)

mkuron:
You can directly write your values to the Redis database if you want to automate things.

andryyy:
I'm fine with it, as long as we don't list it as official way to change those values. It can be helpful and it doesn't really hurt. :)

mkuron:
We will get bug reports from people who set the options in the admin UI and them complain that the environment variables don‘t work anymore. At least that‘s what my experiences suggest... @wedi, is there any reason why you can‘t just write the variables straight to the redis database?

wedi:
Because I would like to avoid opening one more way of fiddling with mailcows settings but I completely understand your reasoning. I thought that including `default` in the name and hiding them in the Dockerfile might be sufficient but probably it isn't.
Looks like I need to find a way to maintain a fork without getting behind with updates. This is probably the easiest way to keep track of changes. My first take was checking out a vanilla mailcow and applying changes with ansible but that gradually gets out of hand. Do you have another suggestion? You are probably doing it the other way round: your private copy is the main repository where you integrate changes and then you merge things back to the open source one?

mkuron:
I‘d still suggest just setting these variables in Redis via a script. It‘s unlikely that that will ever break due to changes in Mailcow (at least not any more often than your environment variables solution would get broken). I‘m not sure it‘d be worth keeping a fork just for this.

Also, I‘m trying to understand why this feature is needed. Do you regularly create and destroy Mailcow instances and want to automate that? What values do you want to use for these variables? Maybe our defaults are not chosen optimally?

wedi:
> (at least not any more often than your environment variables solution would get broken)

I understand this is no solution for widespread use. I will close this pull request.


>  I‘m not sure it‘d be worth keeping a fork just for this.

haha. No, not just for this little change. That would definitely be overkill. :D I have a few customizations, the biggest beeing the removal of SOGo which is baked in so tightly at various places.


> Do you regularly create and destroy Mailcow instances and want to automate that?

If something goes wrong, I want to be able to reprovision a mailserver with all required settings in no time. Mail is too essential and it's hard to keep someone waiting who gets bugged by his mail client saying something is wrong.


> Maybe our defaults are not chosen optimally?

As usual: it depends. I think they are ok for a private mail server to keep friends from guessing the password. But blocking a whole IP for 30 minutes after 10 failures per 10 minutes is begging for a DoS attack in a business use case. From my experience, the very tight default limit gets hit too often as soon as more than very few users are accessing mailcow from one place. This is especially true in a migration situation (thus, soon after creating a new mailcow instance) when changed account settings are involved and some mail clients retry again and again with an old username and/or password.  

A proper solution could be locking single mailboxes which hit a certain failure rate (maybe using some exponential backoff) and locking IPs at a failure rates which suggest an automated attack, only.  
Of course you can get arbitrarily sophisticated with this by including successful logins per IP, number of mailboxes per domain etc but if this is really needed one would probably look somewhere else anyways.

For now I am happy with allowing 150 attempts per 10 minutes. I am pretty sure 1 failure per 4 seconds will still effectively stop any serious brute force attack while making sure - subject to more experience - no more or less hand dialed passwords keep a whole office from work.

PR #2461, opened 2019-03-21
Only relay emails from local docker network

AikoAppeldorn:
Hi,

is there a reason why all private networks are allowed to relay emails via Postfix?
I want to use mailcow in a private network and have found that anyone in that network can send email without authentication.

Now only the Docker network can forward e-mails.


andryyy:
What about non-default IPv4 networks? This will blow those installations.

AikoAppeldorn:
Yeah, you're right. But I think this is a security issue for mailcow installations in a private network. 
Maybe we could point this out at least in the documentation?

andryyy:
Feel free to submit a doc update, it would be very appreciated! :)

Howaner:
Hello,

this is a very big security problem which affects all mail server installations in a local subnet.

Imagine someone introducing Mailcow to a company.
Every single employee could send e-mails with arbitrary mail addresses without authentication.
Open Relays should always be deactivated by default.

Is it not possible to integrate a reading of the Docker networks and updating of Postfix local_networks into the postfix.sh?

Kind regards,
Franz

andryyy:
Check the code before deploying anything you didn’t write yourself. Add documentation or whatever. PRs are welcome. 

You need to take care of your stuff yourself. It is not a security issue.

andryyy:
I will try to add a check or something like that. A doc update is still appreciated. :)

patschi:
Why was this PR closed, and what you're meaning with "add a check"? I'm basically with @DevTek314 that only the internal Docker network should be able to send mails, and not all internal subnets.

Forcing an authorization might be a way better idea. If anything or someone infects the internal environment, it's possible to abuse the open internal relay. Also users are able to bypass any authorization.

IMHO it should be disabled by default, and docs should be updated to optionally opt-in enable this behavior in required environments. Quite many users might even not be aware of this and so there are quite a lot of "less secured" mailcow installations out there.

PR #2463, opened 2019-03-22
Only generate a LE certificate for domains that have a mailbox

develth:
This SQL query will just provide Domains that have a Mailbox created.
With this change we can reduce the amount of certificate requests that are not necessary.

### DIFF
--- a/data/Dockerfiles/acme/docker-entrypoint.sh
+++ b/data/Dockerfiles/acme/docker-entrypoint.sh
@@ -256,7 +256,7 @@ while true; do
   # IP and webroot challenge verification #
   while read domains; do
     SQL_DOMAIN_ARR+=("${domains}")
-  done < <(mysql --socket=/var/run/mysqld/mysqld.sock -u ${DBUSER} -p${DBPASS} ${DBNAME} -e "SELECT domain FROM domain WHERE backupmx=0" -Bs)
+  done < <(mysql --socket=/var/run/mysqld/mysqld.sock -u ${DBUSER} -p${DBPASS} ${DBNAME} -e "SELECT d.domain as domain from domain d left join (select domain, count(username) as mailboxes from mailbox group by domain) m on m.domain = d.domain where m.mailboxes > 0 and d.backupmx=0" -Bs)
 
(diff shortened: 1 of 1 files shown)


mkuron:
Good idea, but this change is going to be extremely annoying for people who set up a new Mailcow. They would need to manually restart the acme container after they add their first mailbox to a domain.

develth:
I can´t find a trigger for acme container restart on Domain add (or i don´t see it). So when the acme container is triggered when you add a domain?
Edit: maybe i can trigger the same (if exists) when adding the first mailbox to a domain

develth:
I just wanted to check my latest added Domains and recognized, that the script added a domain and requested a certificate that is not listed by this query.
How can this happen?

andryyy:
Why would a domain without mailboxes exist?

develth:
When you want to setup a Domain that just redirects named based and not a exact match.
Example:

```info@example.com``` to ```userxy@gmail.com```
```post@example.com```to ```useryz@web.de```

andryyy:
I see. :)

Adorfer:
Aren't those ceritificates as well used for smtp-tls? 
In that case: What will happen to alias-domains? And what about backup-MX? those do not have any mailbox as well.

ntimo:
@Adorfer No the cert is not used for smtp-TLS the cert that is used for that is the mail server Hostnamen certificate (mail.domain.com). The other domains are just added to cert for autoconfig.

Adorfer:
perhaps i am missing the point. So the LE certificates for the different mx-hostnames  (e.g. mail.domain1.com, mail.domain2.net, mail.domain3.org) come from another validation-script, which runs independend? Or are those only self-signed?

develth:
There are no different mx hostnames

Adorfer:
Perhaps setups differ. either with cnames or with a/aaaa inside every domain.
in other words: for this PR i would recommend an option/warning, that no mx-fqdn in a domain is possible if no mailboxes are created.

develth:
But a certificate for mail.[domain] will not be generated. 
So why warn for something that currently not exists?

PR #2464, opened 2019-03-22
added Forum link to README.md

dekyfin:
I added a link to the forum to the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,3 +12,7 @@ Please see [the official documentation](https://mailcow.github.io/mailcow-docker
 
 **Important**: mailcow makes use of various open-source software. Please assure you agree with their license before using mailcow.
 Any part of mailcow itself is released under **GNU General Public License, Version 3**.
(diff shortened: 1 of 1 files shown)


andryyy:
Please add it to the docs as unofficial community support source.

extremeshok:
alas, why not the official ?

andryyy:
I really, really appreciate the board.

But as long as I don't host it myself, I don't want to call it official. It is also not an official way to get support. I will probably not even participate in the near future (enotime). So it is hard to call it official. :(

Again, I really appreciate your work and hope people will frequently post/share on mailcow.farm!!
It is just that I don't want to call anything official, I don't host myself and have full control over.

I don't have much free time to maintain a board, that's why I never created a board in first place. :) So it will not help to grant me access or something - I cannot handle it right now.

Please add it as community support to the docs. It is still a great source to get support and ask questions!

PR #2493, opened 2019-04-03
Update generate_config.sh

zedtux:
This commit fixes issues when running the `generate_config.sh` script on macOS.

__Aside note__ it doesn't make sens to check the memory on macOS as it is used only to generate
the `mailcow.conf` file so using the `3670017` value makes all available and then it's up to the user to change the flags manually.

### DIFF
--- a/generate_config.sh
+++ b/generate_config.sh
@@ -51,7 +51,7 @@ done
 
 MEM_TOTAL=$(awk '/MemTotal/ {print $2}' /proc/meminfo)
 
-if [ ${MEM_TOTAL} -le "2621440" ]; then
+if [ "${MEM_TOTAL:-3670017}" -le "2621440" ]; then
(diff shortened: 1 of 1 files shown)


zedtux:
Yeah well, "it has not had recent activity" like I'm waiting for the mailcow guys reviewing this code, not like I should answer or give a feedback.

How much I hate those stupid bots ...

andryyy:
Instead of setting a fake value, shouldn't we add a warning and skip those checks?

zedtux:
Indeed, it sounds like a better option. Thank you @andryyy, I'll update my PR. 👍

zedtux:
Just like that

andryyy:
We will re-open when you update the PR. :)

zedtux:
Yep yep, thank you.

PR #2497, opened 2019-04-04
Fix HTML bug with French language

solracsf:
Just go to user demo on https://demo.mailcow.email/user?lang=fr to see the bug in action (no tab switching as tag isn't closed).


andryyy:
Thank you. :)

PR #2506, opened 2019-04-07
Added org labels to docker compose

ntimo:
The org labels can be used in combination with node exporter and Grafana to visualise the performance of the Mailcow Stack.


andryyy:
That’s not useful for everyone. 

Please add it to the docs to be used in a docker-compose. override.yml.

ntimo:
Sure but it also won't hurt to have to label added to the containers :)

ntimo:
Hier mal ein Beispiel Grafana board:
Screenshot 2019-04-07 at 13 28 24

andryyy:
But you don’t want to set fixed labels for everyone for something individual

Issue #461, opened 2017-07-15
The Let's Encrypt subjectAltName limit of 100 domains


chriscroome:
Let's Encrypt has a [limit of 100 Names per Certificate](https://letsencrypt.org/docs/rate-limits/) this means that a mail server at `mail.example.org` with `autoconfig.example.org` and `autodiscover.example.org` could only have a maximum of 48 additional domains adding with `autoconfig.*` and `autodiscover.*` subjectAltNames.

The situation will improve some when [Let's Encrypt supports wild cards in January 2018](https://letsencrypt.org/2017/07/06/wildcard-certificates-coming-jan-2018.html) then there could be a maximum of 100 domains per cert.

However I have a mailserver with 200+ domains that I'm in the process of migrating to Mailcow and I'd like to help to add support for Mailcow to run with multiple certs as a work around for this issue.

mkuron:
Basically, you'd need to automatically modify data/conf/nginx/site.conf to contain one `server` section for the first 48 domains plus the mailcow hostname and additional `server` sections for 50 further domains each. The ACME script would need to be modified to request certificates only containing the SANs for one `server` section each. Wildcard certificates "only" save you a factor of two, so you might just as well do it now without wildcard certificates since you a total of 400 SANs anyway.

Unfortunately, Nginx config is currently static in Mailcow and cannot easily be made such that it reads from the database.

My suggestion would be to simply use a reverse proxy to deal with the large number of domains. Modify the ACME script (data/Dockerfiles/acme/docker-entrypoint.sh) to set `$VALIDATED_CONFIG_DOMAINS` to be empty if there are more than 48 domains. Then, configure Apache or Nginx as a reverse proxy with one site that serves the mailcow hostname and anything in the `$ADDITIONAL_SAN` variable. Configure further sites for 50 domains each, each of them with `autoconfig` and `autodiscover` entries in the `ServerAlias`/`server_name` field. Now you can use Certbot to automatically request the certificates for the autoconfig and autodiscover domains and have the ACME container deal with the mailcow hostname's certificate only.

mkuron:
Actually, there is also another solution that wouldn't require so many certificates:

1. The [Thunderbird autoconfiguration specification](https://developer.mozilla.org/en-US/docs/Mozilla/Thunderbird/Autoconfiguration) says that only HTTP is used, so you don't need any certificates for _autoconfig.*_.
[...]

PR #2509, opened 2019-04-08
Different SSL Certificate for each Domain with TLS SNI

mhofer117:
This addresses https://github.com/mailcow/mailcow-dockerized/issues/461 and most of the related issues:

- Adds a new configuration option `ENABLE_SSL_SNI` which is disabled by default. It has to be specifically enabled or there will still be only one certificate (the same as before).
- The acme container was cleaned up a bit
- It now stores all certificates and related files in `data/assets/ssl/{certificate-name}/`. The main server certificate will additionally be copied to the existing location directly at `data/assets/ssl/`
- When upgrading, the existing certificate will be migrated to the new folder structure, but nothing else unless `ENABLE_SSL_SNI` is enabled.
- When the feature is enable, instead of one certificate, the following certificates will be generated:
  - A main server certificate with the hostname and all fqdn additional SANs
  - For each domain found in the database, one additional certificate with autodiscover.\*, autoconfig.\* and any other `ADDITIONAL_SAN` configured in this format (subdomain.\*)
- The existing account and private key is used for all certificates, this means all existing TLSA dns records stay valid.
- Changed nginx to serve individual vhosts with each certificate (SNI), unkown domains are still [served with the first server block](http://nginx.org/en/docs/http/request_processing.html) - acme can still validate new domains without restarting nginx first.
- Changed dovecot to serve all certificates with SNI using [local_name](https://wiki.dovecot.org/SSL/DovecotConfiguration#With_client_TLS_SNI_.28Server_Name_Indication.29_support)
- Changed postfix to serve all certificates with SNI using [tls_server_sni_maps](http://www.postfix.org/postconf.5.html#tls_server_sni_maps)

All existing config parameters and the default certificate location were unchanged. The user can also still place a custom default certificate which will disable acme.

#### Main benefit:
- allows unlimited domains because we are no longer limited to a single certificate with max. 100 subject alternative names. This should eliminate the need for most reverse proxy setups.

#### Drawbacks
- Not all clients support SNI, see: https://wiki.dovecot.org/SSL/SNIClientSupport, however the server hostname or any other configured SAN in the default server certificate should still be valid with these clients as the default certificate will be served on requests without SNI.

In my test setup with a few domains, everything is working as expected, but further testing is probably a good idea.

If you see any problems with this PR or have major concerns let me know and i'll try to solve them.


mkuron:
Why do Postfix and Dovecot even need certificates for host names other than $MAILCOW_HOSTNAME?

Proxying Postfix will likely break fail2ban because it won't pass through source IP addresses. Proxying port 25 would be really dangerous for that reason too (DNSBLs and trusted networks need the correct IP).

mhofer117:
Thanks for your feedback.

> Why do Postfix and Dovecot even need certificates for host names other than $MAILCOW_HOSTNAME?

Postfix does not get any other certificates.
Enabling other certificates with SNI allows using white-label domains when configuring the mail client (ie mail.anydomain.com).
While Dovecot does not _need_ them, I added this as default because if someone is already using subdomains as server addresses, this would break it, since the subdomain.* are not longer included in the default server certificate.

> Proxying Postfix will likely break fail2ban because it won't pass through source IP addresses. Proxying port 25 would be really dangerous for that reason too (DNSBLs and trusted networks need the correct IP).

It does not break it, as dovecot will do the authentication in place of postfix, see this log from dovecot:
```
submission-login: Remote closed connection (auth failed, 2 attempts in 108 secs): user=<user@x.y>, method=LOGIN, rip=84.x.x.x, lip=172.22.1.250, TLS, TLSv1.2 with cipher ECDHE-RSA-AES256-GCM-SHA384 (256/256 bits)
```
which is the same format that f2b already recognized from imap / pop3.

Port 25 is not proxied for the reason you mentioned - still directly mapped to postfix and not using SNI, only the server certificate.

Is there any other reason why postfix might cause issues when mails from authenticated users are proxied? Any spam checks or similar stuff on outgoing mails?

mkuron:
What is `submission_relay_port` in your dovecot.conf used for? If Dovecot delivers everything to that port, Postfix will not know the submitting user name and can't perform sender checks, rate-limiting, etc.

Does TLS (via STARTTLS) on port 587 still work?

> It does not break it, as dovecot will do the authentication in place of postfix, see this log from dovecot:

Interesting. I did not know that that is how it works.

> Is there any other reason why postfix might cause issues when mails from authenticated users are proxied? Any spam checks or similar stuff on outgoing mails?

I don't think we do rate-limiting per IP, so that would be fine.

mhofer117:
> What is `submission_relay_port` in your dovecot.conf used for? If Dovecot delivers everything to that port, Postfix will not know the submitting user name and can't perform sender checks, rate-limiting, etc.

That is correct and probably an issue atm. Thanks for pointing it out. I have researched this and think the XCLIENT commands from dovecot should solve this, which are already enabled by `submission_relay_trusted = yes`. However I will try to verify this evening what changes are needed in postfix so it trusts these commands and identifies the correct sender (or maybe it already does).

> Does TLS (via STARTTLS) on port 587 still work?

Yes, STARTTLS on port 587 and direct SSL on 465 work with no issues in my tests.

mhofer117:
After some changes, postfix now gets the correct sasl_sender as authenticated in dovecot.

However it's not applying the sender restrictions, still have to figure out how to apply them when relaying via dovecot.

mhofer117:
I have now figured out the correct XCLIENT config, turned out i had to whitelist the dovecot IP instead of hostname.
All sender restrictions are working because dovecot is telling postfix the real client IP and sender login with XCLIENT, see the following log:

```
---- SUCCESSFUL MAIL ----
Apr 10 22:25:01 mail2 postfix/postscreen[340]: CONNECT from [172.22.1.250]:50142 to [172.22.1.9]:25
Apr 10 22:25:01 mail2 postfix/postscreen[340]: WHITELISTED [172.22.1.250]:50142
Apr 10 22:25:01 mail2 postfix/smtpd[342]: connect from mailcowdockerized_dovecot-mailcow_1.mailcowdockerized_mailcow-network[172.22.1.250]
Apr 10 22:25:01 mail2 postfix/smtpd[342]: 97530166AAB: client=mailcowdockerized_dovecot-mailcow_1.mailcowdockerized_mailcow-network[84.x.x.x], sasl_method=XCLIENT, sasl_username=mhofer117@domain.local
Apr 10 22:25:01 mail2 postfix/cleanup[346]: 97530166AAB: message-id=<082f49bf-f372-3319-b2f4-1b1d7cf9c520@domain.local>
Apr 10 22:25:01 mail2 postfix/qmgr[337]: 97530166AAB: from=<mhofer117@domain.local>, size=1680, nrcpt=1 (queue active)
Apr 10 22:25:01 mail2 postfix/smtpd[342]: disconnect from mailcowdockerized_dovecot-mailcow_1.mailcowdockerized_mailcow-network[84.x.x.x] ehlo=1 mail=1 rcpt=1 data=1 quit=1 commands=5
Apr 10 22:25:02 mail2 postfix/lmtp[348]: 97530166AAB: to=<mhofer117@domain.local>, relay=dovecot[fd4d:6169:6c63:6f77::10]:24, delay=0.59, delays=0.39/0.03/0.05/0.13, dsn=2.0.0, status=sent (250 2.0.0 <mhofer117@domain.local> tAprAx5RrlyZAAAABx98KA Saved)
Apr 10 22:25:02 mail2 postfix/qmgr[337]: 97530166AAB: removed

---- SENDER RESTRICTIONS ----
Apr 10 22:24:54 mail2 postfix/postscreen[340]: CONNECT from [172.22.1.250]:50100 to [172.22.1.9]:25
Apr 10 22:24:54 mail2 postfix/postscreen[340]: WHITELISTED [172.22.1.250]:50100
Apr 10 22:24:54 mail2 postfix/smtpd[342]: connect from mailcowdockerized_dovecot-mailcow_1.mailcowdockerized_mailcow-network[172.22.1.250]
Apr 10 22:24:54 mail2 postfix/smtpd[342]: NOQUEUE: reject: RCPT from mailcowdockerized_dovecot-mailcow_1.mailcowdockerized_mailcow-network[84.x.x.x]: 550 5.1.0 <mhofer1171@domain.local>: Sender address rejected: User unknown in virtual mailbox table; from=<mhofer1171@domain.local> to=<mhofer117@domain.local> proto=ESMTP helo=<mail2.domain.local>
Apr 10 22:24:56 mail2 postfix/smtpd[342]: disconnect from mailcowdockerized_dovecot-mailcow_1.mailcowdockerized_mailcow-network[84.x.x.x] ehlo=1 mail=1 rcpt=0/1 quit=1 commands=3/4

---- RATE LIMITED ----
Apr 10 22:25:08 mail2 postfix/postscreen[340]: CONNECT from [172.22.1.250]:50200 to [172.22.1.9]:25
Apr 10 22:25:08 mail2 postfix/postscreen[340]: WHITELISTED [172.22.1.250]:50200
Apr 10 22:25:08 mail2 postfix/smtpd[342]: connect from mailcowdockerized_dovecot-mailcow_1.mailcowdockerized_mailcow-network[172.22.1.250]
Apr 10 22:25:09 mail2 postfix/smtpd[342]: 017B1166AAB: client=mailcowdockerized_dovecot-mailcow_1.mailcowdockerized_mailcow-network[84.x.x.x], sasl_method=XCLIENT, sasl_username=mhofer117@domain.local
Apr 10 22:25:09 mail2 postfix/cleanup[346]: 017B1166AAB: message-id=<112afa8a-57d0-1357-25dd-ff0ab587a64c@domain.local>
Apr 10 22:25:09 mail2 postfix/cleanup[346]: 017B1166AAB: milter-reject: END-OF-MESSAGE from mailcowdockerized_dovecot-mailcow_1.mailcowdockerized_mailcow-network[84.x.x.x]: 4.7.1 Ratelimit "mailcow" exceeded; from=<mhofer117@domain.local> to=<mhofer117@domain.local> proto=ESMTP helo=<mail2.domain.local>
Apr 10 22:25:10 mail2 postfix/smtpd[342]: disconnect from mailcowdockerized_dovecot-mailcow_1.mailcowdockerized_mailcow-network[84.x.x.x] ehlo=1 mail=1 rcpt=1 data=0/1 quit=1 commands=4/5
```

Please let me know if you have further questions or concerns - or how I can improve this PR.

andryyy:
I’m not sure I want Dovecot to do the submission. How does that limit our possibilities to configure the service in Postfix?

You relay over port 25 back to postfix. You are always using an unencrypted connection. I know SOGo does the same, but one can avoid SOGo, while this would be active for all submissions. While I’m fine with this over sockets, we should avoid it for submission. 

How about failed certificate requests for a single domain? What’s the fall back? Where are additional SANs appended to? How do additional SAN migrate (or where to)? What if someone decided previously to add an additional SAN to connect to his mail server? This name is put into the general certificate of the server, I guess?

I’m on mobile and haven’t checked yet, so I’m not sure if/how that’s added: What is the default host of each individual FQDN? Like “mail” in “mail.example.org”. Can people change that?

If customer A asks for “mymx” and the other asks for “mail01”, it will probably fall back to the general certificate and they will see each other’s name in the certificate, right? I cannot add additional SANs per domain. 

Wouldn’t it be easier and less hassle to just place a HAProxy in front of mailcow? :/

mhofer117:
> I’m not sure I want Dovecot to do the submission. How does that limit our possibilities to configure the service in Postfix?

I don't think it limits the possibilities with Postfix. The only thing i can think of that Postfix does not know about is the client connection encryption. Are there any rules regarding incoming connection encryption?
For authentication I believe Postfix was handing it off to Dovecot already. Now Dovecot will do the authentication directly before acting as a relay (on Port 465 and 587).

> You relay over port 25 back to postfix. You are always using an unencrypted connection. I know SOGo does the same, but one can avoid SOGo, while this would be active for all submissions. While I’m fine with this over sockets, we should avoid it for submission.

Yes, I am using Port 25 to avoid an additional tls overhead for each connection. We could change it to SSL or STARTTLS, but why do you think it is important in a container to container connection?

> How about failed certificate requests for a single domain? What’s the fall back?

The same as before: The container will retry after 30 minutes instead of 1 day. Other certificates that were successful will still be activated immediately.

> Where are additional SANs appended to? How do additional SAN migrate (or where to)? What if someone decided previously to add an additional SAN to connect to his mail server? This name is put into the general certificate of the server, I guess?

Yes. I am still using the same config value and domain variables:
```
  IFS=',' read -r -a TMP_ARR <<< "${ADDITIONAL_SAN}"
  for i in "${TMP_ARR[@]}" ; do
    if [[ "$i" =~ \.\*$ ]]; then
      ADDITIONAL_WC_ARR+=(${i::-2})
    else
      ADDITIONAL_SAN_ARR+=($i)
    fi
  done
ADDITIONAL_WC_ARR+=('autodiscover')
```
The only difference is that ADDITIONAL_WC_ARR will be used to create individual certificates, while ADDITIONAL_SAN_ARR will be appended to the default server certificate.


> I’m on mobile and haven’t checked yet, so I’m not sure if/how that’s added: What is the default host of each individual FQDN? Like “mail” in “mail.example.org”. Can people change that?

The default host will be the first domain.* available, in the order defined in the existing ADDITIONAL_SAN config. To change it one has to change the order in the ADDITIONAL_SAN config and it will be applied to all domains.
I did not add any more per default, so in the standard configuration it will be autodiscover.example.org (and nothing else).

> If customer A asks for “mymx” and the other asks for “mail01”, it will probably fall back to the general certificate and they will see each other’s name in the certificate, right? I cannot add additional SANs per domain.

Yes this is true for this implementation, except if you apply the subdomain to every domain with mymx.* and mail01.*
I did not expect to have individual SAN requirements per customer. So this will still be limited to 99 additional fully qualified SANs. We could probably try to map each one to the respective autodiscover.* certificate if this is desired?

> Wouldn’t it be easier and less hassle to just place a HAProxy in front of mailcow? :/

The point of this PR was to make this work out of the box, without needing to configure and maintain an additional reverse proxy / loadbalancer, since all required tools are already included.
In my opinion, these changes are quite simple, but I have also never configured a HAProxy.


Thank you for your feedback!

mkuron:
> Are there any rules regarding incoming connection encryption?

Yes, you can require incoming TLS per user. It's not a commonly used feature of Mailcow because some mail servers still don't deliver over TLS, but in theory it's a nice frature that shouldn't be broken.

mhofer117:
Does this apply for submission by users or only incoming mail by other providers? If it's only the latter it should not be broken as port 25 is still directly mapped to postfix.

patschi:
TBH: I've no clue about the real advantages and if this is really necessary. In general I don't like the idea that Postfix is now depending on Dovecot to work.

However. I'm not a fan of the migration over to certbot:
1. certbot is still not capable of issuing ECDSA certificates. Breaking #2443 completely.
2. certbot is so much bigger than ACME with so many python libraries and basically does the same thing.

mhofer117:
The reason i choose certbot is because it simplifies renewal and cert management, is well documentend, [is the recommended acme client](https://letsencrypt.org/docs/client-options/) and offers a useful plugin and hooks system.

The image size difference with certbot is 19MB and I personally prefer a long term reliable solution over a few megabytes.
While ECDSA is indeed missing, it appears to be in implementation with backwards compatibility in mind. But if this is the only reason preventing an approval of this PR, I'm willing to add the required implementation with a different client.

I'd like to know more about the reservations regarding the Dovecot submission proxy.
- Postfix already depends on Dovecot for authentication.
- Unauthenticated delivery is still directly provided via port 25.
- Dovecot uses the same authentication methods as with IMAP and POP3.
- The [XCLIENT](http://www.postfix.org/XCLIENT_README.html) extension appears to works well  and basically makes Dovecot a transparent proxy.
- And it even [adds some possible features like BURL](https://wiki.dovecot.org/Submission).

For me the Dovecot proxy is the natural choice in favor of additional software.


Personally I'd require this feature for a seamless migration off cPanel and the confidence I don't need to create new instances or manage additional proxies when reaching a certain amount of domains. Looking at the previous issues I'd say there are also other people who would benefit from this.

mkuron:
> I'd like to know more about the reservations regarding the Dovecot submission proxy.

I'm actually not too concerned about it. It just needs extensive testing etc. and we need to check whether it breaks any Postfix feature we use or might use in the future. As @andryyy said,

> How does that limit our possibilities to configure the service in Postfix?

If `XCLIENT` takes care of everything, it is probably ok.

mkuron:
One thing I just thought of: you need to make sure that the same peivate key is used for the certificates as before. Otherwise existing TLSA records become invalid. Does your pull request generate a new key for each domain? Does it continue to use the key generated by the old acme client?

mhofer117:
OK this is a really good point. My PR will create a new private key for each domain and certbot will create a new one on each renewal (but that can be changed). If all the existing TLSA records must be re-used, we'd have to use the same private key for each domain and keep it forever. This definitely requires some additional changes for this PR. I'm thinking about some compatibility mode which is enabled on upgrades but disabled on new installations, because I don't think it is a good practice to re-use the private key for multiple certs (or even for a longer duration). I'll try to solve that as soon as I can.

chriscroome:
Thanks for all your work on this @mhofer117, I hope your changes are able to be merged at some point, I'm one of those people running a virtual server just for hosting `auto*.*` reverse proxies.

For what it is worth I generally use [acme.sh](https://github.com/Neilpang/acme.sh) rather than certbot, it is written in Bash and does [support ECC certs](https://github.com/Neilpang/acme.sh#10-issue-ecc-certificates).

mhofer117:
I have reverted to use acme-tiny and using the existing private key for all certificates, so existing TLSA will still be valid for all domains. I also changed the PR description to match the changes.
This should also make it easy to integrate with https://github.com/mailcow/mailcow-dockerized/pull/2443, if you want I can already merge those changes into this PR?

Let me know if you see any other possible issues, for exemple (as already mentioned) regarding Postfix features we use or might use in the future that could be affected by the dovecort proxy.

lenovouser:
This won't break `SKIP_LETS_ENCRYPT=y` with custom certificates, correct?

mhofer117:
Yes it still works as before

mhofer117:
Because the ECDSA PR https://github.com/mailcow/mailcow-dockerized/pull/2443 looks complete and will probably be merged I have merged it into this PR too and made all required changes for SNI with both RSA and ECDSA certificates.

Let me know if something is unclear or needs to be changed.
If everything is OK and you consider mergin I'll add the required documentation changes.

andryyy:
I don't know if this is a merger tbh.

I use custom submission configs quite often. That's to much of a hassle for the benefit in return.

mhofer117:
What type of custom submission configs? What would prevent you from still using them?

SamuelNitsche:
Any updates on this? I would love to have these features added.

andryyy:
I will not use Dovecot submission.

Adorfer:
It looks like this is a fix for the issue of "DNS setup issues for individual domains-> breaking many maildomains". 
I host the mail for some domains where the DNS is not under my control. And if things go wrong there, it spoils the CertificateRenewal for the other domains until i find time to look into the logs and disable stuff manually.

andryyy:
If it fails to validate a domain, it skips this domain and goes on.

SamuelNitsche:
> I will not use Dovecot submission.

I'm not quite sure what you mean by that. Can you explain that any further?

mamash:
BTW Postfix *does* support SNI [as of the February 3.4.0 release](http://www.postfix.org/announcements/postfix-3.4.0.html).

mkuron:
That‘s good to hear and would be a much simpler solution. Perhaps @mhofer117 wants to adapt his pull request once we upgrade to Postfix 3.4. Since we use distribution packages, that would require switching to Debian 10 or waiting for Ubuntu 20.04.

andryyy:
I created an image yesterday, needs more testing, but should be stable (based on buster).

The TLSA stuff is still preventing me to consider this. Just like the other PR.

I don't want to break existing setups on updates. But I also want to give them the possibility to upgrade etc.

Watchdog should probably test for TLSA records, too.

mkuron:
i think the fix for upgrading with TLSA was to make ECDSA optional, as is currently implemented here.

mkuron:
i think the fix for upgrading with TLSA was to make ECDSA optional, as is currently implemented here. And the problem with separate certificates could be to use the same (existing) private key on all of certificates when upgrading.

mhofer117:
Thanks for the heads up @mamash !
I can definitly remove the XCLIENT stuff in favor of native SNI in Postfix 3.4+.

@andryy TLSA won't break because the existing private key is reused. It only needs to be touched if ECDSA certs are activated.

anned20:
What's the state of this issue? We're hitting the 100 domain limit too

mhofer117:
Since https://github.com/mailcow/mailcow-dockerized/commit/22cecc3f854f030fde5ef4bdd52acba980e8cbcc mailcow postfix is on debian buster and [should be on version 3.4](https://packages.debian.org/buster/postfix).

I started checking the resources/docs on it and will start migrating this PR to remove XCLIENT once I have some more time. Will need some re-testing and ensure that nothing was lost on the upstream merges.

mhofer117:
I have removed XCLIENT and enabled native postfix SNI with this PR.
Additionally, I added a configuration option to enable this feature, which is disabled by default.
The initial comment in this PR was updated to reflect the new changes.

Please have a look @andryyy  , @mkuron

anned20:
Alright, we've taken the risk and forked this pull request. We're successfully running it for around 4 days now. It's working great! Thanks for this @mhofer117, I hope it gets merged soon.

mhofer117:
Ping @andryyy can you give a heads up on what is preventing this from getting merged? Is it the ECDSA part or something else?
https://github.com/mailcow/mailcow-dockerized/issues/461 is a long standing issue and has the "sponsorship request" tag which this PR is. It's a hassle checking all the commits to the master branch and manually re-doing them here when conflicts arise.

patschi:
I'd also like to see this in mailcow, even the fact I don't really need it for my setup. However I think Andre is a bit concerned about the upgrade path or rolling out so many chances at once. Even when it works on particular or new setups, it might cause issues on some (older) existing setups.

@andryyy will hate me for that for sure, but IMHO there should be at least a tiny bit of progress of this PR. @mhofer117 invested so much effort and time into this huge PR so that it might be worth continuing it, even tiny steps forward. Even when it's just a list of test scenarios or whatsoever.

There were huge changes in this PR, so the migration of existing setups to this changes must be basically bulletproof.

mamash:
That's exactly my concern. I wish I'd waited a bit more before migrating my dozen domains to Mailcow in August. Now I have absolutely no idea what'd happen if I pulled this into my local branch.

anned20:
A couple of weeks ago we migrated our production server to a local fork of this branch with the patches of @mhofer117 applied because we hit the domain limit. It was a smooth transition and everything has worked perfectly since then. I can confirm that he did a great job to make it as smooth as possible.

mhofer117:
Hi @patschi and @mamash, thank you for the feedback.

Most of the commit history is from merging the master branch and resolving conflicts.

The changes in postfix/dovecot/nginx are actually minimal. The changes to the acme container are also moderate, it's just that I moved some functionality out of the docker-entrypoint.sh to make it smaller and easier to understand/maintain. In addition, ENABLE_SSL_SNI has to be manually activated so in existing setups barely anything changes.

The php/frontend stuff is only from the ECDSA PR which I was under the impression would be merged soon when reading the discussion there... that's why it was included here.

Because alot has been changed since the opening of this PR and the feedback, maybe it makes sense to create a new PR and only include the SNI part which would be smaller and create a new discussion for the current state? Still waiting for feedback from @andryyy

irgendwr:
@mhofer117 I asked andre about this PR a while ago and his response was that he has responded before and doesn't want to repeat himself just to hear everyone saying that his concerns are not valid.

I translated some of the things he mentioned ([see telegram messages](https://t.me/mailcow/15789)):

- it could work _but_...
- cert splitting or sth. should be integrated into the UI somehow
- this isn't a very clean solution
- users need to be able to upgrade without anything breaking (TLSA records, etc.)
- **documentation**(!) needs to be updated:
  - how to migrate keys on existing setups
  - what if I want to use something other than RSA?
  - what happens when splitting certs and the TLSA records are no longer valid?
  - etc.
- this is a huuuge change and he doesn't have a need for it
- even little changes can cause issues that can't be predicted, this is a big change and has to be rigorously tested which takes a lot of time and motivation that he doesn't have because he doesn't have a need for this change

**Disclaimer:** I don't have a deep technical understanding of this and the points above may not be entirely correct or complete.

**I would really recommend [joining the Telegram Channel](https://t.me/mailcow) and talking to him about this.**

andryyy:
@anned20 it is crazy to assume this is not going to fail on some systems as it is, because it worked for you.
You could have easily setup a reverse proxy, it would have been easier after all...

@mhofer117 what exactly happens, when a domain is no longer part of mailcow? Can you revoke it?
Could you please adjust site-defaults.conf to match the current site.conf?
How about setups with custom site.conf extensions? Extensions will be merged into each site config file, right? Not a big deal, just want to clarify. 

You reuse the private key, that's good. No broken TLSA.

You need to restart Nginx after adding a domain to be able to validate it? Can you change that behavior?
What happens anyway after adding the domain? Is anything triggered to add new config files etc.? You cannot tell people to restart the stack after adding a new domain. What exactly do you need to do to trigger a new certificate?

If you change that, try to keep it in a single place. Not some stuff here, some there. I can see acme-mailcow racing with the Nginx container for example. No site is available yet, acme-mailcow tries to validate and fails.

andryyy:
If we clarify this, I will merge it. Ok? :)

There is a change in my queue, that will need minor adjustments in this PR. Hope that's not a problem. I changed some file names in Dockerfiles. Not a big deal, but it will break this PR (as it happened a lot of times now).

anned20:
@andryyy 
> @anned20 it is crazy to assume this is not going to fail on some systems as it is, because it worked for you.
> You could have easily setup a reverse proxy, it would have been easier after all...

I know, you're right. But I wanted to see if it could be fixed by this pull request and it did. Just wanted to share a personal experience :)

mamash:
> @andryyy
> 
> > @anned20 it is crazy to assume this is not going to fail on some systems as it is, because it worked for you.
> > You could have easily setup a reverse proxy, it would have been easier after all...
> 
> I know, you're right. But I wanted to see if it could be fixed by this pull request and it did. Just wanted to share a personal experience :)

Also, I should say that for anything other than a bunch of alias domains, I feel more 'right' by taking the the cert-per-domain way, as long as there's proved technology allowing it.

(I know there are always conservative ways to achieve goals and work around problems, and I myself have played the "I'll just write this in Bash" card way too many times and it worked.)

andryyy:
The proven technologies? Which? All of this is pretty new (SNI and multi-cert in Dovecot/Postfix). 

Splitting sites is not a technology. It is technology wise not different from what we do now. It is about the implementation. SNI can be a problem for older clients, too. 

Thanks for your feedback @anned20 :)

mhofer117:
> @mhofer117 what exactly happens, when a domain is no longer part of mailcow? Can you revoke it?

the certificate will be moved to the backups, it will not revoke it on LE, but this could be added if wanted.

---

> Could you please adjust site-defaults.conf to match the current site.conf?

It is already up to date (just checked again). The missing lines are the following:
```
ssl_certificate /etc/ssl/mail/cert.pem;
ssl_certificate_key /etc/ssl/mail/key.pem;
listen 127.0.0.1:65510;
include /etc/nginx/conf.d/listen_plain.active;
include /etc/nginx/conf.d/listen_ssl.active;
include /etc/nginx/conf.d/server_name.active;
```
which are all handled per vhost now in the `site.template.sh`.

---

> How about setups with custom site.conf extensions? Extensions will be merged into each site config file, right? Not a big deal, just want to clarify.

This is correct.

---

> You need to restart Nginx after adding a domain to be able to validate it? Can you change that behavior?

Nginx does not need to be restarted for validation, as the first nginx server will be the default vhost for 'unknown' domains. But it has to be restarted instead of reloaded after a new or removed certificate because the server directives need to be re-generated. This is handled by the acme container.

---

> What happens anyway after adding the domain? Is anything triggered to add new config files etc.? You cannot tell people to restart the stack after adding a new domain. What exactly do you need to do to trigger a new certificate?

I haven't specifically checked what happens after adding a domain, but I did not change anything there. **Only the acme-container needs to be restarted** as it will handle any other container that needs a restart / reload (if needed). This must be done manually now if you don't want to wait for the next automatic run, same as it is now.
I'm open to suggestions here, maybe trigger the acme restart when adding a domain? Or add a button somewhere in the domain management?

---

> If you change that, try to keep it in a single place. Not some stuff here, some there. I can see acme-mailcow racing with the Nginx container for example. No site is available yet, acme-mailcow tries to validate and fails.

This should not be an issue because no other container needs to be restarted before acme, only after. Validation with nginx works for any new domain without restarting nginx.

---

> There is a change in my queue, that will need minor adjustments in this PR. Hope that's not a problem. I changed some file names in Dockerfiles. Not a big deal, but it will break this PR (as it happened a lot of times now).

no problem go ahead

---

Here is some shell output / logs where i verified some of the statements here: [acme tests](https://gist.github.com/mhofer117/a08afd57b697f323991c831941c6c9fe)

---

On the weekend I will try to split ecdsa from this PR.

mhofer117:
I have moved the ECDSA part to a different branch now so it can be evaluated individually.
I also archived and rebased this one to clean up the commit history.

andryyy:
Do you think this can be merged?
@mkuron?
If @mkuron says it is fine and you think this PR is complete, let's do it.

mkuron:
Looks good overall, but I want to test it myself. I'll do that either later today or tomorrow.

andryyy:
Great, ok.

mkuron:
1. Looked at the code and it seems fine.
2. Merged it on my production server and confirmed that it doesn't break anything.
3. Enabled SNI and restarted my acme container. It correctly fetched the certificates and configured and restarted the services.

------

However, the migration did not correctly deal with the fact that the old certificate still has some of the same server names as the new one:
```
$ grep . data/assets/ssl/*/domains
data/assets/ssl/autodiscover.example.com/domains: autodiscover.example.com autoconfig.example.com
data/assets/ssl/mailcow.example.com/domains: mailcow.example.com autoconfig.example.com autodiscover.example.com somethingelse.example.com
```
The Nginx/Postfix/Dovecot SNI maps contain duplicates/conflicting information:
```
$ grep 'server_name\|cert\.pem' /etc/nginx/conf.d/sites.active 
  ssl_certificate /etc/ssl/mail/cert.pem;
  server_name mailcow.example.com autodiscover.* autoconfig.*;
  ssl_certificate /etc/ssl/mail/autodiscover.example.com/cert.pem;
  server_name autodiscover.example.com autoconfig.example.com;
  ssl_certificate /etc/ssl/mail/mailcow.example.com/cert.pem;
  server_name autoconfig.example.com autodiscover.example.com somethingelse.example.com;
```
Note that I have one additional SAN (somethingelse.example.com) configured.

After manually removing autodiscover.example.com and autoconfig.example.com from data/assets/ssl/mailcow.example.com/domains, everything was fine.

------

Also, we need to add a warning to the news at https://mailcow.email that the certificate path has changed from data/assets/ssl/ to data/assets/ssl/$MAILCOW_HOSTNAME. This is relevant to people who manually manage their certificates and have set up some kind of script that copies in the certificate from elsewhere.

data/Dockerfiles/acme/acme.sh
```diff
@@ -56,97 +46,21 @@ mkdir -p ${ACME_BASE}/acme
...
+  if verify_hash_match ${ACME_BASE}/acme/cert.pem ${ACME_BASE}/acme/key.pem; then
+    log_f "Migrating to SNI folder structure..." no_nl
+    CERT_DOMAIN=($(openssl x509 -noout -text -in ${ACME_BASE}/acme/cert.pem | grep "Subject:" | sed -e 's/\(Subject:\)\|\(CN = \)\|\(CN=\)//g' | sed -e 's/^[[:space:]]*//'))
+    CERT_DOMAINS=(${CERT_DOMAIN} $(openssl x509 -noout -text -in ${ACME_BASE}/acme/cert.pem | grep "DNS:" | sed -e 's/\(DNS:\)\|,//g' | sed "s/${CERT_DOMAIN}//" | sed -e 's/^[[:space:]]*//'))
+    mkdir -p ${ACME_BASE}/${CERT_DOMAIN}
+    mv ${ACME_BASE}/acme/cert.pem ${ACME_BASE}/${CERT_DOMAIN}/cert.pem
+    # key is only copied, not moved, because it is used by all other requests too
+    cp ${ACME_BASE}/acme/key.pem ${ACME_BASE}/${CERT_DOMAIN}/key.pem
+    chmod 600 ${ACME_BASE}/${CERT_DOMAIN}/key.pem
+    echo -n ${CERT_DOMAINS[*]} > ${ACME_BASE}/${CERT_DOMAIN}/domains
```
mkuron, line 59: This is where things are going wrong. Please filter out any domains starting with _autoconfig._ or _autodiscover._. Once that's done, we can go ahead and merge this pull request.

andryyy:
Changing the path is indeed critical. A lot (!) of users use an external acme client. We cannot break these setups or only write small info on mailcow.email to inform them. I think this needs be handled differently.

mkuron:
It's not too bad actually. If you just put new certificates in the old location even after a migration has already been performed, they will be correctly migrated to the new location the next time acme runs.

mhofer117:
> The Nginx/Postfix/Dovecot SNI maps contain duplicates/conflicting information:
> 
> ```
> $ grep 'server_name\|cert\.pem' /etc/nginx/conf.d/sites.active 
>   ssl_certificate /etc/ssl/mail/cert.pem;
>   server_name mailcow.example.com autodiscover.* autoconfig.*;
>   ssl_certificate /etc/ssl/mail/autodiscover.example.com/cert.pem;
>   server_name autodiscover.example.com autoconfig.example.com;
>   ssl_certificate /etc/ssl/mail/mailcow.example.com/cert.pem;
>   server_name autoconfig.example.com autodiscover.example.com somethingelse.example.com;
> ```

I don't completely understand your setup. Do you maybe have "autoconfig.example.com"  and "autodiscover.example.com"  (including the domain part) in your SAN config?

----

> Also, we need to add a warning to the news at https://mailcow.email that the certificate path has changed from data/assets/ssl/ to data/assets/ssl/$MAILCOW_HOSTNAME. This is relevant to people who manually manage their certificates and have set up some kind of script that copies in the certificate from elsewhere.

The certificate in `data/assets/ssl/$MAILCOW_HOSTNAME/` is symlinked to the old location at `data/assets/ssl/` specifically for compatibility. You can see it in your nginx config, the `/etc/ssl/mail/cert.pem` is still used as the default server certificate. And all certificates under `/etc/ssl/mail/{hostname}/` are only loaded if present with the domains file.

mhofer117:
> It's not too bad actually. If you just put new certificates in the old location even after a migration has already been performed, they will be correctly migrated to the new location the next time acme runs.

If someone is using an external acme client, he should have disabled the mailcow-acme. But anyway the old certificate location is still valid.

data/Dockerfiles/acme/acme.sh
mhofer117: What if someone has configured "autoconfig.something.tld" as a SAN and "something.tld" is not setup inside mailcow? Then that domain would be missing from the certificate. Maybe we should filter all domains part of the default certificates from all additional certificates instead?

mkuron:
Thanks for clarification. For some reason, I indeed had autoconfig.example.com and autodiscover.example.com as additional SANs. Removed them and am currently testing whether that resolves all my concerns (I expect it does).

mkuron:
Thanks for clarification. For some reason, I indeed had autoconfig.example.com and autodiscover.example.com as additional SANs. Removed them and everything is fine now. Please add a check whether any of the automatically-generated autoconfig.example.com/autodiscover.example.com are present in ADDITIONAL_SAN and remove them from ADDITIONAL_SAN if found. Once that's done, I think this pull request should be merged.

> The certificate in data/assets/ssl/$MAILCOW_HOSTNAME/ is symlinked to the old location at data/assets/ssl/ specifically for compatibility.

It seems like they are copied, but it was the same before. Previously they came from data/assets/ssl/acme, now they come from data/ssl/$DOMAIN.

mhofer117:
> Thanks for clarification. For some reason, I indeed had autoconfig.example.com and autodiscover.example.com as additional SANs. Removed them and everything is fine now. Please add a check whether any of the automatically-generated autoconfig.example.com/autodiscover.example.com are present in ADDITIONAL_SAN and remove them from ADDITIONAL_SAN if found. Once that's done, I think this pull request should be merged.

Would it also be OK if all fully qualified names in the ADDITIONAL_SAN are removed from the auto-generated certificates? Because I think it's a valid use-case that did not occur to me before: Someone might want to have autoconfig.example.com in the default server certificate instead of the autogenerated one for example.com.

What do you think?

mkuron:
I didn't really understand what you said there. If you meant that we should not put autoconfig.example.com/autodiscover.example.com into the secondary certificate if ADDITIONAL_SAN already puts it into the main certificate, then I'm fine with that.

andryyy:
Not sure. One of the goals was not to mix different domain names in a cert (""information leak"").

mhofer117:
> I didn't really understand what you said there. If you meant that we should not put autoconfig.example.com/autodiscover.example.com into the secondary certificate if ADDITIONAL_SAN already puts it into the main certificate, then I'm fine with that.

This is exactly what I meant. Thank you.

---

> Not sure. One of the goals was not to mix different domain names in a cert (""information leak"").

This is true but in this case it would be explicitly defined and the admin can remove them in ADDITIONAL_SAN to prevent this. Otherwise there is no option to include a domain explicitly in the default certificate if it is also in the autogenerated certificate.

I'm fine with both solutions but would prefer the one I suggested if that's OK for you @andryyy ?

mkuron:
> One of the goals was not to mix different domain names in a cert

No big deal here since it‘s explicit. It‘s not going to be a common scenario anyway, just meant to eliminate the problem I ran into.

Hackebein:
> This is true but in this case it would be explicitly defined and the admin can remove them ADDITIONAL_SAN to prevent this. Otherwise there is no option to include a domain explicitly in the default certificate.

Think about that: How do I add an FQDN explicitly only for one domain? Example, i got want mail.example.com but don't want expose it via default cert and don't want mail.example2.com.
I think this is the best solution:
> Please add a check whether any of the automatically-generated autoconfig.example.com/autodiscover.example.com are present in ADDITIONAL_SAN and remove them from ADDITIONAL_SAN if found. Once that's done, I think this pull request should be merged.

Or did i miss something here that this is not possible?

mhofer117:
> 
> 
> > This is true but in this case it would be explicitly defined and the admin can remove them ADDITIONAL_SAN to prevent this. Otherwise there is no option to include a domain explicitly in the default certificate.
> 
> Think about that: How do I add an FQDN explicitly only for one domain? Example, i got want mail.example.com but don't want expose it via default cert and don't want mail.example2.com.
> I think this is the best solution:
> 
> > Please add a check whether any of the automatically-generated autoconfig.example.com/autodiscover.example.com are present in ADDITIONAL_SAN and remove them from ADDITIONAL_SAN if found. Once that's done, I think this pull request should be merged.
> 
> Or did i miss something here that this is not possible?

Yes adding an FQDN only for one domain will always include it in the default certificate, the autogenerated certificates only contain entries for *all* domains (autoconfig.\*, autodiscover.\*, mail.\*, ...). It would be quite complex to check every ADDITIONAL_SAN for matches against the DB and include them in the respective certificate (which could still be different than what the user actually wants).

mhofer117:
I have added the suggested check / fix.

mkuron:
Ok, looks good to me. @andryyy, you can merge if you agree.

We just need to remember that we cannot easily revert this pull request. Files would need to be moved back to their original location etc. So we will need to fix any bugs that may arise. I don‘t expect anything major to come up though (I‘ve tested it and @mhofer117 has done a solid job) and it‘s definitely a worthwhile improvement.

andryyy:
Yes, looks good. I'm fine with it.

mkuron:
I think you also need to rebuild postfix. 0efa5e50b816ee0b761b76f78161d4461c48abea only updates acme and dovecot.

andryyy:
Thanks

mhofer117:
Thanks everyone for the feedback and recommendations!

I'll follow the telegram chat and github to assist on any issues.

mamash:
Awesome, thanks guys! I'll update my installation next weekend.

Geitenijs:
Looks good!

Only thing I noticed is that [this configuration](https://mailcow.github.io/mailcow-dockerized-docs/u_e-80_to_443/) does no longer work. Also tried commenting out the lines in listen_plain.active, but it gets reverted when restarting nginx.

mhofer117:
Did you pull the latest changes? There was a fix for this some hours ago.

I also recommend not changing the site.conf but instead create a new file for this, for example: `data/conf/nginx/redirect.conf`

This prevents git conflicts

andryyy:
Well, if that's broken and everyone needs to apply a change to site.conf to keep the redirect, that's not good.

Your recent change should not break it, right? :)

We should change the docs to redir.conf anyway, no matter if it works now, that's true.

Geitenijs:
@mhofer117 Yes I did. The redirect.conf file doesn't work either :(

mhofer117:
> Your recent change should not break it, right? :)

No should not. I think that documentation page is misleading anyway. That config would also redirect acme challenges? And it includes the same domain names as the integrated server directive, so nginx will complain about duplicates. I think all of that was the case also before this PR.

I have updated the docs here: https://github.com/mailcow/mailcow-dockerized-docs/pull/175

---

> @mhofer117 Yes I did. The redirect.conf file doesn't work either :(

Can you try the config from the documentation PR I linked?
Was it working before the TLS-SNI update?

Geitenijs:
@mhofer117 Just tested it, doesn't work. Pretty sure it worked before the TLS-SNI update, at least a couple of weeks ago it did.

mhofer117:
Hmm I think i know why, let me test and provide a fix.

mhofer117:
I think the following PR fixes this:
https://github.com/mailcow/mailcow-dockerized/pull/3071

There is no reason why the TLS-SNI servers should also listen on http-80.
So now any server with `listen 80 default_server;` will serve all domain name again.

Andre merged already, can you retest @Geitenijs

Geitenijs:
@mhofer117 Great! It seems to work only for the mail. subdomains etc, but not for the server hostname (mx.example.com)

@andryyy Shouldn't we just disable HTTP altogether by default? Always redirect to HTTPS, unless the server admin prefers to have HTTP support

mhofer117:
Check the latest change from andre:
https://mailcow.github.io/mailcow-dockerized-docs/u_e-80_to_443/

(server_name directive)

Geitenijs:
@mhofer117 Sorry, missed that one. Works perfectly now, thanks!

christianbur:
Has someone tested if the SNI extension works with wilcard domains (external acme client)?

mhofer117:
SNI is unrelated to wildcard certificates.
When using an external acme client:
- You should not ENABLE_SSL_SNI, except if your external acme client creates more than one certificate with the same folder structure and files as the mailcow-acme container.
- If your wildcard certificate is at the default certificate location, it will just work as before and be the only certificate (without SNI).

Please elaborate some more if you have a specific setup / question?

christianbur:
Under data/assets/ssl/ a folder is created for each domain (e.g. mail.example.net), in this folder you will find cert.pem, key.pem and a domains file (e.g. mail.example.net). 
For the main domain the files data/assets/ssl/cert.pem and data/assets/ssl/key.pem are created in addition. 
ENABLE_SSL_SNI does not need to be activated with an external acme client, since this variable is only evaluated by the acme container.

What I currently don't understand is how to name the domain folder and what I write to the domains file when I use a wildcard domain. 

According to [Postfix instructions](http://www.postfix.org/postconf.5.html) you should use .example.net, so I would name the folder "example.net" and write ".example.net" into the domains file. 
Whether Dovecot and nginx understand this is not clear to me at the moment, so I simply asked.

mhofer117:
You are right, ENABLE_SSL_SNI is not relevant.
The folder name at data/assets/ssl/* is not important, you can name it anything.
The data/assets/ssl/*/domains file is important, however I could not find the correct syntax for postfix.
For dovecot you need `example.net .example.net` inside the domains file.
nginx has a different syntax and requires `*.example.net`.
Not sure if it works if you combine it like `example.net .example.net *.example.net` or if any of the services will complain then...
Can you try it out and report back?

christianbur:
so my test results:

### Postfix
`echo "Q" | openssl s_client -starttls smtp -crlf -connect mail.example.net:587 -servername mail.example.net|grep subject`
  -> example.net in domain file -> main domain
  -> .example.net in domain file -> sni domain
  -> *.example.net in domain file -> main domain
  INFO: http://www.postfix.org/postconf.5.html, section tls_server_sni_maps -> use ".example.net" for wildcard domains

### Dovecot
`echo "Q" | openssl s_client -showcerts -connect mail.example.net:993  -servername mail.example.net |grep subject `
   -> example.net in domain file -> main domain
  -> .example.net in domain file -> main domain
  -> *.example.net in domain file -> sni domain
  INFO: No wildcard example found 

### Nginx (test port 8443)
`echo "Q" | openssl s_client -connect mail.example.net:8443 -servername mail.example.net |grep subject`
  -> example.net in domain file -> main domain
  -> .example.net in domain file -> sni domain
  -> *.example.net in domain file -> sni domain
  INFO: https://nginx.org/en/docs/http/server_names.html, section Optimization
            .example.net = *.example.net and example.net
            *.example.net = .example.net without example.net

so an adjustment in Dovecot script (Dockerfiles/dovecot/docker-entrypoint.sh) would be necessary (rewrite .example.net -> *.example.net) or you use *.example.net and adapt nginx and Postfix.

andryyy:
Whatever you change. Make sure it does not break anything that works _now_.

Issue #2555, opened 2019-04-24
GZIP Eroor while installation


den1982:
Hello,

i get a gzip error during the docker-compose up -d installation phase:

`Processing triggers for ca-certificates (20161130+nmu1+deb9u1) ...
Updating certificates in /etc/ssl/certs...
0 added, 0 removed; done.
Running hooks in /etc/ca-certificates/update.d...
done.
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   263  100   263    0     0   4534      0 --:--:-- --:--:-- --:--:--  4614

gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
[...]

PR #2556, opened 2019-04-24
Fixing #2555 by updating Dockerfile to use the correct dovecot.org url (and fixing the http 304 curl redirect problem)

tha80:
Just a quick fix to have to dovecot image in a buildable state again :)


mkuron:
That's an unexpected fix. But it seems like the official download page is without www, so it's ok.

andryyy:
Yes, they reworked their sites. Thanks.

PR #2565, opened 2019-04-28
Updated php-mime-mail-parser library to 5.0 to fix webui html preview

Howaner:
Preview versions of php-mime-mail-parser uses iconv for the most cases and iconv isn't properly supported in alpine php (https://github.com/docker-library/php/issues/240)

Fixes the following error at quarantine mail preview: ```Parser failed, assuming HTML```

### DIFF
--- a/data/web/inc/lib/vendor/php-mime-mail-parser/php-mime-mail-parser/compile_mailparse.sh
+++ b/data/web/inc/lib/vendor/php-mime-mail-parser/php-mime-mail-parser/compile_mailparse.sh
@@ -1,0 +1,10 @@
+#!/bin/sh
+
+git clone https://github.com/php/pecl-mail-mailparse.git
+cd pecl-mail-mailparse
+phpize
+./configure
+sed -i 's/#if\s!HAVE_MBSTRING/#ifndef MBFL_MBFILTER_H/' ./mailparse.c
+make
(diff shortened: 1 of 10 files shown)


andryyy:
Thanks. :)

andryyy:
How does it fix it?

image

Does it need further changes? :) I can check it then.

data/web/inc/lib/vendor/php-mime-mail-parser/php-mime-mail-parser/compile_mailparse.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
```
ntimo, line 1: Why is there a .sh file in the web directory? This makes no sense. Also the file uses paths that have nothing to do with Mailcow? `/home/travis/.phpenv/versions/7.3.2/lib/php/extensions/no-debug-zts-20180731/` .

andryyy:
It is a compile script. Probably for Travis? We can delete it, but it does not harm. 

Im almost asleep. You can create a PR to delete that file.

Howaner:
> How does it fix it?
> 
> image
> 
> Does it need further changes? :) I can check it then.

It worked with iso encoding, but it doesn't work with UTF-8.
I started a discussion in the php-mime-mail-parser project: https://github.com/php-mime-mail-parser/php-mime-mail-parser/pull/242

andryyy:
Ahhh. I see. Thanks!!

PR #2567, opened 2019-04-30
Updated php-mime-mail-parser library to 5.1 to really fix webui html preview

Howaner:
Correctly fixes https://github.com/mailcow/mailcow-dockerized/pull/2565

The last charset fix in the php-mime-mail-parser library didn't work with utf-8.
Now it should work correctly.


andryyy:
image

Doesn't work at all anymore.

andryyy:
It is quite possible that I was stupid - once again - and forgot to add a text in my previous test mail.

PR #2569, opened 2019-04-30
Added domain alias handling to quarantine mails and added recipients row to quarantine mail display

Howaner:
If a mail is sent to a domain alias domain and rejected, mailcow does not currently store the mail in quarantine.
This commit adds domain alias handling to the reject code and should fix this behavior.

Also added displaying of recipient addresses into the quarantine mail dialog to be able to see what mail address was "leaked".

Dialog:
image


andryyy:
Really cool, will test and merge tomorrow or this weekend, okay? :)

andryyy:
Looks good! Thank you!

PR #2574, opened 2019-05-03
Handle mobileconfig display names with special characters

emericklaw:
If the account display name contained special characters like & the mobileconfig would fail to import on the iOS device.


mkuron:
You'll only want to replace ampersands and less/greater signs. Your current pull request also replaces double quotes (which carry no special meaning inside XML tags), so you should add `ENT_NOQUOTES`.

emericklaw:
Thanks @mkuron I have updated

Issue #2534, opened 2019-04-15
Sync Jobs with own parameter and space in them


sh-hilden:
I want to add an Sync Job which only download the inbox folder of my external mail account.
If I add 
`--folder "INBOX"` or `--folder INBOX` 
to the custom parameters I get the error: Unknown option: folder "inbox"
I have also tried 
`--include "^INBOX$"` 
with an empty exclude field but their I get the same error.
If  I enter commands without an space they work. Like `--dry`

jlelse:
I got the same error yesterday with regextrans2, when I tried a regex to move all mails to one folder. Gave up in the end.

markus-seidl:
Exactly the same problem, imapsync states, there may be a quoting issue. But without seeing the actual command line we can only guess.

andryyy:
https://github.com/mailcow/mailcow-dockerized/issues/2017

@hunter-nl may help

andryyy:
I mean ... we could get rid of that Perl stuff. :-D I don't know why we keep fixing it.

hunter-nl:
You should not use any quotes. The field in the web interface is just 1 string. This gets split on the -- characters (not on any spaces). Then each parameter will be added in quotes when the sync job command is executed. The = character behind the parameter is needed if parameter has a value 
Example:
`--addheader --domino1 --resyncflags --f1f2=Deleted Messages=Trash --f1f2=Sent Messages=Sent`
`--f1f2` is the parameter, `Deleted Messages=Trash` the value.

@sh-hilden Try `--folder=INBOX`

@andryyy imapsync script version 1.937 has been released (https://imapsync.lamiral.info There is also a docker image available https://hub.docker.com/r/gilleslamiral/imapsync/
imapsync script download location and other perl module requirements can be found in the dockerfile). It provides some useful fixes. However, It does not "fix" the issue above. The issue above caused by the way the shell command is combined and is executed. Maybe this part can get some improvement by doing this differently.

andryyy:
Oh, I will try to update asap. Please remind me if I forget. 

Id gladly donate to you for fixing this issue. :)

hunter-nl:
When I got back from my business travel, I'll have a look to try to fix the "issue" ;-)

andryyy:
@hunter-nl That would be amazing!

markus-seidl:
[...]

Issue #2017, opened 2018-11-16
IMAP sync custom options with a space in argument results in error


hunter-nl:
**Describe the bug**
When using custom options with a IMAP sync job, resulting in the following error:
`Extra arguments found: Messages"=Trash
It usually means a quoting issue in the command line
Docker context detected with /.dockerenv
Changing current directory to /var/tmp/
`
I guess the custom parameter field is not correctly quoted when added to the imapsync command line. I need the quotes because the space in a folder name.

**How to reproduce Reproduce**
Steps to reproduce the behavior please explain how to reproduce the bug:
1. sync job created
2. with option 'Try to automap folders' enabled.
3. Set custom parameter: `--dry --f1f2 "Deleted Messages"=Trash`

I tried different combinations of the double/single quotes.
Like:
`--dry --f1f2 'Deleted Messages'=Trash`
`--dry --f1f2 "'Deleted Messages'='Trash'"`
`"--dry --f1f2 'Deleted Messages'=Trash"`
etc
Results always in the similar error message.

**Expected behavior**
sync job should run with complete custom commands including the correct quotes.
`--dry --f1f2 "Deleted Messages"=Trash`
I added dry option to see if it will process the custom parameter(s) correctly or not.

**Logs**
Output of sync job:
`Extra arguments found: Messages"=Trash
It usually means a quoting issue in the command line
Docker context detected with /.dockerenv
Changing current directory to /var/tmp/
`

**System**
 - OS: Centos 7.5

hunter-nl:
In data/Dockerfiles/dovecot/imapsync_cron.pl file at line 21:
`sub qqw($) { split /\s+/, $_[0] }` is always split on any space it founds in the custom parameters field.
Split function should only happen on  each `--argument 'eventually some value'` parts.

hunter-nl:
@andryyy Can you apply this fix?
Replace 
In data/Dockerfiles/dovecot/imapsync_cron.pl file at line 21:
`sub qqw($) { split /\s+/, $_[0] }`
into
```
sub qqw($) {
        my @values = split('(?=--)', $_[0]);
        foreach my $val (@values) {
                $val=trim($val);
        }
        return @values
}
```
This results that custom parameters are split on the -- part. Example:
`'--option1 --option2 --option3 "test string " --option4 test with spaces '`
Results into a array:
```
[
          '--option1',
          '--option2',
          '--option3 "test string "',
          '--option4 test with spaces'
]
```

[...]

PR #2576, opened 2019-05-04
imapsync command execution changed

hunter-nl:
The imapsync_cron.pl script changed to solve users expectation of the provided parameters.
The execution of imapsync command line is now combined one command string instead of array of parameters. This allows the users to use special characters (like " ' | { } space) in the Username, Sync into subfolder on destination, Exclude objects (regex) and Custom parameters fields.
The user must now use quotes string values, like they already expected in the first place. It behaves now as if the parameters are provided on the CLI.
Example Custom parameters: --addheader --resyncflags --f1f2 "Deleted Messages"=Trash
Solve issues: #2534, #2017

Other fixes:
- Removed second unlock file statement
- Changed database schema: In table imapsync with column returned_text type has been changed into longtext. When imapsync is running on a large mailbox and for example --debugflags parameter has been provided then the returned_text output can be >16Mb.
- Changed the setting "is_running" status. There was a possibility to stuck into running state when an exception occurred.

### DIFF
--- a/data/Dockerfiles/dovecot/imapsync_cron.pl
+++ b/data/Dockerfiles/dovecot/imapsync_cron.pl
@@ -146,7 +146,6 @@ while ($row = $sth->fetchrow_arrayref()) {
     $update = $dbh->prepare("UPDATE imapsync SET last_run = NOW(), is_running = 0 WHERE id = ?");
     $update->bind_param( 1, ${id} );
     $update->execute();
-    $lockmgr->unlock($lock_file);
   };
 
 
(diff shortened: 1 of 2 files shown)


andryyy:
Thank you!!

andryyy:
Do you have a PayPal.me address?

andryyy:
Oh, damn, that's really dangerous what you did there.

It breaks every single sync job that contains special chars in excludes. Even the default exclude will break!

I also think I can inject code. You just use "``" to run the command, stuff like the exclude is not at all escaped. O_O

PR #2577, opened 2019-05-05
Update imapsync

hunter-nl:
- The imapsync_cron.pl script changed to solve users expectation of the provided parameters. Processing of custom parameters field changed.
Example Custom parameters: --addheader --resyncflags --f1f2 "Deleted Messages"=Trash
Solve issues: #2534, #2017

- Removed second unlock file statement
- Changed the setting "is_running" status. There was a possibility to stuck into running state when an exception occurred.
- Updated imapsync script to version 1.937 and added additional perl modules to dovecot Dockerfile
- Changed setting and clear "is_running" state.
- Added noreleasecheck parameter when imapsync executed.


data/Dockerfiles/dovecot/imapsync_cron.pl
```diff
@@ -19,11 +19,20 @@
 }
 
 sub qqw($) {
-  my @values = split('(?=--)', $_[0]);
+  my @params = ();
+  my @values = split(/(?=--)/, $_[0]);
   foreach my $val (@values) {
+    my @tmpparam = split(/ /, $val, 2);
+    foreach my $tmpval (@tmpparam) {
```
andryyy, line 26: This is exactly the Perl magic I was trying to implement for like 2 hours now... I am not good at Perl.

I am not good at Python either. Or PHP. Perl is just worst.

hunter-nl: Perl is also not my favourite. I had to remove the old dust on my knowledge. ;-)

Adorfer:
i do not know if it's related, but the frequent (error) message is a bit annoying: 

"PID file is  ( to change it use --pidfile filepath ; to avoid it use --pidfile "" )
Could not open  for writing. Check permissions or disk space."

hunter-nl:
This warning message was also with the previous versions of imapsync. You can ignore this warning.

PR #2596, opened 2019-05-11
Faster backup and restore

wucke13:
This is my proposal to speed up the backup and restore process. Any feedback is welcome :smile:

<details>
   <summary>Questions already answered</summary> 
### What compression to choose?
I considered `gzip`, `xz` and `bzip2`. I chose `bzip2`, because it is common enough to have its own flag in gnutar ( `-j`) and seems to offer a good compromise between compression time, decompression time and compression ratio, but I'm open to hear your opinion on this topic.

### What Tape Archiver to use?
Most people will opt for gnutar, because they are used to it. And its highly available. Somehow I made the discovery, that gnutar was by far slower in comparision to schily tar, when only creating tar archives (without compression at all). However, this has to be investigated more, a friend of mine had contradictory findings when comparing the two on his system. This seems to be a second order priority, right now the big margin in speed can be made using a parallel, fast compression algo, IMHO.

Please share your thoughts on this.

### Update 1

After some testing, I came to the conclusion that there is no measurable benefit in using schily tar, performance wise. So probably gnutar will do fine.

</details>
<details> 
  <summary>Measurements of questionable accuracy</summary> 

### Update 2

Did some initial measurement using the current solution:
```
1.66s user 1.54s system 0% cpu 21:55.73 total
total 9.6G
517 backup_crypt.tar.gz
2.1M backup_mysql.gz
1.4K backup_postfix.tar.gz
9.9M backup_redis.tar.gz
3.8M backup_rspamd.tar.gz
9.6G backup_vmail.tar.gz
4.2K mailcow.conf
```

### Update 3

Using `lbzip2`, I got better results:
```
1.92s user 1.54s system 0% cpu 10:23.68 total
total 9.5G
1.2K backup_postfix.tar.bz2
9.2M backup_redis.tar.bz2
3.3M backup_rspamd.tar.bz2
9.5G backup_vmail.tar.bz2
4.2K mailcow.conf
```

### Update 4

Using `lz4`, I got more results:
```
1.67s user 1.58s system 0% cpu 5:30.03 total
total 13G

1.2K backup_postfix.tar.bz2
9.2M backup_redis.tar.bz2
3.3M backup_rspamd.tar.bz2
9.5G backup_vmail.tar.bz2
4.2K mailcow.conf
```

### Update 5
Obviously some volumes where missing in the above tests. I'm going to focus on `vmail` only for now, as it is probably the biggest volume for most people.

</details>

### Benchmark based on vmail
|compressor|time result|file size|
|---|---|---|
|`zstd -T0`|4:18.01|9.4G|
|none (only tar)|4:52.28|14G|
|`lz4 -2`|4:57.72|13G|
| `lz4`| 5:11.35 |13G|
|`pigz`| 6:00.88 |9.6G|
|`lbzip2`|10:09.34|9.5G|
|`zstd -T0 --adapt`|24:50.76|9.1G|
|`pixz`| I stopped it after 35 minutes|7.7G|

### Worth noting

+ omitting `--verbose` on tar gives a small performance boost
+ `lz4` is almost as fast as no compressor at all both for compression and decompression
+ ~to get 100% cpu usage with `lz4`, your storage will have to be extremely fast~
+ `lz4` is really fast, but does not support multi threading.
+ `lbzip2` gives me 100% load on all cores
+ `pixz` with default options is stronger than my patience 
+ `zstd --adapt` does not help in our use case
+ `zstd -T0` is my favorite (below 100% cpu usage, fast, good compression)

My explanation regarding `zstd -T0` being faster as plain tar is this: writing costs speed, so if your disk writes more, you have less read performance. So comes that `tar` + `zstd` being faster than plain tar, because more disk IO is available for reading.

### Conclusion

We should use `zstd`.

### Note

Currently, the sql dump is not compressed. The reason is simple: in contrast to all other backup processes, the image used for creating the dumps is not arbitrary. We do use exactly the same image, which also hosts the database in production. This image unfortunately does not come with `zstd`. A way to workaround this would be to run another image after creating the dump, which then compresses it. Taking into account, that most mysql dumps tend to be quiet small, this is of questionable value. However, I'm open to hear your feedback on this.

### DIFF
--- a/data/Dockerfiles/archiver/Dockerfile
+++ b/data/Dockerfiles/archiver/Dockerfile
@@ -1,0 +1,3 @@
+FROM alpine:latest
+RUN apk add --no-cache -X http://dl-cdn.alpinelinux.org/alpine/edge/testing \
+	    tar mariadb-client zstd 
--- a/helper-scripts/backup_and_restore.sh
+++ b/helper-scripts/backup_and_restore.sh
@@ -48,10 +48,15 @@ fi
 BACKUP_LOCATION=$(echo ${BACKUP_LOCATION} | sed 's#/$##')
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 COMPOSE_FILE=${SCRIPT_DIR}/../docker-compose.yml
+COMPRESSOR="zstd -T0 --rsyncable"
+DECOMPRESSOR="zstd -T0"
+ARCHIVE_SUFFIX=".zst"
 echo "Using ${BACKUP_LOCATION} as backup/restore location."
 echo
 source ${SCRIPT_DIR}/../mailcow.conf
 CMPS_PRJ=$(echo $COMPOSE_PROJECT_NAME | tr -cd "[A-Za-z-_]")
+
+docker build  --tag=mailcow-archiver - < ${SCRIPT_DIR}/../data/Dockerfiles/archiver/Dockerfile
 
 function backup() {
   DATE=$(date +"%Y-%m-%d-%H-%M-%S")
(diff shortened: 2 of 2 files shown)


hachre:
Do you have any before-and-after benchmark results?

wucke13:
@hachre That is hard, as running it the second time may profit from caching. I will try to collect some data using a local test instance, and report back though.

wucke13:
@hachre I added a few results in the original PR post, I will add more later.

Just thinking: 
Maybe LZ4 would be nice for compression too. It is fairly wide spread, compression is really fast (except when using **H**igh **C**ompression, which will take much longer but get better compression ratios) and decompression is always really fast no matter if HC was used.
If I do backups, I'm ok if they take some time, but if I need them (especially for something like a webserver), I'd want them to extract as fast as possible. Probably no other compression algo is going to beat LZ4 decompression wise.

### Edit

`zstd` seems to offer really good compression combined with high speed.

wucke13:
After a fresh reboot, with `zstd -T0`:
+ 4:39 (min:sec)
+ 9.5G in total
+ No compression for the `sql` dump
+ CPU usage around 50% on all cores, `zstd` distributed across all cores

no further rebooting (so caching may have helped), with `gzip`, the current way:
+ 21:20 (min:sec)
+ 9.6G in total
+ CPU usage around 35% on all cores, one core >90% with `gzip`

So to me the choice is clear. Let's use `zstd`. On my fairly old machine with low IO the backup is more than four times faster, and the resulting backup is a tad smaller, with neglectable cpu hit.

I did not use `zstd` to compress the `sql` file. See original PR for a reasoning.

hachre:
Sounds very promising!

NicoDreamzZ:
Will this break old backups?

wucke13:
@NicoDreamzZ Yes! I haven't thought about that, but yes, it will break old backups. Though converting from old to new is trivial, something like this could already work (run when `$PWD` is like `$BACKUP_LOCATION/mailcow-2019-05-18-13-13-00` aka when you're in the folder of one backup):
```
for file in *.tar.gz
do
  pigz --decompress < ${file} | zstd -T0 > ${file%%.gz}.zst
done
pigz --decompress < backup_mysql.gz > backup_mysql.sql
```

NicoDreamzZ:
@wucke13 I think it would be nice when the script detects which format the backup is and then restores the correct way. My backups are all saved in a big borg repository and converting all is not possible, and later i maybe forget the snippet you posted :D

wucke13:
@NicoDreamzZ Pardon, but I oppose. Having some sort of decision logic inside of the backup script will make it even more complicated, and more difficult to verify. 

Why would you want to convert *all* your existing backups? As long as it is trivial to convert a backup to the new format, there is no need to convert *anything* until you eventually happen to need a backup. But of course, this conversion snippet should find it's way into the documentation, if this PR gets merged.

Also: nice to hear that you do use borg :+1:

iuri-gg:
will the backups still be `--rsyncable`?

wucke13:
@ocha What do you mean by that? The backups will reside in TAR archives, just as before, the only difference is the compression algorithm used. So if you talk about `rsync`ing these files to some other host, of course that'll work just as before.

iuri-gg:
@wucke13 current solution is using `--rsyncable` flag on gzip which speeds up rsync and saves on bandwidth when changes are small/additive. How important is that feature for overall backup speed assuming one is syncing backup files to a remote server? Below is the explanation of the flag. 

              While  compressing, synchronize the output occasionally based on
              the input.  This increases size by  less  than  1  percent  most
              cases,  but  means that the rsync(1) program can much more effi‐
              ciently synchronize files compressed  with  this  flag.   gunzip
              cannot  tell  the  difference  between a compressed file created
              with this option, and one created without it.

wucke13:
@ocha Ah I see. Thanks for the explanation. Reading the man page of `zstd`, I found this:
```
 --rsyncable
              zstd will periodically synchronize the compression state to make
              the compressed file more rsync-friendly. There is  a  negligible
              impact  to  compression ratio, and the faster compression levels
              will see a small compression speed hit. This  feature  does  not
              work  with  --single-thread.  You  probably don't want to use it
              with long range mode, since it will decrease  the  effectiveness
              of the synchronization points, but your milage may vary.
```
So, `zstd` offers it too. I will adjust the PR accordingly.

wucke13:
ping @andryyy 
Do you have any interest in this?

andryyy:
I don't know. A container for this is too much.

I plan to create a script to create full clones, that can be updated with a command.

wucke13:
Ok. Will you incorporate `zstd` for compression or ditch compression at all?

wucke13:
@andryyy Other than that, would it be feasible to you to have an option in the script to have the container pipe out `tar` archives and send them through any command on the host system? That way one doesn't have to alter the container (as no special tools are needed inside of them) and one can easily change the tool used for compression?

dragoangel:
I♥️Snapshots.
On CentOS7 I using btrfs with btrbk on systems where data can be added and removed before backup actually was done. Unfortunately RHEL discontinued BTRFS support in 8 release. So I went to LVM and Veeam for Linux Free Agent with LVM snapshots. It has wery good speed & compression level vs well cpu usage. In my case bottleneck is 100mbps network between data center and my backup NAS. And good in this all part with backups monitoring. I have nice mail each day with success or failed status, and all details.

marrco:
any update on this?

wucke13:
Closing due to lack of interest from maintainers and totally outdated PR. The issue is still relevant IMO. Hopefully someone else will succeed with this!

PR #2597, opened 2019-05-12
Update update.sh

Javinator9889:
*curl* on some systems is failing due to no URL specified (1.1.1.1) and ISP blocking

*curl* needs an URL for working correctly, so sometimes it is failing as it detects **No Internet connection** but there is. In addition, at some countries some ISP did not update their routers and/or network so the Cloudflare DNS `1.1.1.1` are not working either they are blocked. I suggest using `ping` instead of `curl` with a 3 seconds timeout and pinging to Google DNS (8.8.8.8) instead of Cloudflare ones, as they are universally accepted and used.

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -260,7 +260,7 @@ for option in ${CONFIG_ARRAY[@]}; do
 done
 
 echo -en "Checking internet connection... "
-timeout 3 ping -c 1 8.8.8.8 > /dev/null
+timeout 3 ping -c 1 9.9.9.9 > /dev/null
 if [[ $? != 0 ]]; then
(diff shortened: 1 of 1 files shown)


andryyy:
I have changed it from 8.8.8.8 to 1.1.1.1, because Google is blocked in some countries.

Other ideas? 9.9.9.9?

Javinator9889:
@andryyy maybe just an URL instead of DNS?

andryyy:
Naa, I think quad 9 is fine. :)

Javinator9889:
@andryyy changed ^^

andryyy:
Thanks :)

PR #2608, opened 2019-05-16
Adds only existing domains in table to the filter and removes additional ajax request.

### LINKED ISSUE #2601: Alias domain filtering
The "All domains" Dropdown is not working in "Mail Setup"

Kraeutergarten:
This fixes #2601

### DIFF
--- a/data/web/js/site/mailbox.js
+++ b/data/web/js/site/mailbox.js
@@ -12,7 +12,7 @@ $(document).ready(function() {
       var domains = [];
       
       $.each(self.ft.rows.all, function(i, row){
-        if($.inArray(row.val().domain, domains) === -1) domains.push(row.val().domain);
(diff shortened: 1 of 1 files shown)


evilstiefel:
LGTM, the only issue I guess is that it will fail "ungracefully" if anybody ever decides to rename the row for domains in the table, but that's hardly limited to this PR.

Thanks!

Kraeutergarten:
@evilstiefel 
New commit should fix your concern.

andryyy:
image

I have three domains, no mailboxes. In the mailbox table it returns one domain.

andryyy:
Nah, something else is broken now. I will check it. :) Probably not related.
Edit: Yes, it was indeed not related. :)

evilstiefel:
Well, technically it now uses the data that populates the rows of the table. That means if there are no entries, there should be no options for the filter.

PR #2613, opened 2019-05-17
Allow hostnames for fail2ban whitelist.

### LINKED ISSUE #2609: [Feature] Support hostnames for Fail2ban whitelist.
Since i have a dynamic ip address i can not whitelist it. 
On my hosts where i use fail2ban i can add my dyndnsname to avoid blocking me out.
Is there a chance to get this added?

Kraeutergarten:
Closes #2609

### DIFF
--- a/data/Dockerfiles/netfilter/server.py
+++ b/data/Dockerfiles/netfilter/server.py
@@ -42,10 +42,27 @@ RULES[6] = '([0-9a-f\.:]+) \"GET \/SOGo\/.* HTTP.+\" 403 .+'
 WHITELIST = []
 
 bans = {}
-log = {}
+
 quit_now = False
 lock = Lock()
 
+def log(priority, message):
+  tolog = {}
+  tolog['time'] = int(round(time.time()))
+  tolog['priority'] = priority
+  tolog['message'] = message
+  r.lpush('NETFILTER_LOG', json.dumps(tolog, ensure_ascii=False))
+  print(message)
+  
+def logWarn(message):
+  log('warn', message)
+  
+def logCrit(message):
+  log('crit', message)
+  
+def logInfo(message):
+  log('info', message)
+  
 def refreshF2boptions():
   global f2boptions
   global quit_now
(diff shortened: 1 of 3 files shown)


data/Dockerfiles/netfilter/server.py
```diff
@@ -119,6 +127,19 @@ def ban(address):
   self_network = ipaddress.ip_network(address.decode('ascii'))
   if WHITELIST:
     for wl_key in WHITELIST:
+      if not is_ip_network(wl_key):
+        hostname = wl_key
+        try:
+          wl_key = socket.gethostbyname(hostname)
```
andryyy: You know how long this takes to timeout if the name is invalid? :)

andryyy:
How about blacklists?

Kraeutergarten:
I don't see a usage for blacklists.

andryyy:
Why?

Kraeutergarten:
If i manually blacklist, i normally use the ip of the attacker or i block the network.

data/Dockerfiles/netfilter/server.py
Kraeutergarten: I don't think invalid names are a problem, the bigger problem is a slow responding dns.

Hm, what think you about a local cache? 
I looked into the Fail2ban source, they solve this issue with a cache ^^.

Maybe you have a suggestion?

**Edit:** I found a library where a timeout for DNS requests can be defined. Would this solve the problem? 

**PS:** Is there a reason for using python 2?

Kraeutergarten: Timeout of 1 sec is ok?

ntimo:
But not having this option for blacklists would be quite inconsistent :( And from a user perspective I would expect the black and whitelist to behave in the same way.

Kraeutergarten:
I will try to add it.
How often should the whitelist/blacklist checked for changes? (1-5 minutes?)

andryyy:
If i manually blacklist, i normally use the ip of the attacker or i block the network.

Hm, I see. I agree, but I also agree with @ntimo. Would be nice to not have that inconsistency. :) 
If you find time, it would be amazing to have this implemented. 

The whitelist is currently checked, whenever a rule was triggered. Blacklists apply on startup. Feel free to change that.

data/Dockerfiles/netfilter/server.py
andryyy: Oh, sorry, feel free to change it to whatever you think fits best!

Switching to Py3 would be great. No reason to stick with Py2. :) I think a timeout of 2-3s would be good.

Kraeutergarten:
I would like to do blacklist resolving like the same. Whats do you think about it?

I really don't like how the "logging" was done. 
Hopefully you like my version?

Kraeutergarten:
Now it's done. Maybe someone can help testing?

andryyy:
Nice! I will try to test it asap.

andryyy:
Yes, everything looks so much better now. Please let me use/test it for a few more hours.

Kraeutergarten:
Of course, test as long as you want. 
If you have some suggestions, please let me know.

ssuess:
So...if this was merged, I should be able to add a hostname (or partial hostname, or regex for hostnames) into the Blacklisted networks/hosts box in the Fail2ban config area?

PR #2623, opened 2019-05-19
acme: changed the threshold for certificate renewal

### LINKED ISSUE #2607: add config option to when a cert is renewed
Let’s encrypt sends an email 20 days before a cert expires. 
```
We recommend renewing certificates automatically when they have a third of their
total lifetime left. For Let's Encrypt's current 90-day certificates, that means
renewing 30 days before expiration.
-- https://letsencrypt.org/docs/integration-guide/
```
However mailcow renews certificates only 14 days before the expiry date.
Those 14 days (1209600 seconds) are hardcoded here:
https://github.com/mailcow/mailcow-dockerized/blob/9eebc7c524f46b1916b318b22bc2dbcdca607c41/data/Dockerfiles/acme/docker-entrypoint.sh#L431

I suggest adding an option into a config file (```mailcow.conf``` maybe?). It means the number of days a cert is renewed before its expiry.
This would add a little bit of convenience over editing the above mentioned file and changing the number of seconds.

Arlon1:
The threshold for certificate renewal is changed to 30 days before
expiry date (was 14 days).
fixes #2607

### DIFF
--- a/data/Dockerfiles/acme/docker-entrypoint.sh
+++ b/data/Dockerfiles/acme/docker-entrypoint.sh
@@ -428,8 +428,8 @@ while true; do
 
   if [[ ${SAN_CHANGE} == 0 ]]; then
     # Certificate did not change but could be due for renewal (4 weeks)
-    if ! openssl x509 -checkend 1209600 -noout -in ${ACME_BASE}/cert.pem; then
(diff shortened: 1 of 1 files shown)


ntimo:
@andryyy I think we can merge this, what do you think?

andryyy:
Yes, I will rebuild the images then.

PR #2628, opened 2019-05-20
[Web] Fix showing domain with disabled sender check

foutrelis:
If a mailbox is allowed to send as any address under its domain (+ alias
domains) and the domain itself has no aliases configured, no information
about this fact is shown to the user. That is to say, the "Do not check
sender access for the following domain(s) and its alias domains" field
under mailbox details is empty.

The above is happening because the second GROUP_CONCAT() returns NULL
making the enclosing CONCAT() return NULL as well. Fix this by using
CONCAT_WS() which correctly handles the case of zero domain aliases.

Furthermore, move the IFNULL() to the first GROUP_CONCAT() because
CONCAT_WS() returns an empty string when both GROUP_CONCAT()'s are
NULL. We can be certain that when the first GROUP_CONCAT() is NULL
the second one will be as well, so it's safe to use IFNULL() there.

### DIFF
--- a/data/web/inc/functions.inc.php
+++ b/data/web/inc/functions.inc.php
@@ -676,7 +676,7 @@ function user_get_alias_details($username) {
   while ($row = array_shift($run)) {
     $data['aliases_also_send_as'] = $row['send_as'];
   }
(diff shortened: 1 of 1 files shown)


andryyy:
Thank you! :)

PR #2629, opened 2019-05-20
acme-tiny with python3

christianbur:
Uses acme-tiny with python3, because python2 is End of Life on January 1, 2020. 
Unfortunately I could only test the start of the container, because I am using acme.sh.


andryyy:
I will check it. Thanks!

andryyy:
I will rebuild the images (including PHP) very soon.

PR #2342, opened 2019-02-17
Refactoring dockerapi and preparation for Python3

christianbur:
I wanted to rewrite the dockerapi to python3, but unfortunately there is a bug in the method 'exec_run' of docker-py https://github.com/mailcow/mailcow-dockerized/pull/2297 (in my opinion), for which I can't find a workaround at the moment. The PR contains my changes so far, I think the code is better readable. As far as I could test (not all methods are used by Mailcow) there were no bugs, but this should be checked again. When I have solved the problem with 'exec_run' I will create a new PR with Pyton3.

PR #2630, opened 2019-05-20
dockerapi with python3

christianbur:
Uses dockerapi with python3, because python2 is End of Life on January 1, 2020.
This is the second PR, the first one can be found here https://github.com/mailcow/mailcow-dockerized/pull/2342

### DIFF
--- a/data/Dockerfiles/dockerapi/Dockerfile
+++ b/data/Dockerfiles/dockerapi/Dockerfile
@@ -1,11 +1,13 @@
 FROM alpine:3.9
 LABEL maintainer "Andre Peters <andre.peters@servercow.de>"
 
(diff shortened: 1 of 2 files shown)


Kraeutergarten:
Do we need the pyOpenSSL?
 I think you don't use it anymore.

andryyy:
I wonder when we actually needed it. Edit: I'm blind.

Yes, we can probably remove it.

christianbur:
pyOpenSSL is not really needed anymore, because I create the certificate file directly on the cli with openssl. 
In the old script it was needed to create the certifiact with python-openssl. 

I've updated the Dockerfile.

andryyy:
Thanks. Sorry guys, need a little bit more time. :-(

andryyy:
I will rebuild the image :)

PR #2631, opened 2019-05-20
Update phpfpm baseimage from alpine 3.8 to 3.9

christianbur:
Since I didn't find a reason why Alpine 3.8 is still used here, I created a PR with Alpine 3.9, because Alpine 3.9 is also used in this project.


andryyy:
You didn’t just change the label without any tests, right?

Did it build? Do all modules compile and run fine?

christianbur:
I tested it locally, the mailcow wegui and sogo runs. However, I still don't know the project enough to know all the dependencies.

andryyy:
No problem. Thank you! Let's merge and rebuild, I cannot seem to find a bug with Alpine 3.9, too.

PR #2638, opened 2019-05-23
[Updater] First pull images then stop containers

### LINKED ISSUE #2637: Update script: Pull images before stopping containers
Currently the update.sh script will stop all containers before pulling the new images, as seen [here](https://github.com/mailcow/mailcow-dockerized/blob/master/update.sh#L295) and [here](https://github.com/mailcow/mailcow-dockerized/blob/master/update.sh#L351).

Wouldn't it be better to do it the other way around to minimize downtime during updates? Pull images and then stop the containers.

The separate stopping routine could even be skipped entirely if the files that are being changed do not interfere with the running instances.

ntimo:
Closes https://github.com/mailcow/mailcow-dockerized/issues/2637

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -292,13 +292,12 @@ if [[ ! "$response" =~ ^([yY][eE][sS]|[yY])+$ ]]; then
   exit 0
 fi
 
-echo -e "\e[32mUpdate docker-compose.yml...\e[0m"
-git fetch origin #${BRANCH}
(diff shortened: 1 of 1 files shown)


andryyy:
This ignores every potential problem discussed in the issue. :/

ntimo:
@andryy I am going to fix that. :)

ntimo:
@andryyy

PR #2647, opened 2019-05-26
[GitHub Templates] Added funding template

ntimo:
Added funding configuration for the new GitHub Support button.

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,5 @@
+# These are supported funding model platforms
+
+github: [andryyy]
+custom: https://liberapay.com/mailcow
+custom: https://www.paypal.com/cgi-bin/webscr?cmd=_s-xclick&hosted_button_id=JWBSYHF4SMC68


patcon:
Sorry, may I ask why this was removed @andryyy? :)

I am reflecting on an initiative/tool to help people know funding options for opensource/free business tools that they use, ideally working only from source code to check, and I was just coming to Mailcow to add (at the very least) these two:
https://www.servercow.de/mailcow#support
https://www.servercow.de/mailcow#sal

But I see that funding.yml was removed for some reason. Would love to know why, as it's the closest thing we have to a standard <3

andryyy:
The tax office. :(

They hunt you down and take your firstborn. It is easier for me this (=SAL) way. :)

patcon:
OK, totally! Def don't want you to be troubled :)

Would we be able to bring it back and only link to the SAL/support tiers, or is just the using of the FUNDING.yml file what makes you nervous?

andryyy:
The links you posted are totally fine. :)

andryyy:
It is not that taxes were not paid, but it is hard to actually tax them, as I needed to see them as a kind of service I provided and create invoices etc. ... it is much easier the way it is now.

patcon:
ok! will submit a PR to bring it back with those safe ones :) (correct me if i'm wrong)

PR #2659, opened 2019-05-30
[ClamAV] Update to 0.101.2

ntimo:


andryyy:
Thanks, but you don't need to bump the image version next time. :)

PR #2693, opened 2019-06-08
[Postfix] Retry for 5 days instead of 1

dofl:
I was looking into creating a backup mx server for a high availability mailcow setup. It seems that this is not easily done. While researching to find out how long an average SMTP server keeps trying to send to a server that is down I found that RFC 5321 advises at least 4 to 5 days. Mailcow has a custom setup of 1 day, which is very short. The user will be unaware for 5 days that his mail has not been delivered, which can be negative. But I still would like to follow the advice of the RFC.

RFC 5321, in section 4.5.4.1, has this to say:
Retries continue until the message is transmitted or the sender  up;  the give-up time generally needs to be at least 4-5 days.  It MAY be appropriate to set a shorter maximum number of retries for non-delivery notifications and equivalent error messages than for standard messages. 

Postfix default is also 5 days: http://www.postfix.org/postconf.5.html

https://tools.ietf.org/html/rfc5321#section-4.5.4


mkuron:
We should certainly be RFC-compliant, thanks for checking up on that. @andryyy, was there a specific reason for the shorter time?

Does Postfix send these “your message has been delayed, will keep retrying for five days“ notifications so that the sender knows that his message may take longer than expected to be delivered, @dofl? If not, we should probably add that after one day.

andryyy:
I think it was just a personal preference. Should be merged, si.

dofl:
@mkuron I've added the delay_warning_time to the pull request. I've set it to 4 hours. When I send an e-mail to someone or a company it could have a certain priority. A day might be too long for me to know that the receiving e-mail server hasn't been working for 4 hours, which is quite some downtime already in my opinion.

There is also a setting called confirm_delay_cleared (http://www.postfix.org/postconf.5.html#confirm_delay_cleared), but according to the Postfix this can lead to a sudden burst of notifications at the end of a prolonged network outage.

foutrelis:
I'm not sure `delay_warning_time` does what we would expect it to do. It sends the delay warning to `$delay_notice_recipient` (default: postmaster) and not the sender. Does this mean the warning simply gets lost?

foutrelis:
I must have confused delay warnings sent to the sender with delay notices sent to the postmaster if `notify_classes` includes the `delay` error class. Sorry for the noise, it most likely works as expected after all. :rabbit:

Adorfer:
delay-warning of 4 hours may create a lot of requests if customers mailserves are down for maintenance. 
(and clients do not read what's inside the bounce, often they even tend to complain about "got spam". but that's another story)

in other words:  
1) would recommend to make both values configurable (either via ui or at least via configfile)
2) default for delay warning 12h would be more suitable.

andryyy:
You can change the main.cf

raoulbhatia:
👍 for `delay_warning_time = 4h` - this is the defacto standard.
Also, I cannot think of a lot of people who want to wait until the next (work) day to know that email delivery failed.

dofl:
Default = 4 hours.

If the e-mail is important a notification after 12 hours is a very long time. No mail server should be offline for this period.

The notification is only send once. A lot of requests will not be generated as each e-mail will only generate one notification and I am not expecting that a mail server like gmail or outlook will ever be offline for 4 hours.

PR #2594, opened 2019-05-11
Default quota for mailbox

feldsam:
Signed-off-by: Kristián Feldsam <feldsam@gmail.com>

PR #2694, opened 2019-06-09
Increase db version for merge #2594

AikoAppeldorn:
With Merge #2594 the database init script was adapted, but the version was not increased.
Existing Mailcow installations will not update the database.

Every mailcow instance that has been updated causes an SQL error when opening the domain overview.


andryyy:
Thanks

PR #2702, opened 2019-06-12
[Rspamd] Enable history compression

ntimo:
This PR enables the history compression for Rspamd as documented here: 
https://www.rspamd.com/doc/modules/history_redis.html


andryyy:
That's default. 

https://github.com/rspamd/rspamd/blob/master/src/plugins/lua/history_redis.lua#L26

PR #2737, opened 2019-06-28
Omits the verbose option for gzip in the backup script

jeroenj:
The verbose option dumps a progress percentage to `STDERR`. I'm piping `STDOUT` to `/dev/null` in my backup cron job but I'd like to keep `STDERR` in case something goes wrong. However because the progress gets dumped to `STDERR` by gzip it means it will always look like something went _wrong_.


andryyy:
Yes, I think we should omit the v flag. :) Thanks.

PR #2745, opened 2019-06-29
Create sieve filter to drop mail if user has catch-all active

aghodsi:
Hi.

The purpose of temporary email aliases is that they don't bug you anymore after its time period has expired, I noticed however that if the user has a catch-all active for a domain to his account, this will of course not work and emails to the generated email addresses aren't dropped but will be delivered to the user.

This code checks, after creating the temp email, if the user has a catch-all that is redirected to his email address. If this is the case, the code will create a pre-filter sieve to drop any emails to this temp address.

### DIFF
--- a/data/web/inc/functions.mailbox.inc.php
+++ b/data/web/inc/functions.mailbox.inc.php
@@ -65,8 +65,8 @@ function mailbox($_action, $_type, $_data = null, $_extra = null) {
           if($count > 0){
             $stmt = $pdo->prepare("INSERT INTO `sieve_filters` (`username`, `script_data`, `script_desc`, `script_name`, `filter_type`)
             VALUES (:username, :script_data, :script_desc, :script_name, :filter_type)");
-            $prefilter_tempEmail = 'if address :is "To" "' .  $temp_email . '" {
-              discard;
+            $prefilter_tempEmail = 'require ["reject"];if address :is "To" "' .  $temp_email . '" {
+              reject "Not Accepted";
             }';
             $stmt->execute(array(
               ':username' => $username,
--- a/data/web/inc/functions.mailbox.inc.php
+++ b/data/web/inc/functions.mailbox.inc.php
@@ -71,8 +71,8 @@ function mailbox($_action, $_type, $_data = null, $_extra = null) {
             $stmt->execute(array(
               ':username' => $username,
               ':script_data' => $prefilter_tempEmail,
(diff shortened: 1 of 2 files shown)


andryyy:
That's actually a nice idea. I will test it. :) Thanks!!

aghodsi:
Just realised that there is something missing, this creates the filter instantly and drops the emails straight away. It needs to create the filters after the expiration time... will try to figure out a fix for that

andryyy:
Or just don't use a catch all and spam aliases. That's totally expected after all.

aghodsi:
True, closing this PR till I find something useful/more clean :)

andryyy:
I really like the sieve approach, but another idea would be to change the event, that triggers the deletion in https://github.com/mailcow/mailcow-dockerized/blob/master/data/web/inc/init_db.inc.php#L968

fbcd4f50ea5350f051450323b5abfd0b:
Another Idea:
Have a fixed and a generated part in the temporary aliases. That way we could have a regex filter that matches the fixed part and discards the messages if the mailbox doesn‘t exist.

andryyy:
I think it is easier to just add them to the aliases (can be done through the event from above) and send them to null@localhost when they expire. You could even add a public/private comment like "expired alias by xy". :)

You'd also see a list of expired temp aliases this way. 

PS: Why is the iPhone auto correction so fucked up?

aghodsi:
Hey @andryyy,

Took you up on that idea of using the trigger :) Now, the SIEVE will be inactive when added and it will get activated with the event. Any thoughts?

mkuron:
Silently dropping messages is generally a bad idea. Better reject it during the SMTP transaction. I think Sieve has a "reject" or "refuse" action for that purpose.

aghodsi:
Hey @mkuron, I’m quite new to this, could you explain to me why it is a bad idea? The change you propose is a simple one, would like to understand why though :)

mkuron:
If your server accepts a message, the sender can reasonably assume that you have received the message. So silently discarding messages creates a legal liability. Also senders of newsletters etc. will eventually remove you from their list if your message bounces, so rejecting helps reduce traffic.

aghodsi:
Makes perfect sense, thanks, updated!

mkuron:
Could you please check whether the sieve reject action actually works and check what lines get printed to the Postfix log?

PR #2747, opened 2019-06-30
Added mta-sts validation for Postfix

ntimo:
Added mta-sts validation for Postfix using https://pypi.org/project/postfix-mta-sts-resolver/ .

https://tools.ietf.org/html/rfc8461


vsund:
Hey @ntimo, any reason this was closed instead of merged?
Anything I could help with? :)

ntimo:
@vsund This has been closed because adding another container to the project just for mta sts, is a bit too much. This should better be integrated directly into Postfix.

extremeshok:
@ntimo , technically its a completely separate daemon , so it should have its own container. It has its own requirements and dependencies, opens a port etc. It should also be connected to redis to provide persistence, rather than using a cache file.

And on the same merit, why was olefy-mailcow given its own container and not integrated into postfix ?

extremeshok:
could this please be re-opened ?

@andryyy what are your thoughts on this as part of postfix or seperate docker ?

mkuron:
I think it would be fine to have in a separate container. It's a TCP map for Postfix. Olefy is a similar situation, or ClamAV -- these two are TCP services for rspamd.

andryyy:
STS is bs for me at this point. Check the downsides of STS in the RFC. For me it is only useful with DNSSEC (I could be wrong with this).

I won't discuss why oletools has a higher value, it should be obvious. 

Just force TLS with all servers and make some special exceptions with TLS policy maps (don't know, but should work). Or skip the exceptions part and don't mail with them at all. Ooooor ... create TLS policies for Google etc. manually.

**This is incomplete** - how about our own STS policy?

I will not merge this. We can leave it open to not forget about it.

DatAres37:
It's actually a alternative to DANE as far as I understand:
> The primary motivation of MTA-STS is to provide a mechanism for domains to ensure transport security even when deploying DNSSEC is undesirable or impractical

Could we have an option to place a /.well-known/mta-sts.txt file or is there an easy way to do this at the moment?

mkuron:
I believe MTA-STS just moves the trust from DNSSEC to HTTPS. So it's an alternative to DANE and probably has similar advantages and disadvantages. I'm not sure it's necessary though: can anyone find a production mail server on the internet that has MTA-STS enabled but not DANE?

DatAres37:
> I believe MTA-STS just moves the trust from DNSSEC to HTTPS. So it's an alternative to DANE and probably has similar advantages and disadvantages. I'm not sure it's necessary though: can anyone find a production mail server on the internet that has MTA-STS enabled but not DANE?

My domain hoster doesn't support TLSA records, so I would be the person you're looking for I guess :smile:

mkuron:
> My domain hoster doesn't support TLSA records

In that case, you might want to consider switching providers or try to convince them to add the record type. I was referring to other people's servers that publish MTA-STS instead of TLSA, not ones you control yourself.

DatAres37:
> I was referring to other people's servers that publish MTA-STS instead of TLSA, not ones you control yourself.

I know, but if a user here would already use it like that, I assume that I'm not the only one on the internet. MTA-STS is great, because you'd be independent from your hoster and the standard was certainly not added because there isn't any demand. You also have to consider, that it's fairly new if you're asking who's using it at the moment.

Zudem hat DNSSEC auch noch einige Downsides: https://de.wikipedia.org/wiki/Domain_Name_System_Security_Extensions#Sicherheitsrelevante_Schwachstellen_von_DNSSEC
https://www.golem.de/news/imho-dnssec-ist-gescheitert-1506-114940.html

Over all MTA-STS seems more attractive and simpler to setup for me tbh. Just an option to add a mta-sts.txt would be great.

mkuron:
None of that DNSSEC criticism is actually relevant for the security of DANE. DNSSEC works, even though it may be a bit difficult to implement -- but that's mostly your domain registrar's problem.

Unlike MTA-STS, DANE is well established and widely implemented. There's no point to this pull request until you find a server that publishes a MTA-STS policy and no TLSA records.
We could already start publishing a MTA-STS policy -- but again, this makes little sense until you find a server that doesn't validate DANE but validates MTA-TLS.

This sounds like a chicken-and-egg problem, but it's not. DANE provides the same protection as MTA-STS and is widely supported already, so MTA-STS might never gain enough momentum to be worth supporting.

andryyy:
That's well said. :)

FingerlessGlov3s:
> > My domain hoster doesn't support TLSA records
> 
> In that case, you might want to consider switching providers or try to convince them to add the record type. I was referring to other people's servers that publish MTA-STS instead of TLSA, not ones you control yourself.

@DatAres37 One of my domains the TLD (tk.) doesn't support DNSSEC, I'm slowly moving to a new domain where the TLD (pm.) does support DNSSEC.

google,yahoo,outlook,protonmail,gmx.com,mail.com probably a lot more have setup MTA-STS.

Funny enough outlook.com has it but not microsoft.com 😂

Only 129 tlds out of 1527 (8%) doesn't support DNSSEC
http://stats.research.icann.org/dns/tld_report/

mkuron:
> One of my domains the TLD (tk.) doesn't support DNSSEC

Fair point, but in 2019 that just seems like negligence of the registry.

> google,yahoo,outlook,protonmail,gmx.com,mail.com probably a lot more have setup MTA-STS.

They publish a policy (though most of them only with `mode: testing`. We don't know whether they perform validation too. All of them do support DANE, so Mailcow doesn't need MTA-STS support to send encrypted messages with them.

FingerlessGlov3s:
> > One of my domains the TLD (tk.) doesn't support DNSSEC
> 
> Fair point, but in 2019 that just seems like negligence of the registry.
>
I think it is negligence, but unfortunately that's the case.


> > google,yahoo,outlook,protonmail,gmx.com,mail.com probably a lot more have setup MTA-STS.
> 
> They publish a policy (though most of them only with `mode: testing`. We don't know whether they perform validation too. All of them do support DANE, so Mailcow doesn't need MTA-STS support to send encrypted messages with them.

In my mind we should implement it or maybe make it optional in the config or a guide on the documentation. It is a standard and ideally you should support as many as you can, when it comes to security in my eyes anyway. The major email providers have DANE, but there maybe other email servers where MTA-STS is setup but they haven't enabled DNSSEC or they're DNS resolver for their mail server doesn't support DNSSEC.

They do recommend you put it in testing for a while anyway when you set it up, along with the TLS-RPT

I might implement this either way on my Mailcow install, just thought I'd contribute to the discussion.
@ntimo did you leave it setup on your mailcow instance?

andryyy:
I think it's useless. If you _need_ to talk TLS with a server, create a policy or enforce TLS per mailbox. Or just validate via DANE.

If you want to use MTA-STS, you should also deliver a policy per domain besides reading them.

It is just not worth it imo. Some bigger providers have implemented it half-assed to look modern and up-to-date. I think they don't really care anyway. Cannot blame them. :) I see there are at least _some_ advantages, but it is just not enough.

FingerlessGlov3s:
You raise a good point, I'll leave it for now, unless it gains some traction.

mkuron:
> It is a standard and ideally you should support as many as you can, when it comes to security in my eyes anyway. The major email providers have DANE, but there maybe other email servers where MTA-STS is setup but they haven't enabled DNSSEC or they're DNS resolver for their mail server doesn't support DNSSEC.

Right now, it won‘t give you encrypted connections with any more mail servers than DANE already does (well, except for the case of your .tk domain). Adding features to Mailcow today requires maintaining features in the future, so one always needs to consider the maintenance effort vs. how useful the feature is. Usually that‘s strongly in favor of the new feature, but in the present case, the feature seems almost useless.

In your .tk scenario, you don‘t even need this pull request. You can still publish a policy for incoming connections via a static file (which you don‘t currently do) and validate outgoing connections via DANE.

dragoangel:
Actually MTA-STS is not reject DANE must be in place, if somebody can MITM & modify DNS response it can block resolving of mta-sts.example.com and tatadam - MTA-STS will not work. But if response from MTA-STS has big max_age: and previously some mailservers cache them - then it will be save you from real mitm. For me MTA-STS looks like HSTS in HTTPS. The second part of MTA-STS says that SSL cert of Postfix must be same as MX record itself which is help to authorize all stuff. STS says that TLS1.2 is required. And in bonus you will be notified with report in json format that store success and reject counters =). And as mentioned before for example my TLD .ua looking like support DNSSEC, but on reality it not supported now by any of Domain Registrators that provides .ua TLD

Snawoot:
> I'm not sure it's necessary though: can anyone find a production mail server on the internet that has MTA-STS enabled but not DANE?

Just short list of such examples:

gmail.com
outlook.com
yahoo.com

whalehub:
https://starttls-everywhere.org/

@andryyy Can this be reconsidered by now?

andryyy:
Voting against it, sorry.

mkuron:
MTA-STS validation adds significant extra complexity. DANE is the RFC standard and supported natively by Postfix. We could however consider incorporating the [STARTTLS Everywhere policy](https://starttls-everywhere.org/policy-list/). Apparently you can automatically convert it to a Postfix TLS map file and thus force TLS for hosts like Gmail and Yahoo.

Snawoot:
> DANE is the RFC standard and supported natively by Postfix.

MTA-STS is also an accepted RFC standard.

>  We could however consider incorporating the STARTTLS Everywhere policy.

This project is basically dead. By the way, they were encouraging MTA-STS adoption and require MTA-STS policy to be in place for inclusion into list.

At this moment by default S-E policy list delivers no value because all domains in list in testing mode, so it doesn't enforce anything and generates empty policy list. There is some sort of "override button" (EFForg/starttls-policy-cli#24, this one is mine, BTW) which considers "testing" domains as "enforce" ones, but it may be undesired for environments intolerant to delivery disruption.

> Apparently you can automatically convert it to a Postfix TLS map file and thus force TLS for hosts like Gmail and Yahoo.

Static files are not manageable and not scalable.

~For example, recently gmail.com STS policy changed its MX servers from `gmail-smtp-in.l.google.com, *.gmail-smtp-in.l.google.com` to `aspmx.l.google.com, *.aspmx.l.google.com` and in static policy case it definitely would cause complete delivery disruption.~ My bad, confused `gmail.com` and `google.com`. Anyway, MX update still may happen.

Another problem with static policy lists for well-known domains is their bias in favour of those well-known mail services and negligence to operators of smaller mail servers, just like mailcow. Smaller services can't discover policy of each other this way.

Last (but not least) issue with static policy lists is inability of mailserver operator to downgrade such policy learned somehow by other domains. On the other hand, MTA-STS provides a way for authenticated denial of policy existence, allowing to disable it any time. With static files, domain operator has practically no influence on already distributed policies, making STS deployment more risky or even surprising if such policies were inferred by other domain operators voluntarily.

Static policies is not even remotely a substitution for MTA-STS and even can be harmful.

mkuron:
> Static files are not manageable and not scalable.

They provide tools to automatically pull that file, it has an expiration date, etc. Sure, lists like that don‘t scale well and would have to be limited to big providers, but smaller ones could just use DANE (and often do that rather than MTA-STS).

> This project is basically dead.

That wasn‘t obvious to me from their website. There are just too many different ways to enforce encryption now (DANE, MTA-STS, STARTTLS everywhere)... I actually still don‘t understand why some big providers like Gmail don‘t support DANE.

Snawoot:
> That wasn‘t obvious too me from their website.

Project exists more than 5 years, still only "testing", last year almost without meaningful commits (including site frontend and backend as well). Untill late 2018 policy cli tool even wasn't generating syntactically correct postfix map files, no one were even noticing. Since release of RFC8461 mostly focused on motivation of people to implement MTA-STS. More insights available on dev mailing lists, but I'm not sure if they are public.

> I actually still don‘t understand why some big providers like Gmail don‘t support DANE.

DANE is awesome, but DNSSEC didn't took off for various reasons (and there is pretty large list of DNSSEC drawbacks). Sad story, but facts are facts - [share of DNSSEC-enabled domains is declining](https://blog.apnic.net/2018/02/26/peak-dnssec/). I believe it's the reason why MTA-STS appeared in first place. MTA-STS RFC has statement on this subject:

```
   The primary motivation of MTA-STS is to provide a mechanism for
   domains to ensure transport security even when deploying DNSSEC is
   undesirable or impractical.  However, MTA-STS is designed not to
   interfere with DANE deployments when the two overlap; in particular,
   senders who implement MTA-STS validation MUST NOT allow MTA-STS
   Policy validation to override a failing DANE validation.
```

Seemingly there is barely one way out of all options to enforce encryption which has a chance to become widespread.

mkuron:
> DNSSEC didn't took off for various reasons

That may be, but it isn‘t going away anytime soon. It‘s still the only way to sign your DNS zones, and that‘s the right thing to do. I‘m actually surprised Gmail.com isn‘t signed; I‘m pretty sure it was previously signed. Also, MTA-STS kind of depends on DNSSEC too.

> Seemingly there is barely one way out of all options to enforce encryption which has a chance to become widespread.

That’s certainly true. Perhaps the conclusion is also: email transport encryption isn‘t easy. Seemingly Gmail etc. don‘t care enough about it. Otherwise they would be supporting both the long-time standard of DANE and their new MTA-STS standard.

ghost:
@mkuron @Snawoot It looks like STARTTLS Everywhere is officially dead now:

https://www.eff.org/deeplinks/2020/04/winding-down-starttls-everywhere-project-and-future-secure-email
https://www.eff.org/deeplinks/2020/04/technical-deep-dive-winding-down-starttls-policy-list

Snawoot:
@whalehub Yeah, thanks! Actually I'm glad now it's clear and there is no need to speculate about status of that project like I did [before](https://github.com/mailcow/mailcow-dockerized/pull/2747#issuecomment-568129981). Now only two options remain for secure email exchange between public mail domains: extinct (DANE) and questionable (MTA-STS).

FingerlessGlov3s:
Ideally you would deploy both wouldn't you? As not all TLDs or Registrars support DNSSEC? So MTA-STS as a fallback would be good?

Snawoot:
@FingerlessGlov3s Yes, that's what I did for my email domains. Even if you have DANE for your email domain, you will only benefit from ability to discover and apply MTA-STS policies of other domains.

FingerlessGlov3s:
I've got both setup now and used https://aykevl.nl/apps/mta-sts/, to make sure it's correct. All green ticks 👍

PR #2755, opened 2019-07-03
Update docker-entrypoint.sh

gmarke:
get mysql ID only for current project


andryyy:
Does it look good for you, @mkuron? I am not sure this will help with multi-cow installations though.

mkuron:
It's certainly a necessary change, so let's merge it. If other changes turn out to be necessary for multicows, they can be made later.

andryyy:
```
php-fpm-mailcow_1    | .[] | {project: .Config.Labels[com.docker.compose.project], name: .Config.Labels[com.docker.compose.service], id: .Id}
php-fpm-mailcow_1    | jq: 2 compile errors
```

Why don't you even test such a change?

PR #2768, opened 2019-07-08
fix(imapsync): Delete from source when completed

djskinner:
Change the argument from --delete to --delete1

I noticed this functionality wasn't working and discovered that --delete does not exist (https://imapsync.lamiral.info/README). I believe it should be --delete1

**Update**

Turns out this is not working for me for other unknown reasons. `--delete` and `--delete1` are aliases.

Still, consider merging this change as it's recommended to use `--delete1`.

From the release notes:

> Usability: Options --delete1 and --delete are now aliases. Option --delete1 is preferable over --delete (--delete is still supported).

### DIFF
--- a/data/Dockerfiles/dovecot/imapsync_cron.pl
+++ b/data/Dockerfiles/dovecot/imapsync_cron.pl
@@ -133,7 +133,7 @@ while ($row = $sth->fetchrow_arrayref()) {
   ($maxbytespersecond eq "0" ? () : ('--maxbytespersecond', $maxbytespersecond)),
   ($delete2duplicates ne "1" ? () : ('--delete2duplicates')),
   ($subscribeall  ne "1" ? () : ('--subscribeall')),
-  ($delete1 ne "1" ? () : ('--delete')),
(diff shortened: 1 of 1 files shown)


djskinner:
After investigating this further I'm still not seeing the emails being deleted from host1 even when supplying `--delete1` as a custom flag. From the imapsync logs it looks like the delete/expunge is happening but the emails are never deleted from host1 (even after using 'compact' in the mail client etc.).

So not currently sure it's a bug. Maybe `--delete` does the same as `--delete1`? There's no mention of `--delete` in the docs although it is mentioned in the latest release [release notes](https://imapsync.lamiral.info/S/news.shtml). Going to try upgrading imapsync to latest.

andryyy:
Any results? :)

Should indeed be deleted1. 

I think we already use the latest imapsync (1.937).

djskinner:
I tried 1.945 and still experienced the same issue, which I assume is an issue with my `host1`. I ended up working around after not being able to find the root cause. No regressions to report in [1.945](https://github.com/imapsync/imapsync/releases/tag/imapsync-1.945).

PR #2787, opened 2019-07-16
Remove score from R_DKIM_PERMFAIL

### LINKED ISSUE #2785: R_DKIM_PERMFAIL
If I recieve mails from a specific domain they always get headers added to the mails through rpsamd.

R_DKIM_PERMFAIL (10) [example.org:s=dkim]

I checked rspamd config and all I could do is to set this to another value, but that doesn't really solve the problem.

Strange thing is, the senders domain doesn't even use dkim, so I wonder why this error happens and the score is so high, as many (many many many) providers don't use dkim.

**System information and quick debugging**
General logs:
- I don't think they're needed here, if you need them I'll provide later.

Further information (where applicable):
 - Your OS (is Apparmor or SELinux active?)
ubuntu 18.04 (apparmor yes i guess)
 - Your virtualization technology (KVM/QEMU, Xen, VMware, VirtualBox etc.)
lxc
 - Your server/VM specifications (Memory, CPU Cores)
4GB and 4 cores
 - Don't try to run mailcow on a Synology or QNAP NAS, do you?
hell no
 - Docker and Docker Compose versions:
Docker version 18.09.1, build 4c52b90
docker-compose version 1.24.1, build 4667896b
 - Output of `git diff origin/master`, any other changes to the code?
```
diff --git a/data/conf/dovecot/dovecot.conf b/data/conf/dovecot/dovecot.conf
index cf6b5d12..fea6544c 100644
--- a/data/conf/dovecot/dovecot.conf
+++ b/data/conf/dovecot/dovecot.conf
@@ -32,6 +32,7 @@ ssl_min_protocol = TLSv1
 
 ssl_prefer_server_ciphers = yes
 ssl_cipher_list = ALL:!ADH:!LOW:!SSLv2:!SSLv3:!EXP:!aNULL:!eNULL:!3DES:!MD5:!PSK:!DSS:!RC4:!SEED:!IDEA:+HIGH:+MEDIUM
+#ssl_cipher_list = ALL:!LOW:!SSLv2:!EXP:!aNULL:!ECDH+AESGCM:DH+AESGCM:ECDH+AES256:DH+AES256:ECDH+AES128:DH+AES:ECDH+3DES:DH+3DES:RSA+AESGCM:RSA+AES:RSA+3DES:!aNULL:!MD5
 
 # Default in Dovecot 2.3
 ssl_options = no_compression
diff --git a/data/conf/nginx/site.conf b/data/conf/nginx/site.conf
index ccb94fa0..9eac0d7a 100644
--- a/data/conf/nginx/site.conf
+++ b/data/conf/nginx/site.conf
@@ -7,6 +7,13 @@ map $http_x_forwarded_proto $client_req_scheme {
      https https;
 }
 
+server {
+  listen 8080 default_server;
+  listen [::]:80 default_server;
+  include /etc/nginx/conf.d/server_name.active;
+  return 301 https://$host$uri$is_args$args;
+}
+
 server {
   include /etc/nginx/mime.types;
   charset utf-8;
```

mkuron:
This error happens when there is no public key in DNS for that selector. According to RFC 6376, such a signature should silently be ignored. If no other signature is present, the message should be treated as unsigned.

> In the following description, text reading "return status
   (explanation)" (where "status" is one of "PERMFAIL" or "TEMPFAIL")
   means that the Verifier MUST immediately cease processing that
   signature.  The Verifier SHOULD proceed to the next signature, if one
> is present, and completely ignore the bad signature.

> 3. If the query for the public key fails because the corresponding
       key record does not exist, the Verifier MUST immediately return
       PERMFAIL (no key for signature).

> 5. If the result returned from the query does not adhere to the
       format defined in this specification, the Verifier MUST ignore
       the key record and return PERMFAIL (key syntax error).

Fixes #2785

### DIFF
--- a/data/conf/rspamd/local.d/policies_group.conf
+++ b/data/conf/rspamd/local.d/policies_group.conf
@@ -11,9 +11,6 @@ symbols = {
     "R_DKIM_REJECT" {
         score = 10.0;
     }
-    "R_DKIM_PERMFAIL" {
-        score = 10.0;
-    }
     "DMARC_POLICY_REJECT" {
         weight = 20.0;
         description = "DMARC reject policy";


mkuron:
Note that treating an invalid *SPF* record as spam is fine per RFC 7208 (the SPF RFC):

>  A "permerror" result means the domain's published records could not
   be correctly interpreted.  This signals an error condition that
   definitely requires DNS operator intervention to be resolved.  If the
   message is rejected during the SMTP transaction for this reason, the
   software SHOULD use an SMTP reply code of 550 [...]

patschi:
Just an FYI, as I'm not sure if this makes any differences: rspamd does have `R_DKIM_DNSFAIL` symbol for cases when DKIM fails due to DNS issues: https://github.com/rspamd/rspamd/blob/4cd06657503bf26b46a0e64350599bb4ab8dbb2a/conf/composites.conf#L60-L61

However I'm not finding any more concrete documentation about that.

mkuron:
I don‘t think DNSFAIL exists anymore. PERMFAIL and TEMPFAIL can indicate permanent and temporary DNS failures (among other less common failures).

PR #2788, opened 2019-07-16
Ignore DKIM/SPF/DMARC scores on mailing lists

patschi:
Removing weight and leaving symbol intact makes it easier to debug

### DIFF
--- a/data/conf/rspamd/local.d/composites.conf
+++ b/data/conf/rspamd/local.d/composites.conf
@@ -26,3 +26,9 @@ OLEFY_MACRO {
   score = 20.0;
   policy = "remove_weight";
 }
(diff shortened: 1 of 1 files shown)


mkuron:
I understand why you are suggesting this. However this allows any spammer to bypass these checks by adding a `List-Id` header.  So we are not going to merge it. This problem should really be fixed on the listserver, e.g. by rewriting _From_ addresses. Further details at https://wiki.list.org/DEV/DMARC and https://wiki.list.org/DEV/DKIM.

andryyy:
Yes, indeed. This was my first thought, too. I am not too sure about how often this is used in spam mail. I should have checked it before.

PR #2791, opened 2019-07-16
Fixing duplicated port for profile downloading.

### LINKED ISSUE #2718: Profile download for iOS not possible
I changed the ports in mailcow.conf to 8081 (HTTP) and 1443 (HTTPS). When I am trying  to download the Profile following url is called:

`https://(replaced domain):1443:1443/mobileconfig.php`

**System information:**

 - Ubuntu 18.04 LTS
 - Proxmox
 - 30 GB Ram 8 cores
 - Docker version 18.09.6, build 481bc77
 - docker-compose version 1.24.0, build 0aa59064

mkapra:
Fixes #2718

### DIFF
--- a/data/web/user.php
+++ b/data/web/user.php
@@ -77,7 +77,7 @@ elseif (isset($_SESSION['mailcow_cc_role']) && $_SESSION['mailcow_cc_role'] == '
   $username = $_SESSION['mailcow_cc_username'];
   $mailboxdata = mailbox('get', 'mailbox_details', $username);
 
-  $clientconfigstr = "host=" . urlencode($mailcow_hostname) . "&email=" . urlencode($username) . "&name=" . urlencode($mailboxdata['name']) . "&ui=" . strtok(urlencode($_SERVER['HTTP_HOST']), ':') . "&port=" . urlencode($autodiscover_config['caldav']['port']);
+  $clientconfigstr = "host=" . urlencode($mailcow_hostname) . "&email=" . urlencode($username) . "&name=" . urlencode($mailboxdata['name']) . "&ui=" . urlencode(strtok($_SERVER['HTTP_HOST'], ':')) . "&port=" . urlencode($autodiscover_config['caldav']['port']);
   if ($autodiscover_config['useEASforOutlook'] == 'yes')
   $clientconfigstr .= "&outlookEAS=1";
   if (file_exists('thunderbird-plugins/version.csv')) {


data/web/user.php
```diff
@@ -76,7 +76,10 @@
   $_SESSION['return_to'] = $_SERVER['REQUEST_URI'];
   $username = $_SESSION['mailcow_cc_username'];
   $mailboxdata = mailbox('get', 'mailbox_details', $username);
-
+	
+  $mailcow_hostname_array = explode(":", $mailcow_hostname)
+  $mailcow_hostname = $mailcow_hostname_array[0]
+
```
mkuron: That would work, but change it to use the same `strtok` solution that we have in the line below.

mkuron (changes requested):

data/web/user.php
patschi: This wouldn't work, to be correct. The missing semicolons `;` would cause a parse error.

mkapra: @patschi  ah sure! sorry I was in python :D

mkapra:
@mkuron 
I changed the explode to strtok() and rebased the branch so there is only one commit.
Is it okay now? :)

mkuron:
Looks good. Did you test whether it solves your problem?

mkapra:
> Looks good. Did you test whether it solves your problem?

@mkuron How can I use my branch instead of another on my mailserver or do I don't understand you?

mkuron:
Go to your mailcow-dockerized directory and run the following commands:

```
git add remote mkapra https://github.com/mkapra/mailcow-dockerized.git
git fetch mkapra
git checkout mkapra/patch-1
```

mkuron:
And actually your patch won't work. You need to do `strtok` before the `urlencode`. Same goes for the HTTP_HOST. Could you fix it there too?

mkapra:
@mkuron: I fixed your suggestions. But when I try to execute:
`git remote add mkapra https://github.com/mkapra/mailcow-dockerized.git`

I am getting the following error:
`fatal: pathspec 'remote' did not match any files`

Do you know what I am doing wrong?

mkapra:
@mkuron Found the problem. I was only typing the command and not verifying it.
It should be called like this:
`git remote add ...`

mkapra:
@mkuron 
I can  confirm that my fix works! Rebased it again.

Braintelligence:
@mkuron Are you 100% sure about the strtok order? The urlencode was what returned a urlencoded string with a colon that had to be stripped, why should the colon be stripped from the parameter of urlencode instead?

mkapra:
> @mkuron Are you 100% sure about the strtok order? The urlencode was what returned a urlencoded string with a colon that had to be stripped, why should the colon be stripped from the parameter of urlencode instead?

@Braintelligence 
I am cutting at the ":" sign. After url encode the ":" would be "&#58" so the strtok will not work. I also checked my fix already and it works.

Braintelligence:
My earlier PR should also be changed then. I'll create a new PR in a few mins.

mkapra:
@Braintelligence I also can fix this if you want?

mkapra:
@Braintelligence I see that I already fixed this in this pull request by demand of @mkuron. So no need to change anything.

Braintelligence:
Oh it was the same line as my earlier PR?

mkapra:
@Braintelligence Yes. The problem was that the port was also in the host option. I am thinking about if we need the strtok at HTTP_PORT? I think it's already right there.
I will test it without it.

Braintelligence:
Yup, you touched a new part and also the old one with the wrong strtok order. I approve :)

mkapra:
@Braintelligence @mkuron I was wrong by adding strtok() to the host option. @Braintelligence 's fix was nearly the right one. It only was the order of urlencode() and strtok() there.
I tried it already on my server.

andryyy:
Looks good to me now. Any complains?

mkuron (approved):
Haven't tested it, but it looks correct to me now.

Braintelligence (approved):

Issue #241, opened 2017-05-03
High Availablilty


SomeGeek:
Implement the functionality for setting up:

- A backup MX. In the installer, one can choose the behaviour, like the installation of a anti-spam stack (to prevent backscatter mail) _(No mails get lost, service will be interrupted, less hardware required)_
- A hot spare. Synchronize the configuration and contents. When one server goes down, the other immediately continues the service. _(No mails get lost, service will NOT be interrupted, at least the same hardware required)_

philip-ulrich:
You can already configure this as a backup mx. The hot spare would be cool though.

Braintelligence:
And to counter it being down or not functional you can incorporate healthchecks.

thomas126:
I like the hot spare request. Looking forward to it!

Braintelligence:
@andryyy Wouldn't traefik help with high-availability on a docker-scale? It could route everything as needed, but how would postfix/dovecot and so on replicate properly?

andryyy:
No, it wouldn't help or do anything better than Nginx.

And yes, the important replication is for Dovecot.

I don't plan this in the open source mailcow, same with ACL in the long run. But you can still go ahead and configure two mailcows to replicate, that's not a problem. I did this quite often over the past weeks. But that's something used and asked for primarily by commercial users. I offer it as commercial service by mail and soon via mailcow.email and servercow.de. As part of a service package or one-time setup.

Braintelligence:
Would it be alright to submit a PR for the mailcow docs to include how to achieve replication?

andryyy:
Sure, but there are several things to note... about the SQL cluster and its hurdles (versions, upgrading, witness etc., failover), replication of Dovecot itself, HAProxy (most will want this) and the proper service configuration + changes etc.

A simple Dovecot replication is not enough in most cases.

thomas126:
So you are not planning on including this in the open source version but offer this as a paid one time setup service? Or did I misunderstood?

andryyy:
As of today I offer to create replicated setups, but it is not part of mailcow dockerized.

thomas126:
Do you have an email address for further information?

Also would it be possible to manually restore one server out of the primary one and then use some kind of software to sync IMAP mailboxes? That would at least be some kind of workaround right?

andryyy:
Info@servercow.de :-)

Yes, you can do it like this. Or restore files from the maildir. Dovecot is very good in auto-fixing its index.

[...]

PR #2796, opened 2019-07-19
Multiply instances

gmarke:
As I already mentioned before in issue created some time ago => with this solution it's possible to run multiply instances of mailcow on one host.

All is about to consider COMPOSE_PROJECT_NAME when selecting data from docker api.

in this commit I made some review and change it in all files where CONTENER_D is assigned.

Please merge this in to project ==> this is not impacting any security or performance part it's just extended select.

### DIFF
--- a/data/Dockerfiles/acme/docker-entrypoint.sh
+++ b/data/Dockerfiles/acme/docker-entrypoint.sh
@@ -60,9 +60,9 @@ mkdir -p ${ACME_BASE}/acme
 reload_configurations(){
   # Reading container IDs
   # Wrapping as array to ensure trimmed content when calling $NGINX etc.
-  local NGINX=($(curl --silent --insecure https://dockerapi/containers/json | jq -r '.[] | {name: .Config.Labels["com.docker.compose.service"], id: .Id}' | jq -rc 'select( .name | tostring | contains("nginx-mailcow")) | .id' | tr "\n" " "))
-  local DOVECOT=($(curl --silent --insecure https://dockerapi/containers/json | jq -r '.[] | {name: .Config.Labels["com.docker.compose.service"], id: .Id}' | jq -rc 'select( .name | tostring | contains("dovecot-mailcow")) | .id' | tr "\n" " "))
-  local POSTFIX=($(curl --silent --insecure https://dockerapi/containers/json | jq -r '.[] | {name: .Config.Labels["com.docker.compose.service"], id: .Id}' | jq -rc 'select( .name | tostring | contains("postfix-mailcow")) | .id' | tr "\n" " "))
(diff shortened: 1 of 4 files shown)


mkuron:
The original pull request, #2755, only dealt with php-fpm and had to be reverted because it resulted in an invalid select statement. Clearly, this pull request needs to be properly tested before merging.

andryyy:
Yes, we can only merge if it is fully tested. The last PR did not address all problems, just like this does not cover all problems with multiple mailcows on a single host (netfilter for example).

Someone needs to pull and test, I will not try it again, as I don't see an advantage anyway. :(
Is this even safe to use with every compose version?

mkuron:
> Is this even safe to use with every compose version?

Good point. I think the first version that supports config format 2.1 was docker-compose 1.9 from three years ago. We need to test whether that already adds the necessary attributes to the containers.

> just like this does not cover all problems with multiple mailcows on a single host

I expect we‘ll get there eventually. If @gmarke really wants to do this, he‘ll need to find and fix the other problems too. But I think it‘s fine if he posts small pull requests like this one that fix issues as he finds them. As long as they don‘t change behavior on regular mailcows.

andryyy:
Yes, that's true.

When I find more time, I will try to contribute and fix netfilter-mailcow.

melihsd:
This is working perfectly fine for me;
I tested my already existing systems and did emailing, attachments etc. but not ipv6 yet;
But its working fairly well without any errors. 

grafik

Hope it will get merged soon and also the ipv6 problem will get solved. 

I may setup a test system with 10 installations to see if there will be any problems.

gmarke:
Cool ProJaCore ;) 
Finally someone interested about this solution..
If you describe what is about concerning ipv6 maybe I can do something..

djdomi:
i am interessted, but more in a solution to create an mailcow cluster on different hosts, because i want to have an active/backup situation without re-creating all accesses/domains on the backup-mx

mkuron:
@djdomi, high availability is a completely different thing. There is a separate issue for that.

djdomi:
yup i agree :) but still a feature i look for
Am Donnerstag, August 22, 2019 09:07 CEST, schrieb Michael Kuron <notifications@github.com>:
 
@djdomi, high availability is a completely different thing.

andryyy:
Not planned in the open source mailcow. If you want to provide a PR, go ahead.

djdomi:
> Not planned in the open source mailcow. If you want to provide a PR, go ahead.

Hm, maybe theres an idea: You can create a site for export all domains for an import of backup-mx i t hink that would be the easiest way at all?

patschi:
In general issue #241 might be more related for this kind of feature request.

> Hm, maybe theres an idea: You can create a site for export all domains for an import of backup-mx i t hink that would be the easiest way at all?

There's a way to retrieve all domains from a mailcow instance using the API: https://github.com/mailcow/mailcow-dockerized/blob/7483f2e04a6e10a6155102cf7f0d07c20992941b/data/web/json_api.php#L228-L242

So you might be able to script that part yourself.

PR #2804, opened 2019-07-23
Fix pointer events for nested button elements

### LINKED ISSUE #2802: Entries in quarantine table getting checked / unchecked on action button click
Every time you click on an icon of any button in the actions column of the quarantine table, the corresponding entry changes the checked / unchecked state.

friedPotat0:
Fixes #2802

### DIFF
--- a/data/web/css/build/008-mailcow.css
+++ b/data/web/css/build/008-mailcow.css
@@ -42,6 +42,9 @@
 .btn {
   text-transform: none;
 }
+.btn * {
+  pointer-events: none;
+}
 .textarea-code {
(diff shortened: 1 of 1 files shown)


andryyy:
Hi, sounds good, but does it break anything else in the UI?

friedPotat0:
As far as I know, all the buttons in the entire UI are constructed in the same way. They consist of a simple text and sometimes of a prefixed or appended icon span.

The pointer events for the buttons themselves remain unaffected by the change. I haven't found a button where a nested element should be clickable in addition to the actual button click.

Even the checkbox lists that are expanded when clicked on a button (e.g. the search filter in the quarantine page) are not implemented as a child element of the button itself, but as a seperate list next to the button.

andryyy:
Thanks. :)

PR #2839, opened 2019-08-06
[ClamAV] Update to 0.101.3

keigel2001:
0.101.3 partly fixed the issue with zipbombs (https://bugzilla.clamav.net/show_bug.cgi?id=12356)


Power2All:
Problem is still there:

``
Hanno Böck 2019-08-06 03:26:14 EDT
Can confirm that there are still issues. I've attached both examples created based on David's instructions.
``

And yes, I noticed the partly, but the problem isn't fixed at all as of yet, as the bug is still there.
We need to wait for them to fix the bug completely.

PR #2845, opened 2019-08-10
htmlspecialchars for value attributes in admin gui

heavygale:
Escaping htmlspecialchars in value-attribute-values. 

For example, that's required when specifying the notifications sender address using this format:
`"name to display" <sender@adress.tld>`

Currently the html is broken if one saves a address with that format:
grafik


andryyy:
Oh, thank you!

PR #2855, opened 2019-08-12
Adding subnets to API_ALLOW_FROM

localevil:
I needed to specify a subnet so that users could register themselves on my network. I thought maybe this logic will come in handy for you. If you see another solution to this problem, please share your opinion.


patschi:
I appreciate the work on this, however IMHO using the CIDR notation might be a better choice. So, for instance, using like `127.0.0.1/8`, `172.16.0.0/12` and so on. This way it's easier and more secure to whitelist own IP subnets, instead of using `*` wildcards at specific places.

For this I'd suggest moving the IP validation part more towards PHP. Either in the phpfpm containeron the start (using a PHP script), or on-demand on the API call. Your opinion on this, @andryyy?

andryyy:
Yes, cidr - functions to verify this already exist in mailcow. Something like filter_var_ip && valid_network. A bash function could be implemented.

PR #2865, opened 2019-08-16
Update main.cf

AndreKoepke:
Dont use old (and unsecure) versions of TLS.

### DIFF
--- a/data/conf/postfix/main.cf
+++ b/data/conf/postfix/main.cf
@@ -94,13 +94,13 @@ smtpd_tls_exclude_ciphers = ECDHE-RSA-RC4-SHA, RC4, aNULL, DES-CBC3-SHA, ECDHE-R
 smtpd_tls_loglevel = 1
 
 smtp_tls_mandatory_protocols = !SSLv2, !SSLv3, !TLSv1, !TLSv1.1
-smtp_tls_protocols = !SSLv2, !SSLv3
+smtp_tls_protocols = !SSLv2, !SSLv3, !TLSv1, !TLSv1.1
 
 lmtp_tls_mandatory_protocols = !SSLv2, !SSLv3, !TLSv1, !TLSv1.1
 lmtp_tls_protocols = !SSLv2, !SSLv3, !TLSv1, !TLSv1.1
 
 smtpd_tls_mandatory_protocols = !SSLv2, !SSLv3, !TLSv1, !TLSv1.1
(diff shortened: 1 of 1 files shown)


andryyy:
We don't want that for non-mandatory TLS connections.
Enforce TLS with a remote to use the mandatory ciphers and protocols. You can either enforce outgoing or incoming* TLS in mailcow UI per mailbox, create TLS policy maps or enforce TLS altogether in main.cf (not a good idea).

Using only super secure ciphers and protocols for non-mandatory connections will result in plain-text sessions whenever a TLS session fails. It is better to use TLS v1.0 with a **potential** dangerous cipher than _no TLS at all_.

Edit: *If you enforce TLS incoming per mailbox via mailcow UI (reject_plaintext_session), the mandatory ciphers are _not_ used. I was wrong. Everything else is correct, though.

AndreKoepke:
It is not even better, when old-TLS-protocols and uncrypted connections are blocked per default?
Many mail-servers are already block uncrypted connections.

Mailcow is a tool, that make mails simple and secure. I think, its not good, when every admin has to setup correct  TLS policy map.

andryyy:
You can set the smtpd policy from "may" to "encrypt" in main.cf or just enforce incoming TLS for your mailbox and increase the non-mandatory ciphers and protocols in your installation. 

Sadly you cannot/shouldn't create a TLS-only setup as of today, if you rely on mail communication. Too many server still send unencrypted. I hope Google starts to reject unencrypted sessions in the future. This will help to speed up the process.

Critical communication should be encrypted by TLS policy maps on both sides.

AndreKoepke:
Sure? My company (a software-house with over 700 employees) starts blocking non-TLS mails since some months and we have no problems.  
I dont know any mail-provider, which cant speek TLS. And you?

So, if I can't convince you, then keep this Pull-Request closed. But I keep my mail-server running TLS-only. :)

andryyy:
This PR does not implement TLS-only. It actually does the opposite. If incoming connections cannot be established with TLS, because the remote does not know how to handle TLSv1.2, it falls back to plain text. Instead of TLSv1 you now use no encryption at all. 

Yes, there are many servers still mailing without TLS. Just enforce TLS on a busy mail server and wait a day for complains. I can see many plain text sessions per day. Sucks, but if mail is business critical, you cannot drop them. 

There is no way a bigger company would not notice an impact when it enforces TLS (sadly).

PR #2880, opened 2019-08-23
rspamd - Switch form Ubuntu to Debian

christianbur:
I know there are no pull requests desired that contain an update of the base image, so see this as a suggestion. Rpamd is the last mailcow (own) image based on ubuntu, because [postfix has already been ported to debian](https://github.com/mailcow/mailcow-dockerized/commit/3136e020f6c0afed83feaddddbd05fff2029fc92#diff-ff63e3fee9825bf5e0806f90529d6eb7). 
Therefore i think it would be good if the mailcow own images were based only on Alpine and Debian. 

I used the following instructions https://rspamd.com/downloads.html.
I did a rudimental test with the new images, so you should try again yourself.

### DIFF
--- a/data/Dockerfiles/rspamd/Dockerfile
+++ b/data/Dockerfiles/rspamd/Dockerfile
@@ -8,7 +8,7 @@ ENV LC_ALL C
 RUN apt-get update && apt-get install -y \
   tzdata \
   ca-certificates \
-  gnupg \
+  gnupg2 \
   apt-transport-https \
   dnsutils \
   && apt-key adv --fetch-keys https://rspamd.com/apt-stable/gpg.key \
--- a/data/Dockerfiles/rspamd/Dockerfile
+++ b/data/Dockerfiles/rspamd/Dockerfile
@@ -1,22 +1,26 @@
-FROM ubuntu:bionic
+FROM debian:buster-slim
 LABEL maintainer "Andre Peters <andre.peters@servercow.de>"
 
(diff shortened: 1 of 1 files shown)


andryyy:
I had some really weird side effects with Debian and Rspamd in Docker a while ago, I tried to debug with Vsevo. One of these side effects were suddenly (yes, unpredictable time) broken regex maps.

I may give it another try anyway, I think. Besides that, there is no need to follow latest distro releases. We don't need to change everything all the time, it works just like it is and all used images are still in support. 

So... there are _always_ unpredictable behaviors I try to avoid as good as I can. You hopefully understand, that Rspamd and Debian is a mix (in Docker!), I want to test more thoroughly. :)

If anyone is willing to test, **please** go ahead and report back. Please test white- and blacklisting.

christianbur:
I agree with you there is no reason for an update at the moment (the dot updates should already be done :-) ), I can understand your position. 
It was just a thought, because in the future you only have to know the peculiarities of two base images (Alpine and Debian), currently there are three.

andryyy:
I agree Alpine and Debian sound solid. :) Hope this time Debian works flawless. I think it was due to some outdated libs. Or just some libs, that were missing in the Docker image of Debian but existed in Ubuntu.

andryyy:
Looks good!

christianbur:
i didn't expect it to be implemented that quickly, and there were even new mailcow images with the current dot releases :-)

PR #2888, opened 2019-08-28
environment variable TZ for olefy, memcached, ipv6nat

christianbur:
i just noticed that the olefy container has a wrong timezone, so i added the TZ environment variable, '[tzdata' is already included in the image](https://github.com/christianbur/mailcow-dockerized/blob/80747e99f365fc87df46ffd2cd2c2440d18ccdea/data/Dockerfiles/olefy/Dockerfile#L8). 

I also added the TZ environment variable to memcached and ipv6nat, these conatiners also have a wrong time zone. Unfortunately, the 'tzdata' is not included in both images (no mailcow own images), so the variable has no effect, but it probably doesn't do any harm.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -260,6 +260,8 @@ services:
     memcached-mailcow:
       image: memcached:alpine
       restart: always
+      environment:
+        - TZ=${TZ}  
       networks:
         mailcow-network:
           aliases:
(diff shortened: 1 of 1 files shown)


andryyy:
Can you resolve the conflicts? Will merge it then. :)

christianbur:
I can only fix it if I remove the TZ variable from the section "memcached-mailcow:". I don't get it
```
      environment:
        - TZ=${TZ}
```

andryyy:
You could try a new PR, the memcached part looks strange.

PR #2913, opened 2019-09-05
Update cs lang definition, fix translation

mamash:
A bunch of entries were missing and the translation was cumbersome in many ways. I went for a slightly simpler wording, cleared up some confused concepts and fixed many typos too. (I'm a part-time translator too.)


andryyy:
Thank you!!!

PR #2916, opened 2019-09-06
Disable SSL ticket support in dovecot

Thomas2500:
Because tickets are normally only generated on service start, we should disable it to provide better PFS.

### DIFF
--- a/data/conf/dovecot/dovecot.conf
+++ b/data/conf/dovecot/dovecot.conf
@@ -34,7 +34,7 @@ ssl_prefer_server_ciphers = yes
 ssl_cipher_list = ALL:!ADH:!LOW:!SSLv2:!SSLv3:!EXP:!aNULL:!eNULL:!3DES:!MD5:!PSK:!DSS:!RC4:!SEED:!IDEA:+HIGH:+MEDIUM
 
 # Default in Dovecot 2.3
-ssl_options = no_compression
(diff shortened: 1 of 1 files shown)


mkuron:
Could you please cite a source that explains why session tickets are supposedly insecure?

Also, are Postfix or Nginx affected too?

Thomas2500:
For example https://blog.compass-security.com/2017/06/about-tls-perfect-forward-secrecy-and-session-resumption/ describes the problem with session tickets which are not rotatet regurarely.
Also cloudflare describes their procedure with using session tokens at https://blog.cloudflare.com/tls-session-resumption-full-speed-and-secure/ which will be rotated hourly.

In nginx session tickets are already disabled
https://github.com/mailcow/mailcow-dockerized/blob/8608ded0ed4c87f40960751c9ad7a9b9eca029e6/data/conf/nginx/site.conf#L22

Postfix does use it's own session database on a per host basis.
https://github.com/mailcow/mailcow-dockerized/blob/8608ded0ed4c87f40960751c9ad7a9b9eca029e6/data/conf/postfix/main.cf#L10

In addition, as of http://www.postfix.org/postconf.5.html#smtp_tls_session_cache_timeout the key get's rotated by default every hour.

andryyy:
Thanks. :)

PR #2924, opened 2019-09-08
Respect COMPOSE_PROJECT_NAME when querying for container name

max-baz:
I noticed this when experimenting with mailcow by running a second instance of it on a single server, these scripts do not respect COMPOSE_PROJECT_NAME and therefore collide with already existing containers.

### DIFF
--- a/data/Dockerfiles/acme/docker-entrypoint.sh
+++ b/data/Dockerfiles/acme/docker-entrypoint.sh
@@ -60,9 +60,9 @@ mkdir -p ${ACME_BASE}/acme
 reload_configurations(){
   # Reading container IDs
   # Wrapping as array to ensure trimmed content when calling $NGINX etc.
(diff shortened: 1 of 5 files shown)


mkuron:
Partial duplicate of #2796.

max-baz:
Oh I didn't see that one! It seems my PR has a few extra bits that are missing in #2796, but I'm totally okay for this PR to be closed if the work is continued there.

andryyy:
This didn't work the last time we merged something similar.

Also not sure if all docker-compose versions use the same json key. Do they?

I'd be fine merging, if we are sure this is tested. :)

max-baz:
I don't know for sure, but some of the code in `master` already uses this exact logic 😉 

https://github.com/mailcow/mailcow-dockerized/blob/a908b2da6e9a6da98b1d0271dca725e8012b62fd/data/web/inc/functions.docker.inc.php#L28

andryyy:
:) Should be fine then or at least break everywhere evenly. 😄 

I will build it and test.

andryyy:
jq -rc 'select(.project == "${COMPOSE_PROJECT_NAME}")'

This would not work for example. 

I will consider a rework.

PR #2936, opened 2019-09-14
Add choice to restore all

LorenzYouKnow:
Just for reasons


andryyy:
Duplicate and also: what if not everything was backed up?

PR #2940, opened 2019-09-15
[RSPAMD] Split bad words into multiple files per language

ntimo:


patschi:
Have you tested this PR? I don't see `BAD_WORDS_DE` defined in your PR, so it's going to be ignored (or rspamd might throw errors due to missing map definition)

Only following `BAD_WORDS` map definition exists:
https://github.com/mailcow/mailcow-dockerized/blob/6e82a3592957e9a2145a875d145423773acd2d0b/data/conf/rspamd/local.d/multimap.conf#L115-L121

ntimo:
@patschi I totally missed that. Thanks, going to add it.

andryyy:
Mörciii

PR #2941, opened 2019-09-15
Make is_valid_domain_name more strict

### LINKED ISSUE #1118: Admin-Interface allows creation of "invalid" domains
Currently, you can create a domain like "mynamespace" without a TLD like .com, .de etc. This is fine, but then you cannot create any mailboxes for those domains (in the creation dialog, you will receive an error: "Mailbox name is invalid").

I believe you should either:

- not be able to create these domains in the interface
- or be able to add a mailbox to them

patschi:
Fixes #1118

### DIFF
--- a/data/web/inc/functions.inc.php
+++ b/data/web/inc/functions.inc.php
@@ -709,14 +709,14 @@ function user_get_alias_details($username) {
   }
   return $data;
 }
-function is_valid_domain_name($domain_name) { 
+function is_valid_domain_name($domain_name) {
 	if (empty($domain_name)) {
 		return false;
 	}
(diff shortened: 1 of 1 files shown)


andryyy:
What was not covered by the previous check? Do you have examples?

patschi:
See issue #1118. Using `asd` was a valid domain, but what in reality isn't. This slight modification forces the existence of a dot in the domain, so only `asd.com` is valid.

andryyy:
I always thought that was fine, if you want to route some crazy internal stuff with your mailcow. Not sure. For me personally it is fine to change it. @mkuron ?

mkuron:
If you want to do crazy stuff, you can just comment out the validation. For most users, more strict validation makes things easier.

andryyy:
Merciiiii

Issue #2948, opened 2019-09-17
Mails from blacklisted mailserves not visible in quarantine and rspamd history


friedPotat0:
**Prior to placing the issue, please check following:** *(fill out each checkbox with a `X` once done)*
- [X] I understand that not following below instructions might result in immediate closing and deletion of my issue.
- [X] I have understood that answers are voluntary and community-driven, and not commercial support.
- [X] I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

---

**Description of the bug**: What kind of issue have you *exactly* come across?

Mails that are blocked using a public blacklist for mailservers (e.g. hostkarma.junkemailfilter.com) are not added to the quarantine. They are not even visible in the rspamd history.

At the moment a t-online mailserver (mailout08.t-online.de) is listed on multiple public blacklists (see https://mxtoolbox.com/SuperTool.aspx?action=blacklist%3amailout08.t-online.de&run=toolpage). Any mail that is sent from this mailserver is directly blocked by the mailcow mail system and the following response is transmitted from the t-online mailserver back to the sender:

> This is the mail system at host mailout08.t-online.de.
> 
> I'm sorry to have to inform you that your message could not
> be delivered to one or more recipients. It's attached below.
> 
> For further assistance, please send mail to postmaster.
> 
> If you do so, please include this problem report. You can
> delete your own text from the attached returned message.
> 
> The mail system
> 
> <xyz@example.com>: host
> mail.example.com[xxx.xxx.xxx.xxx] said: 550 5.7.1 Service
> unavailable; client [194.25.134.20] blocked using
> hostkarma.junkemailfilter.com (in reply to RCPT TO command)
> 
> [...]

**Reproduction of said bug**: How *exactly* do you reproduce the bug?

1. Remote mailserver is listed on a public blacklist
2. A sender sends a mail through the blacklisted mailserver to the mailcow mail system

__I have tried or I do...__ *(fill out each checkbox with a `X` if applicable)*
- [X] In case of WebUI issue, I have tried clearing the browser cache and the issue persists.
- [ ] I do run mailcow on a Synology, QNAP or any other sort of NAS.

**System information**

Further information (where applicable):

I will provide more information if neccessary to solve the problem.

| Question | Answer |
| --- | --- |
| My operating system | I_DO_REPLY_HERE |
| Is Apparmor, SELinux or similar active? | I_DO_REPLY_HERE |
[...]

PR #2950, opened 2019-09-17
update postscreen whitelist by using postwhite

friedPotat0:
Updated postscreen whitelist by using the latest postwhite version (v3.4 from master@https://github.com/stevejenkins/postwhite) and latest SPF-Tools (from master@https://github.com/spf-tools/spf-tools). The default postwhite.conf with no additional custom hosts was used to generate the list.

#2948


mkuron:
Could you please check whether there are any other big email providers that should be added to this list? T-Online from #2948 is certainly a candidate, but there certainly are more. Also, can you please add a script to Mailcow that can be run manually to update the list?

friedPotat0:
I wrote a short shell script to update the postscreen whitelist manually. More custom hosts that need to be whitelisted can be added to the `COSTOM_HOSTS` variable (I've already added the big providers). I've also included the t-online subnet as a static host that is additionally appended to the whitelist after running postwhite.

Should I also add the command for reloading postfix to the end of the update script?

mkuron:
No need to reload Postfix. I didn't intend this script to be used by the average user, just to make it easier for us to update the postwhite list.

andryyy:
Yes, should not be run by the user but us to update the list. :)

andryyy:
Thank you!

PR #2965, opened 2019-09-22
[Postfix] Add NO_RENEGOTIATION to tls_ssl_options

phenomax:
Hi!
Running the [internet.nl email test](https://en.internet.nl/mail/) for the latest mailcow, I nocited a warning for `Client initiated renegotiation`

image

[`postconf(5)`](http://www.postfix.org/postconf.5.html) states that adding the `NO_RENEGOTIATION` flag 

> can reduce opportunities for a potential CPU exhaustion attack.

Which is why I propose to include it by default.


andryyy:
Hi, thank you, NO_RENEGOTIATION seems to be a Postfix 3.4 thing. Nice. :)

PR #2993, opened 2019-10-02
[Web] Return 401 status code when API authorization fails

ntimo:
This PR will return a 401 HTTP status code when the API authorization fails due to a invalid API key or a not whitelisted IP address.


irgendwr (approved):

andryyy:
Thanks!

PR #2999, opened 2019-10-03
[Nginx] Fix nginx config for API docs

ntimo:
This PR fixes the nginx config and also updates the docs to the latest version.


andryyy:
Merci

PR #3002, opened 2019-10-03
[API] Status Codes and only allow valid http method

ntimo:
This PR adds:
- Correct http status codes
- a new response when no API route is matched
- Only allow POST (for add / delete routes) and GET (only for GET routes)

How to test this PR:
- Use the WebUI normally and check if anything is broken or not working anymore.
- Use curl or another tool to test the API

~~Questions that still need to be answered:~~
~~- which status code should be returned when a $generic_failure occurs. Does the UI have to updated when sending a status code other then 200 OK ?~~

Example HTTP POST on GET only Route:
```
curl --request POST \
  --url https://mailcow.host/api/v1/get/logs/acme/1 \
  --header 'content-type: application/json' \
  --header 'x-api-key: apikey' 
```
Output (http status 405):
```
{
  "type": "error",
  "msg": "only GET method is allowed"
}
```

Example GET on POST only route:
```
curl --request GET \
  --url https://mailcow.host/api/v1/add/tls-policy-map \
  --header 'content-type: application/json' \
  --header 'x-api-key: apikey' 
```
Output (http status 405):
```
{
  "type": "error",
  "msg": "only POST method is allowed"
}
```

### DIFF
--- a/data/web/json_api.php
+++ b/data/web/json_api.php
@@ -206,6 +206,14 @@ if (isset($_SESSION['mailcow_cc_role']) || isset($_SESSION['pending_mailcow_cc_u
           case "tls-policy-map":
             process_add_return(tls_policy_maps('add', $attr));
           break;
+          default:
+            http_response_code(404);
+            echo json_encode(array(
+              'type' => 'error',
+              'msg' => 'route not found'
+            ));
+            unset($_POST);
+            die();
         }
       break;
       case "get":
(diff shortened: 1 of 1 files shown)


MAGICCC:
>     * which status code should be returned when a $generic_failure occurs. Does the UI have to updated when sending a status code other then 200 OK ?

What about 500? As long as it's an error on server side obvious

andryyy:
Thanks! :)

PR #3003, opened 2019-10-04
[Web] Replaced die(); with exit(); due to code conventions

ntimo:
Most of the time exit(); is used in the mailcow project. But sometimes die(); is also used. We should only use one of them since both do the same. 

https://www.php.net/manual/en/function.die.php
https://www.php.net/manual/en/function.exit.php


andryyy:
Merci

PR #3007, opened 2019-10-05
quarantine quick_release & quick_delete without buttonclick

tinect:
This will add possibility to quick_delete or quick_release in quarantine email-template by prepending "quick_". So there is no need to press button.

### DIFF
--- a/data/web/inc/triggers.inc.php
+++ b/data/web/inc/triggers.inc.php
@@ -12,10 +12,14 @@ if (isset($_POST["verify_tfa_login"])) {
 
 if (isset($_POST["quick_release"])) {
 	quarantine('quick_release', $_POST["quick_release"]);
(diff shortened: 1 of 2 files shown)


ntimo:
Could you please elaborate more on this quick release feature? I would guess it just changes the way the release is handled and removes the need to click the release button.

tinect:
will inspect it more

tinect:
I don't get any possibility to close the window automatically or something else.

In addition to this PR I'll work on more improvements to the quarantine-management :-)

Adorfer:
"Oneclick Release/Delete&Autolearn" would be very helpful. Because it would avoid enless "back and forth" in the browser with at least 2 clicks and one keypress per individually treated mail in the quarantine.

PR #3009, opened 2019-10-05
sogo_access acl for domain admin

heavygale:
New sogo_access acl for domain admins. If not granted the domain admin can't change sogo_access attribute on mailboxes.

New mailboxes are created with SOGo access if created by admin or domain admin with sogo_access acl, when $MAILBOX_DEFAULT_ATTRIBUTES['sogo_access'] = true;". If acls is mission or default = false, the mailbox is created without sogo_access.

-> Now it's possible to create a domain admin for users, who should not be allowed to grant themselves access to SOGo.


heavygale:
I've undone the automatic cleanup of trailing whitespaces, so my actual changes are better visible.

andryyy:
Can you change the lang files to something like "Can toggle SOGo access"? :)

andryyy:
Thanks. :)

PR #3010, opened 2019-10-05
Remove all trailing whitespaces

heavygale:
Removing trailing whitespaces from all files (except in data⁩/⁨web⁩/inc⁩/lib⁩/vendor⁩).

PR #3011, opened 2019-10-06
refactor code-/-style in autodiscover

tinect:
- codestyle
- using $login_user instead of $_SERVER['PHP_AUTH_USER'] for all queries, cause it has been cleaned up
- pushed $login_role to upper position, so there is an early break, while 'user' is important
- log requested user in `'Error: must be authenticated'`, if there was any sent, instead of just logging `none`

### DIFF
--- a/data/web/autodiscover.php
+++ b/data/web/autodiscover.php
@@ -5,8 +5,10 @@ $default_autodiscover_config = $autodiscover_config;
 if (file_exists('inc/vars.local.inc.php')) {
     include_once 'inc/vars.local.inc.php';
 }
-$autodiscover_config = array_merge($default_autodiscover_config,
-    $autodiscover_config);
+$autodiscover_config = array_merge(
+    $default_autodiscover_config,
+    $autodiscover_config
+);
 
 // Redis
 $redis = new Redis();
@@ -31,11 +33,11 @@ if (strpos($data, 'autodiscover/outlook/responseschema') !== false) {
 //$dsn = $database_type . ":host=" . $database_host . ";dbname=" . $database_name;
 $dsn = $database_type . ':unix_socket=' . $database_sock . ';dbname='
     . $database_name;
-$opt = [
+$opt = array(
     PDO::ATTR_ERRMODE => PDO::ERRMODE_EXCEPTION,
     PDO::ATTR_DEFAULT_FETCH_MODE => PDO::FETCH_ASSOC,
     PDO::ATTR_EMULATE_PREPARES => false,
-];
+);
 $pdo = new PDO($dsn, $database_user, $database_pass, $opt);
 
 $login_user = strtolower(trim($_SERVER['PHP_AUTH_USER']));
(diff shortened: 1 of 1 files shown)


andryyy:
I really appreciate this. But please don't introduce a new code style. Some fixes are fine, but please use 

}
catch {

for example. Writing from mobile, formatting is hard. :smile: 

Also stick to array( instead of [. 

I don't care whatever standard we follow or not. But we should at least try to stick to it. :)

If there is a reason like deprecation, I agree to changing it. 

Using ' instead of " is a good idea. 

Can you implement the changes or enlighten me? :)

André

tinect:
Hey @andryyy 
Changed code as requested.

Long syntax for array sounds legit :-) I personally prefer `write less, do more without destroying the readability of code`. And the readability is a point-of-view-situation ;-)

But `catch` in a new line hurts clarity, so it should stick onto try. At least that's how I feel.

While trying to get into code, I couldn't detect any code style. So I'd suggest introduce any code-style, at least PSR-1/PSR-2. How do think about it?

Best Regards
Sebastian

andryyy:
That's really something I want to think about in the team.

I don't care that much about the syntax, as long as it - at least - somehow looks like what we do now. I don't care about chaos, as long as we don't lose track. That's just how I work. :D Shame on me. 

We can think about that in the future. We would need to adjust the whole code. I'm not speaking out against it.

I'd like to know what @mkuron thinks about it.

tinect:
That's how I worked some years ago, too.

There are many possibilities to apply code-style automatically, without changing it's function. All IDEs should do it. I'm working with PHP-Storm, f.e. So that wouldn't be a con xD

mkuron:
For me this is similar to #3010. We don‘t gain much by reformatting the code, except for merge conflicts. If we do decide to reformat the code, a one-off is pointless as the code would diverge from the prescribed format over time. We would need a pre-commit hook that enforces format, and I don‘t think GitHub can easily do that. Not everyone uses the same IDE, so it‘s a hassle for developers, especially for people who just want to submit a quick bugfix. Also, why stop with PHP? There is lots of Python and Bash and some other stuff too...

tinect:
There is no need everyone uses the same IDE.
Editorconfig f.e. is supported in several IDEs. And there possibilites to let the requested code check automatically in github.

You'll gain much. Like readable code to better check how the code is grouped f.e.

And for sure, don't stop with PHP, but starting with could be a step. :-)

tinect:
@mkuron could you please tell me, why this PR has been closed?

> * codestyle
> * using $login_user instead of $_SERVER['PHP_AUTH_USER'] for all queries, cause it has been cleaned up
> * pushed $login_role to upper position, so there is an early break, while 'user' is important
> * log requested user in `'Error: must be authenticated'`, if there was any sent, instead of just logging `none`

mkuron:
See #3010, we're not going to do style/whitespace-only changes. They are too much of a hassle with customizations. Please feel free though to put any functional changes into a new pull request. Regarding logging, please check whether the fail2ban service is affected by any changes to log messages.

PR #3012, opened 2019-10-06
add .editorconfig

tinect:
Suggested code-style as commented in #3011
This file is a complete .editorconfig-file exported from IDE PHPStorm.

Can be used in several IDEs
see: https://editorconfig.org/#download

especially rules:
- PSR-1/PSR-2
- long array
- catch and finally in new line


andryyy:
That's something we will decide internally, sorry.

PR #3023, opened 2019-10-09
Add option to download quarantine item as eml

friedPotat0:
Adds an action to the detail view of quarantine items to download the raw mail as an eml file for direct inspection in any mail client.


andryyy:
Thank you.

PR #3026, opened 2019-10-10
Fix alignment of senders in quarantine detail view

friedPotat0:
Fixes wrong alignment of senders in the quarantine detail modal if items spread over multiple lines.

### DIFF
--- a/data/web/css/site/quarantine.css
+++ b/data/web/css/site/quarantine.css
@@ -104,6 +104,8 @@ span.mail-address-item {
   border: 1px solid #ccc;
   padding: 2px 7px;
   margin-right: 7px;
+  display: inline-block;
+  margin: 2px 6px 2px 0;
 }
 
 table tbody tr {


andryyy:
If you feel you have time on your hands, you can adjust the Rspamd history in /debug to match your view.

Looks like this atm:

image

friedPotat0:
Actually I like the way all symbols are visible at once with all their corresponding options in the debug log. This gives the user a quick overview when inspecting the entries in the log. I don't think it is necessary to make it look like the quarantine detail view.

An improvement would be to sort the symbols in the same way as in the quarantine detail view. Starting with the negative values, then the positive ones and the ones that don't do anything (zero values) at the end. Additionally the text color of the zero value symbols should not be green, because they are not doing anything positive nor negative.

If you want, I can create a pull request with the proposed changes the next days.

andryyy:
That's a good point. 

Yes, feel free. :)

PR #3036, opened 2019-10-12
Change rspamd debug log and quarantine symbols

friedPotat0:
Changes the appearance of the symbols in the rspamd log according to #3026.

Additionaly improves the sorting of the symbols in the debug log as well as in the quarantine in a way that the smbols with scores that contribute more to the final score are shown first in every group of symbols. E.g. [10, 9, -20, -10, 0, 0]. So far the negative scores (green) were only sorted in a mathematical correct way (-10, -20).

Quarantine:
image

Rspamd debug log:
image

### DIFF
--- a/data/web/js/site/debug.js
+++ b/data/web/js/site/debug.js
@@ -468,7 +468,17 @@ jQuery(function($){
       else {
         item.rcpt = item.rcpt_smtp.join(", ");
       }
-      Object.keys(item.symbols).map(function(key) {
+      Object.keys(item.symbols).sort(function (a, b) {
+        if (item.symbols[a].score === 0) return 1
+        if (item.symbols[b].score === 0) return -1
+        if (item.symbols[b].score < 0 && item.symbols[a].score < 0) {
+          return item.symbols[a].score - item.symbols[b].score
+        }
+        if (item.symbols[b].score > 0 && item.symbols[a].score > 0) {
+          return item.symbols[b].score - item.symbols[a].score
+        }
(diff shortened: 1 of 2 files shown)


friedPotat0:
Please wait for the merge until I changed the name of the commit author to match the GitHub username. I'm working on a different machine so I haven't noticed it in the first place ;).

Edit: Looks good now :D. Feel free to merge...

andryyy:
Mörciii

andryyy:
Can you recheck this, please? :)

image

friedPotat0:
I am able to reproduce the issue by setting item.symbols to undefined. I could fix it by printing an empty string if a mail has no symbols at all.

I'm confused by the fact that a mail can have no symbols, but a final rspamd score.

andryyy:
Can you check with the new Rspamd image? I think symbols might be a value of the metric name now. It was "{symbols}" before and is "{default: symbols}" now. 

We'd need to create a check wether a history or quarantine item uses the old format or the new format.

I could be wrong, but it looks like it changed.

friedPotat0:
I've updated rspamd to 1.47, but couldn't reproduce the issue either.

andryyy:
Have you sent and received multiple (new) mails since then?

Mit besten Grüßen
André Peters

friedPotat0:
Thanks for the hint. I could successfully reproduce the issue now. The issue was caused by a missing `time_virtual` value in some mails that was not set by rspamd 1.47.

Fixed in #3038.

andryyy:
Thanks!!

PR #3038, opened 2019-10-12
Fix rspamd v1.47 debug log

friedPotat0:
Fixes the debug log of rspamd for v1.47. The `time_virtual` value was missing for some mails and caused the log to stop rendering the receiver address and the symbols (exception was thrown inside the process_table_data loop).

Now showing only the `time_real` value for "Scan time" when no virtual time set.


andryyy:
Thank you!

PR #3041, opened 2019-10-14
Sync cs against en language file, improve translation.

mamash:
After a month of intensive usage, time to fix up some bits in the Czech translation.


andryyy:
Nice! Thank you very much!

PR #3045, opened 2019-10-15
Support Filename Blacklist

AndyX90:
Support Filename Blacklist.
Taken from: https://rspamd.com/doc/modules/multimap.html#examples

### DIFF
--- a/data/conf/rspamd/custom/bad_filename.map
+++ b/data/conf/rspamd/custom/bad_filename.map
@@ -1,0 +1,1 @@
+#Insert Filenames to Blacklist them.
--- a/data/conf/rspamd/local.d/multimap.conf
+++ b/data/conf/rspamd/local.d/multimap.conf
@@ -127,3 +127,11 @@ BAD_WORDS_DE {
   regexp = true;
   score = 0.1;
 }
+
+FILENAME_BLACKLISTED {
+  type = "filename";
+  filter = "extension"
+  map = "${LOCAL_CONFDIR}/custom/bad_filename.map";
+  action = "reject";
+  message = "A restricted file type was found";
+}


andryyy:
Thanks, but I think this may collide with `https://github.com/mailcow/mailcow-dockerized/blob/master/data/conf/rspamd/local.d/mime_types.conf`

You can increase the score here. :)

AndyX90:
Thanks, i know about the mime_types.conf. But in my opinion it would be better to combine both.
You can increase the score of some special mime types via mime_types.conf to get a clearer result in some circumstances **and** you can specify fileextensions via map and immediately match and reject(quarantine) if a mail comes in with one of the specified fileextensions.

andryyy:
You can set exe = 2000, it will be rejected. That's pretty much the same. :) Not sure if this confuses more people.

AndyX90:
@andryyy Do you know a solution to only affect inbound mailtraffic with mime_types-module?

andryyy:
Try a new settings map in mailcow UI with this content:

```
priority = 8;
authenticated = yes;
apply "default" {
  symbols_disabled = ["MIME_BAD_ATTACHMENT"];
}
```

Not sure if that's the symbol with the high score. You should recheck this. :)

andryyy:
image

AndyX90:
@andryyy I've tried your approach, but sadly it won't work.
I tried to disable the symbols "MIME_BAD_EXTENSION" which maches for me with 4000 pts.

> priority = 8;
> authenticated = yes;
> apply "default" {
>   symbols_disabled = ["MIME_BAD_EXTENSION"];
> }

I have also tried to change to priority to 'high' and to remove the "default" but both without luck.
The Symbol "MIME_BAD_EXTENSION" matches for my authenticated user.

andryyy:
Try with groups_disabled instead and type "mime_types" as group name. 

groups_disabled = ["mime_types"]

Restart Rspamd, check the settings map in mailcow UI (you can see its content there) and set log level of Rspamd to info, if it still fails.

AndyX90:
> Try with groups_disabled instead and type "mime_types" as group name. groups_disabled = ["mime_types"] Restart Rspamd, check the settings map in mailcow UI (you can see its content there) and set log level of Rspamd to info, if it still fails.
> […](#)
> Am 05.11.2019 um 14:13 schrieb Andy Binder ***@***.***>: ﻿ @andryyy I've tried your approach, but sadly it won't work. I tried to disable the symbols "MIME_BAD_EXTENSION" which maches for me with 4000 pts. priority = 8; authenticated = yes; apply "default" { symbols_disabled = ["MIME_BAD_EXTENSION"]; } I have also tried to change to priority to 'high' and to remove the "default" but both without luck. The Symbol "MIME_BAD_EXTENSION" matches for my authenticated user. — You are receiving this because you were mentioned. Reply to this email directly, view it on GitHub, or unsubscribe.

I have inserted the following:

> priority = 8;
> authenticated = yes;
> apply "default" {
>   groups_disabled = ["mime_types"];
> }

but this is also matching as "MIME_BAD_EXTENSION (4000) [xls]".
I will try to change rspamds log level and will report back.

AndyX90:
@andryyy Sorry for asking again, but i have not found anything relevant in my logs.

Following in Rspamd-Log (Web-UI):
```
18.11.2019, 13:19:38	ixxx@xxxxxx	axxxxx@xxxx.xx	reject
	[ 3980.40 / 10 ]	 ixxx@xxxxxx
IP address	172.22.x.x
Subject	test
Symbols	MIME_BAD_EXTENSION (4000) [xls]
MID_RHS_NOT_FQDN (0.5)
MAILCOW_AUTH (-20)
MIME_GOOD (-0.1) [multipart/mixed, multipart/alternative, text/plain]
FROM_EQ_ENVFROM (0)
MAILCOW_DOMAIN_HEADER_FROM (0) [xxxx.xx]
TO_MATCH_ENVRCPT_ALL (0)
FROM_HAS_DN (0)
RCPT_COUNT_ONE (0) [1]
TO_DN_NONE (0)
RCVD_COUNT_ZERO (0) [0]
FREEMAIL_ENVRCPT (0) [xxx.xx]
HAS_ATTACHMENT (0)
ARC_NA (0)
FREEMAIL_TO (0) [xxx.xx]
MIME_TRACE (0) [0:+, 1:+, 2:+, 3:~, 4:-, 4:~]
DYN_RL_CHECK (0)
Msg size	8.7 KiB
Scan time	0.177
ID	40-5dd28c80-f-5cafd000@134340696
```
Applied User-Settings:
```
priority = 8;
authenticated = yes;
apply "default" {
groups_disabled = ["mime_types"];
}
```
I also tried the following:
```
priority = high;
authenticated = yes;
apply {
groups_disabled = ["mime_types"];
}
```
But it doesn't help.
It seems that the user-settings, set in the UI simply get ignored.

andryyy:
Did you remove your local change in multimap.conf?
Can you post your settings map? Or at least check if it containers the correct rule? Do you have other settings maps active?

andryyy:
I think I know what's going on here. Can you edit a domain of yours and add a useless blacklist or whitelist record? Restart rspamd-mailcow after that, please, and try again.

AndyX90:
@andryyy Thanks a lot!
I did the following: 
1. i entered *@example.org in my domainwide blacklist
2. restarted rspamd
--> same result
3. re-applied the settings map
4. restarted rspamd
--> it works!
Can i now delete the blacklist entry?

andryyy:
Yes. The settings map do not update the last modified header. Rspamd never sees the map as updated. :( Sorry. I will fix it later.

andryyy:
Fixed, thank you!! :)

AndyX90:
@andryyy Thanks for your great work and sorry for posting here again, but I have another mailcow instance with the same settings (on the newest patch-level) and there the increasing of the scores for specific mime-types has absolutely no effects.
For example I increase the score for pptx files by 2000 and the sent Mail with pptx attachement does not get the symbol (MIME_BAD_EXTENSION). There are no user settings maps. Could it be somehow related to this patch?

Thank You!

andryyy:
You need to join the Telegram channel for support. :) It is also community-driven.

andryyy:
You can copy and send me your config changes and settings map. I also need the full, original mail, that is not detected. 

Please post the full log of Rspamd, when this mail arrived. Including all symbols. 

I cannot guarantee I find time for it. :(

info@servercow.de

PR #3052, opened 2019-10-17
Prefetch images before asking for stopping containers

SamuelNitsche:
Since there is now the option to prefetch images, I think it would make more sense to actually prefetch all images before asking to reboot mailcow.

This way you can prefetch the images and stop the update process right after.

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -328,12 +328,6 @@ else
   exit 1
 fi
 
-read -r -p "Are you sure you want to update mailcow: dockerized? All containers will be stopped. [y/N] " response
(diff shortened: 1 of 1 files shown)


andryyy:
I don't think it makes sense to prefetch images before asking if you want to update mailcow.

You can use "--prefetch" to fetch images and quit.

SamuelNitsche:
Hm, I think I missed the "--prefetch" option. Looks good to me. Thank you!

andryyy:
No problem at all. :)

PR #3062, opened 2019-10-19
cache JS- and CSS-File to reduce response-time

tinect:
Just noticed the minifier of JS and CSS results in >300ms response-time on admin-page f.e.
this PR is halving the response-time by hashing content and caching the files in tmp-folder.

Is the tmp-folder cleared automatically by system or will I have to think of a method to clean up?


andryyy:
Thanks. :) Would indeed be nice to search for older files and delete them, when we create the new hash file. Can you implement this?

andryyy:
Nice, thanks!

PR #3065, opened 2019-10-20
ansible installation

ttsak:
Added ansible role and playbook for automated installation of mailcow-dockerized

### DIFF
--- a/helper-scripts/ansible/roles/mailcow-dockerized/tasks/main.yml
+++ b/helper-scripts/ansible/roles/mailcow-dockerized/tasks/main.yml
@@ -1,0 +1,38 @@
+---
+
+- name: Make sure git is installed
+  package:
+    name: git
+    state: present
+
+- name: Clone the master branch of mailcow-dockerized repo
+  git:
+    repo: https://github.com/mailcow/mailcow-dockerized
+    dest: /opt/mailcow-dockerized
+    umask: "0022"
+    #force: yes # overwrite modified files
+  ignore_errors: yes
+
+- name: Add mail.* on ADDITIONAL_SAN
+  replace:
+    path: /opt/mailcow-dockerized/generate_config.sh
+    regexp: "^ADDITIONAL_SAN.*"
+    replace: "ADDITIONAL_SAN=mail.*"
+
+- name: Generate a configuration file
+  shell: ./generate_config.sh
+  environment:
+    MAILCOW_HOSTNAME: "{{ inventory_hostname }}"
+    MAILCOW_TZ: "{{ timezone_selection }}"
+  args:
+    chdir: /opt/mailcow-dockerized
+    creates: mailcow.conf # is linked from .env - https://docs.docker.com/compose/env-file/
+  register: mailcow_generate_config
+
+- name: Pull the images and run the compose file
+  docker_service:
+    project_name: mailcowdockerized # name is fixed in mailcow.conf
+    project_src: /opt/mailcow-dockerized
+    state: present
+    pull: yes
+  when: mailcow_generate_config is changed
--- a/helper-scripts/ansible/roles/mailcow-dockerized/tasks/main.yml
+++ b/helper-scripts/ansible/roles/mailcow-dockerized/tasks/main.yml
@@ -7,7 +7,7 @@
 
 - name: Clone the master branch of mailcow-dockerized repo
   git:
-    repo: https://github.com/mailcow/mailcow-dockerized
+    repo: https://github.com/mailcow/mailcow-dockerized.git
     dest: /opt/mailcow-dockerized
     umask: "0022"
     #force: yes # overwrite modified files
--- a/helper-scripts/ansible/README.md
+++ b/helper-scripts/ansible/README.md
@@ -1,0 +1,24 @@
+## Installing using Ansible
+
+The provided playbook and role will install "mailcow-dockerized" using ansible.
+This is useful in case you are using ansible to setup all other aspects of your environment and do not want to ssh manually into a server for installing mailcow.
+
+## Requirements
+
+The host should already have installed:
+
+- docker (obviously!)
+- pip (to install docker-compose with required module for ansible)
+
+and from pip you should install the following packages:
+
+- docker-compose (for mailcow installation)
+- pyyaml (required for docker_service ansible module)
+
+These requirements are out of scope for this playbook but they are easy to implement in ansible as previous steps before using the role provided here.
+
+# Configuration
+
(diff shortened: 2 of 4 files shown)


andryyy:
I think @ntimo can evaluate that. :)

Thanks!

helper-scripts/ansible/roles/mailcow-dockerized/tasks/main.yml
```diff
@@ -0,0 +1,38 @@
+---
+
+- name: Make sure git is installed
+  package:
+    name: git
+    state: present
+
+- name: Clone the master branch of mailcow-dockerized repo
+  git:
+    repo: https://github.com/mailcow/mailcow-dockerized
```
ntimo: This is not going to work, since this is not the git url. It should be: `https://github.com/mailcow/mailcow-dockerized.git`

ntimo:
Thank you for your contribution. I truly appreciate your work here. I have checked your ansible playbook and the mailcow role and have some questions / comments.

First off I think that the mailcow-dockerized repo is the wrong place for a ansible role I think it would be better to have a dedicated git repo for this role / playbook (@andryyy what do you think?)

The mailcow role seems to clone the mailcow-dockerized repo this is fine, but what happens when the playbook is rerun? It will break existing installations since mailcow has to be shutdown first before updating. (default the git ansible module will fetch changes and merge them locally).

How do you handle new config options that are added to mailcow.conf? These are normally added using update.sh which is not used by this Ansible role. Maybe templating the entire mailcow.conf would be a good option or just using update.sh with help of the Ansible shell module and a few lines of bash.

How do you configure all other config options in mailcow.conf besides ADDITIONAL_SAN? And on that note why do you add mail.* to ADDITIONAL_SAN?

When rerunning this role the step `name: Generate a configuration` will fail since the generate_config.sh script will check that the mailcow.conf already exists and prompt for a interactive input (to ask if it should be overwritten). You should register a variable and check if mailcow.conf exists if it does you should skip this step.

Also this role does not install docker which is defiantly missing. When automating infrastructure with ansible your playbook should be able to setup a new server from ground up, this means installation docker, docker-compose and also python pip packages needed by other Ansible roles. 

Oh and a example inventory would also be nice to have.

helper-scripts/ansible/roles/mailcow-dockerized/tasks/main.yml
```diff
@@ -0,0 +1,38 @@
...
+  ignore_errors: yes
+
+- name: Add mail.* on ADDITIONAL_SAN
+  replace:
+    path: /opt/mailcow-dockerized/generate_config.sh
+    regexp: "^ADDITIONAL_SAN.*"
+    replace: "ADDITIONAL_SAN=mail.*"
+
+- name: Generate a configuration file
+  shell: ./generate_config.sh
```
ntimo, line 23: You need to add:
```
  args:
    executable: /bin/bash
```
since the generate_config.sh is a bash script and the default shell of the ansible shell module is sh.

helper-scripts/ansible/roles/mailcow-dockerized/tasks/main.yml
ttsak: For some reason it worked for me , but in any case you are right, the correct syntax is as you say, so I fixed this

helper-scripts/ansible/roles/mailcow-dockerized/tasks/main.yml
ttsak: Yes but the script has "#!/usr/bin/env bash" at the top so bash is used to run it

ttsak:
> First off I think that the mailcow-dockerized repo is the wrong place for a ansible role I think it would be better to have a dedicated git repo for this role / playbook (@andryyy what do you think?)

Waiting for your decision about that :grin:

> The mailcow role seems to clone the mailcow-dockerized repo this is fine, but what happens when the playbook is rerun? It will break existing installations since mailcow has to be shutdown first before updating. (default the git ansible module will fetch changes and merge them locally).

No this role as it is will not handle updates but also it will not break existing installations (it would need "force: yes" to overwrite). The default behaviour is to fail if the repo exists and that is why I add "ignore_errors: yes". This is "hackish" and might cause problems in case the git repo is not reachable at the moment, but it is required for any next runs after the first one if this role is inculded in a more general playbook.

> How do you handle new config options that are added to mailcow.conf? These are normally added using update.sh which is not used by this Ansible role. Maybe templating the entire mailcow.conf would be a good option or just using update.sh with help of the Ansible shell module and a few lines of bash.
>
> How do you configure all other config options in mailcow.conf besides ADDITIONAL_SAN? And on that note why do you add mail.* to ADDITIONAL_SAN?

I don't handle them at this point :blush: This is an initial approach with only the required options and anything else will have to be added later. The mail.* is a very common SAN used for mail servers of a domain and that is why I add it by default.
 
> When rerunning this role the step `name: Generate a configuration` will fail since the generate_config.sh script will check that the mailcow.conf already exists and prompt for a interactive input (to ask if it should be overwritten). You should register a variable and check if mailcow.conf exists if it does you should skip this step.

No it won't fail because it uses "creates: mailcow.conf" and if this file exists it will just skip

> Also this role does not install docker which is defiantly missing. When automating infrastructure with ansible your playbook should be able to setup a new server from ground up, this means installation docker, docker-compose and also python pip packages needed by other Ansible roles.

I did not include this functionality in purpose to avoid overloading this repo with irrelevant code. I explain this in the README. Somebody that uses ansible will be able to do these simple package installation things. Also as mentioned in README, this is supposed to be used by someone that already uses ansible for his automation and would like to avoid manual ssh into the server to install mailcow. It was not intended to be a full playbook. And that is why I placed it in the "helper-scripts" directory.

> Oh and a example inventory would also be nice to have.

Same as the previous comment :)

ntimo:
Hello @ttsak,
I have created a Ansible role [here](https://github.com/mailcow/mailcow-ansiblerole) with it you can setup mailcow for the first time and also change its configurations as well as install updates if wanted. I also created a Ansible playbook repo but haven't yet found the time to search for a good galaxy docker-ce and docker-compose role to build the example playbook. I also think that maybe a example playbook is not really necessary since someone who knows Ansible will probably be able to setup the Ansible role himself. Also the role that I created currently only works for Debian based systems since it uses apt to install some requirements. This is on purpose because I am personally not a fan of the package module. I am now going to close this PR. If you want to contribute to the Ansible role feel free. And sorry I missed the `create` statement in my previous comment.

Best wishes,
Timo

PR #3071, opened 2019-10-20
Fix custom http redirects with TLS-SNI

mhofer117:
Disable http listener for SNI ssl hosts in nginx. This allows the use of the following config again:
https://mailcow.github.io/mailcow-dockerized-docs/u_e-80_to_443/

However that documentation page should still be updated: https://github.com/mailcow/mailcow-dockerized-docs/pull/175

### DIFF
--- a/data/conf/nginx/templates/sites.template.sh
+++ b/data/conf/nginx/templates/sites.template.sh
@@ -24,7 +24,6 @@ for cert_dir in /etc/ssl/mail/*/ ; do
   esac
   echo -n '
 server {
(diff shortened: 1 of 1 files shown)


mnin:
@mhofer117 @andryyy how to include the recommended `redirect.conf` https://github.com/mailcow/mailcow-dockerized-docs/commit/4de7e5398cbd9ddd363a7364e45029e5f69f1f46#diff-f200190ffb9aaf0b62d77514249e8591 ?

PR #3076, opened 2019-10-22
[lang.pl.php] Additions and fixes

evilMouse:
Added some missing texts (not all) corrected few errors and made quota naming more consistent (i.e. using "size" or "limit" instead of "value")


andryyy:
Thank you!

PR #3080, opened 2019-10-22
OAuth improvements

mkuron:
- Add the `displayName` attribute to the profile. This is what NextCloud uses to populate its _full name_ field.
- Show a reduced login screen when authorizing OAuth: don't display the buttons for SOGo and other applications, don't show the help, and replace the "mailcow UI" title with "Authorize application". Otherwise the user may get confused and think they are actually going to get logged into the Mailcow UI.

### DIFF
--- a/data/web/index.php
+++ b/data/web/index.php
@@ -31,7 +31,7 @@ $_SESSION['return_to'] = $_SERVER['REQUEST_URI'];
         <div class="panel-heading"><span class="glyphicon glyphicon-user" aria-hidden="true"></span> <?= $lang['login']['login']; ?></div>
         <div class="panel-body">
           <div class="text-center mailcow-logo"><img src="<?=($main_logo = customize('get', 'main_logo')) ? $main_logo : '/img/cow_mailcow.svg';?>" alt="mailcow"></div>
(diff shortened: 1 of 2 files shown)


andryyy:
Thank you!!

fabreg:
Hello @mkuron, 
I tried to integrate Mailcow in NextCloud following this docs 
https://mailcow.github.io/mailcow-dockerized-docs/third_party-nextcloud/
but for some reason it does not work.

Error is:

Provider returned an error: invalid_scope An unsupported scope was requested

Sociallogin plugin looks well configured (tried also to add/remove the "index.php" in the oauth url but it does not work.

Any idea?
Thank you.

mkuron:
I assume that error is displayed by Nextcloud? Please make sure that sociallogin is configured for a scope of `profile` and check whether anything is printed in nextcloud.log inside your Nextcloud data directory.

fabreg:
Yes, correct, it was displayed by NextCloud.
I simply restarted all Docker's container and it started working.
Thank you for your reply.

Il giorno mer 10 giu 2020 alle ore 18:59 Michael Kuron <
notifications@github.com> ha scritto:

PR #3084, opened 2019-10-23
rspamd: Don't remove WHITELISTED_FWD_HOST if SOGO_CONTACT present

mkuron:
This is needed for the force-action that depends on WHITELISTED_FWD_HOST

Some of the others might also need to be changed. For example, SOGO_CONTACT_SPOOFED removes the SPF symbols.


andryyy:
Sorry for not checking this. :( Thanks!

PR #3111, opened 2019-11-01
Update README.md

tallship:
provided much needed information on prerequisite preparations related to port firewalling and DNS resource records, as well as capabilities that MailCow provides as a solution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,14 +1,112 @@
 # mailcow: dockerized - 🐮 + 🐋 = 💕
 
+![mailcow](https://www.debinux.de/256.png)
+
+**Official website: <https://mailcow.email>**
+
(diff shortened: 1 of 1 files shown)


andryyy:
Hi, I appreciate your work, thanks! But we keep this kind of stuff in the docs. :)

tallship:
Oh you're welcome!

I was trying to duplicate the utility of the README.md in the standard app, as it applies to a dockerized version.

I'll take a look for docs/ and update with contents/menu links to those aspects of Mailcow and update the main products README.md with the same for brevity and consistency between the two versions of the product if you like.

Also looking forward to the main version of Mailcow being updated so I can upgrade all of our client's installs with the latest updates :)

PR #2922, opened 2019-09-08
Resolve #2794 Add restore for 'all'

timo-reymann:
As mentioned in #2794 it is a real pain to restore a whole mailcow instance, because you need to select every part of mailcow you wish to restore. 

To improve this behaviour this adds a 'all' option to the restore options as pos zero.

PR #3113, opened 2019-11-01
Make restore not only find .tar.gz but also .gz for mysql backup

timo-reymann:
This pull request provides a fix for my previous pr (#2922).

I noticed there is an issue with the mysql backup not getting restored, because the file extension is .gz and not .tar.gz as the others. While testing this i didnt noticed it, so here is the fix ... simple but powerful.

### DIFF
--- a/helper-scripts/backup_and_restore.sh
+++ b/helper-scripts/backup_and_restore.sh
@@ -220,9 +220,9 @@ elif [[ ${1} == "restore" ]]; then
     exit 1
   fi
 
-  echo "[ 0 ] all"
+  echo "[ 0 ] - all"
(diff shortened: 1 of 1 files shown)


helper-scripts/backup_and_restore.sh
```diff
@@ -222,7 +222,7 @@ elif [[ ${1} == "restore" ]]; then
 
   echo "[ 0 ] all"
   # find all files in folder with .tar.gz extension, print their base names, remove backup_, remove .tar, remove .gz
```
ntimo: Can you maybe update the comment above too? So it reflects the change? :)

timo-reymann: Sorry for the delay ... of course the comment should be changed too. 
I created a commit for this :)

PR #3120, opened 2019-11-03
introduce json-files for translations

tinect:
Benefits: Structure data and add ability using translation managements

Had this idea while checking in translation-files. The blocks are random in every translation file.


andryyy:
This adds more overhead and makes maintaining not easier at all. :( you could transform the php arrays in a more readable format.

tinect:
Oh, you're talking about encoded content? These do not have to be encoded, so it would be more readable and maintainable 🤔

andryyy:
...access_denied": "Acc\u00e9s...


Looks much easier to maintain the PHP files if we just pretty-print them. 

We already have PHP and read it into Json. I don't see the point for turning this around. There are other more important things to be changed.

tinect:
Will update nederlands soon :-)

tinect:
Based on current texts (including https://github.com/mailcow/mailcow-dockerized/commit/62a17737801793088b58a296536428fcb810a931). Added existing and same translations on different keys.

andryyy:
Let's do it. :)

PR #3124, opened 2019-11-04
make footable-toggle bigger for fat finger user like me

tinect:
It's annoying to hit the "plus" in current size of 14px on mobile-device :-)

### DIFF
--- a/data/web/css/build/008-mailcow.css
+++ b/data/web/css/build/008-mailcow.css
@@ -52,7 +52,7 @@
 .navbar-nav {
   margin: 0;
 }
-.navbar-fixed-bottom .navbar-collapse, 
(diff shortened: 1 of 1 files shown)


dragoangel:
I think problem not in size. Problem in that it too close to checkbox. Additional space between them will help. Or checkbox can be swapped with ID column. Because ID by itself have no action it will easy to click on + button, and at same time on checkbox too.

andryyy:
Hm. I'm also not sure. Can we move it a bit? :)

tinect:
Good point. But ID isn't in every table :-)
Where can we move it? Increasing space? Or move it to last (new) column (but before action)?

andryyy:
Hmmmm, not sure actually. :( I'd need to try it in the dev console.

PR #3135, opened 2019-11-10
 Fix missing username in username_invalid message

heavygale:
Fix missing username in username_invalid message when adding a new admin or domain-admin:
Schnappschuss 2019-11-10 11 26 21


andryyy:
Thanks!!

PR #3136, opened 2019-11-10
Adding domain admins to domain table

heavygale:
Adds a listing of the assigned domain-admins to each domain in the domain listing. This way it will be possible to filter the domains by domain-admins.

The assigned domain admins are only visible for superadmins. Domain admins won't see the assigned domain-admins in the domain listing as we don't want to reveal other domain admins.


andryyy:
Thanks. :) @mkuron I think it's fine it we eliminate some whitespaces here, what do you think? Perhaps we should finally eliminate all of them. I know it is dangerous for customized setups, but a good code base is something we should look forward to.

mkuron:
I'm fine with minimal whitespace changes like here. Just not thousands of lines, unreadable diffs, etc.

> Perhaps we should finally eliminate all of them.

If we do that, we need some kind of hook script/bot that enforces the style for the future. Otherwise we end up reformatting repeatedly and annoying people with unnecessary merge conflicts.

andryyy:
That's true. Need to think about it.

PR #3137, opened 2019-11-10
add charts.js, remove googlecharts

tinect:
We should get rid of google-charts, as it access and reloads data from google after initialization.
This PR replaces googleCharts with [charts.js](https://www.chartjs.org/) + [chartjs-plugin-datalabels](https://github.com/chartjs/chartjs-plugin-datalabels)

image


tinect:
will change data-displaying like current one :-)

andryyy:
Thanks. :)

PR #3160, opened 2019-11-22
Update lang.ru.json

dragoangel:


MAGICCC:
Merci. Can't speak russian so I will trust you that its somewhat right :)

Issue #3181, opened 2019-12-02
docker-compose binary update


Hackebein:
Please **try** to answer all questions. If you cannot, don't just delete it. If you completely ignore or remove the whole template, your issue is either closed or ignored, too. :( This is (not only) because we are evil. Many awesome mailcow supporters don't bother to reply to an issue, when crucial information is missing. Especially when it could have been answered by using this issue template. We need as much information as possible. Please post all logs, do not copy that one line you think may be interesting. Copy the whole bunch and remove sensible information. :) Thank you!

**Prior to placing the issue, please check following:** *(fill out each checkbox with a `X` once done)*
- [x] I understand that not following below instructions might result in immediate closing and deletion of my issue.
- [x] I have understood that answers are voluntary and community-driven, and not commercial support.
- [x] I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

---

**Description of the bug**: What kind of issue have you *exactly* come across?

https://github.com/mailcow/mailcow-dockerized/blob/master/update.sh#L399
`https://www.servercow.de/docker-compose/latest.php` returns 1.24.1 instead of 1.25.0. 1.25.0 is out for 13 days now.

https://github.com/mailcow/mailcow-dockerized/blob/master/update.sh#L403
This is not working for alpine linux. alpine linux needs "run.sh" instead of "docker-compose-$(uname -s)-$(uname -m)"

I would recommend the use of `file`:
```
# file docker-compose-Darwin-x86_64 docker-compose-Linux-x86_64 run.sh
docker-compose-Darwin-x86_64: Mach-O 64-bit x86_64 executable, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL|PIE>
docker-compose-Linux-x86_64:  ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 2.6.32, BuildID[sha1]=28ba79c778f7402713aec6af319ee0fbaf3a8014, stripped
run.sh:                       POSIX shell script, ASCII text executable
docker-compose-pip:           Python script, ASCII text executable
```

https://github.com/mailcow/mailcow-dockerized/blob/master/update.sh#L395
Maybe we can replace this with the use of `file`. Just because the pip project docker-compose is (not) installed does not mean that we also use this binary in our update process.

[...]

PR #3182, opened 2019-12-02
Improve update-script

Hackebein:
* replace system docker-compose with local docker-compose (#3181)
* add auto-restart for update-script after updating update-script

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,5 +43,6 @@ data/gogs/
 data/conf/sogo/plist_ldap
 update_diffs/
 .github/
+docker-compose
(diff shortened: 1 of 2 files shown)


Hackebein:
I'm still working on this

Adorfer:
if you are working on the update script anyway: Can you include some more sanity checks?
e.g. at least some clear warnings if the updater sees "non mailcow docker containers running" in the system. (that may survive the  update, but might cause problems. so users should at least be warned)

Hackebein:
yes, i think i can do that.

PR #3199, opened 2019-12-05
debug page: disable refresh button, while refreshing

heavygale:
Disable the refresh button, like it is done in mailbox.js, while refreshing the table.

Prevents empty view if the user clicks too fast, as seen here:
doubleclick


andryyy:
Mörci :)

PR #3204, opened 2019-12-08
Update docker-entrypoint.sh

chifu1234:
there is a typo this way unbound wont work properly because the root hints cant be downloaded


andryyy:
This would mean no unbound-mailcow container is working right now. But they are. :) Because `#` is for a progress bar. :)

PR #3228, opened 2019-12-21
[API] Update API docs with app password routes

ntimo:
New API docs for app passwords.

With a special surprise, the dark side awakens...
Screenshot 2019-12-21 at 23 12 02


andryyy:
Thank youuuu

PR #3236, opened 2019-12-24
[API] Added new /system endpoint for restart containers

ntimo:
Added new container endpoint to restart stop and start containers.

Example POST:
```
curl --request POST \
  --url https://demo.mailcow.email/api/v1/system/containers \
  --header 'content-type: application/json' \
  --header 'x-api-key: yourapikey' \
  --data '{
	"container": "sogo-mailcow",
	"action": "restart"
}'
```


andryyy:
The status route should be part of /system then. You also copied a process_delete_return function, that is not used.
Please move the status stuff to /system then. :) 

Please also use '^(bla|bla)$'  for the regex field.

Issue #594, opened 2017-09-13
"Shared" folder in SOGo doesn't work


evilstiefel:
When opening the SOGo interface, the link to "Other users" in the folder list doesn't work. An error-message pops up stating that the communication with the mailserver failed.

Upon inspecting the SOGo-logs, the following entries appear:

`Sep 13 23:28:29 02ac4130b9e5 sogod [100]: [ERROR] <0x0x5654e6779810[NGImap4Connection]> could not select URL: imaps://mail%40*censored*@dovecot/Shared/?tls=YES: {RawResponse = "{ResponseResult = {description = \"Mailbox doesn't exist: Shared (0.001 + 0.040 + 0.039 secs).\"; result = no; tagId = 7; }; }"; reason = "Mailbox doesn't exist: Shared (0.001 + 0.040 + 0.039 secs)."; result = 0; }`

It seems to me that either 
1. the "Shared" folder for new users isn't created properly
2. SOGo shouldn't be showing the link when no folders are shared

The second option seems unlikely, since even after sharing a folder, the recipient of the share still cannot click the folder and the same error is produced.

andryyy:
You need to click the folder icon left to "Other users".

evilstiefel:
Not to sound too offensive, but, did you even read my issue report?

I know that  that's a common UI problem people have with SOGo, this is **not** the case here. 

There is an issue the way folder sharing is handled right now in the backend. We discussed it last night on Freenode and multiple people reproduced it on their setups.

andryyy:
It is a shared namespace, not a real folder. SOGo still shows it as a folder. You can report that to SOGo.

mkuron:
@evilstiefel, you need to click on the folder icon next to "Shared" to expand it to see the shared folders. If you click on "Shared", you get that error because SOGo doesn't know the difference between a folder and a namespace.

andryyy:
Damn it. This was a strange problem related to inbox sharing. :/

PR #3237, opened 2019-12-25
[Dovecot] Fix shared namespace typo

Keessaus:
I noticed the shared namespace was not listed by Dovecot over IMAP and found a typo in the namespace location parameter. Changing the semicolon to a colon fixes the issue.

### DIFF
--- a/data/Dockerfiles/dovecot/docker-entrypoint.sh
+++ b/data/Dockerfiles/dovecot/docker-entrypoint.sh
@@ -195,7 +195,7 @@ namespace {
     type = shared
     separator = /
     prefix = Shared/%%u/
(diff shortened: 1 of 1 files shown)


andryyy:
That's not correct, it works.

It is GOOD that CONTROL was never used. A CONTROL directory must not be per-user. I will remove it. I don't think it is even used anywhere in a shared namespace. Hm.

Keessaus:
Ah, I overlooked that, too. What caused the problem was #594, after I changed the `mail_shared_explicit_inbox` setting. It's not only a SOGo thing, even IMAP returns the `\NoSelect` tag on namespace folders, even though it should link the inbox to it.

PR #3240, opened 2019-12-28
fix alias_domain_alias_hint in de translation

herbetom:
Fixes a translation error in `alias_domain_alias_hint`. The two example addresses now have the same left part. 

The "current" translation lead to some confusion on my side until i understod that the first example got translated and the second not.


andryyy:
Danke :)

herbetom:
Das ging ja schnell. Gern geschehen!

PR #3270, opened 2020-01-16
Update lang.ru.json

dragoangel:
translated quarantine notification BCC & changed relay_all_info to be more clear


MAGICCC:
thanks!

PR #3271, opened 2020-01-16
Add more info about watchdog notification

dragoangel:


andryyy:
Thanks. Could you please add this to generate_config, too?

generate_config.sh
```diff
@@ -233,8 +233,11 @@ USE_WATCHDOG=n
 
 ALLOW_ADMIN_EMAIL_LOGIN=n
 
-# Send notifications by mail (no DKIM signature, sent from watchdog@MAILCOW_HOSTNAME)
-# Can by multiple rcpts, NO quotation marks
+# Send notifications by mail (sent from watchdog@MAILCOW_HOSTNAME)
+# CAUTION: 1. USE ONLY EXTERNAL EMAILS AS RECIPIENTS
```
MAGICCC: I would do a linebreak here imho

MAGICCC:
@andryyy Well he added it to the generate_config script 🤔

andryyy:
Oops. You are correct.

Thought about update.sh, but that's too much hassle to update for this.

generate_config.sh
dragoangel: Done

dragoangel:
@MAGICCC yes I add it to generate new config only. You want that I write some sed in to update.sh to add this notes to existing configs if they not exist?

Issue #3222, opened 2019-12-19
ACME: Wrong certificate for WebGUI


mkapra:
**Prior to placing the issue, please check following:** *(fill out each checkbox with a `X` once done)*
- [x] I understand that not following below instructions might result in immediate closing and deletion of my issue.
- [x] I have understood that answers are voluntary and community-driven, and not commercial support.
- [x] I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

---

**Description of the bug**: What kind of issue have you *exactly* come across?

A few days before I added an IPv6 address to my VPS. On my reverse proxy I forgot to enable that the Webserver listens also on IPv6. I think the acme mail container tried to renew the certificate and got no HTTP Verification. So he set the certificate of a ADDITIONAL_SAN as the current SSL. So I fixed the problem with the reverse proxy so he's listening also for IPv6 now. After that I restarted the complete mailcow mailserver. The certificate for IMAP is working again now.

My problem: 
When I try to access the mailcow Web UI I get a SSL error. The SSL certificate which the Web UI tries to use is from a domain which is set in ADDITIONAL_SAN. So I looked at the acme logs:

```
mv: can't rename '/var/lib/acme/cloud.mydomain': No such file or directory
Thu Dec 19 21:43:12 CET 2019 - Found orphaned certificate: webmail.mydomain - archiving it at /var/lib/acme/backups/webmail.mydomain/
mv: can't rename '/var/lib/acme/webmail.mydomain': No such file or directory
Restarting 7a7c22ef9d3b06cbc8123519a1f65edc99c483b1f46d25b038dd773f3b226b71...
command completed successfully
Restarting 6ca88e0f7b6efbedbd640029d97731df293c155621776bcb674317a09c93b399...
command completed successfully
Restarting aa75bdad8c3dbd58c3fb61c101eee4b352c46bba595a6398b4feb864e7cfdcae...
command completed successfully
Thu Dec 19 21:43:20 CET 2019 - Certificates successfully requested and renewed where required, sleeping one day
```

I am also running nextcloud within the mailcow mailserver. This certificate is set right. Why are the errors appearing?
What is the acme container trying? I also looked at the directory at /var/lib/acme and the above mentioned directories are existing. This is a little bit confusing.

**System information**

Further information (where applicable):

| Question | Answer |
| --- | --- |
| My operating system | Ubuntu 18.04 |
[...]

PR #3272, opened 2020-01-17
[SSL] fix bug with pruning old certificates

mhofer117:
In some cases, pruning old certificates was not working as the base directory for backing it up did not exist yet.
The base path for a cert backup is created when renewing certificates - if this never happened for a specific certificate the pruning would fail.
Now it will also create the path while pruning.

Edit: probably related issue: https://github.com/mailcow/mailcow-dockerized/issues/3222


andryyy:
merci :)

PR #3273, opened 2020-01-17
rspamd: More comprehensive attachment handling

### LINKED ISSUE #3266: Question : DOC files rejected by rspamd and sieve filter
Dear devs & contributors,

why did you chose to bannish the `.doc` files, and not the `.docx` . My users complained about not receiving mails with some attachments and I quickly discovered that this rule has been recently adjusted on December 20, 2019, https://github.com/mailcow/mailcow-dockerized/commit/5a0df09361690ec4b9cb1cbad8a4a34c0f40bc59#diff-9980851224bcd2f7afaed16965d33376.  I see this extension rejected in RSPAMD rules, and when reviewing code there is also a sieve filter https://github.com/mailcow/mailcow-dockerized/blob/d468ecfa65b01f1482eeb3601f458087476ced43/data/web/inc/presets/sieve/sieve_1.yml.

Thanks a lot for your lights. This is a big change for our userbase so I'd like to know how to argument  ;-)

Thanks for your great work !

mkuron:
- block all Office documents with macros
- don’t just block all doc files
- mark some more Windows executable extensions as bad

I have tested this with Word and Excel files, old formats (e.g. doc), new formats (e.g. docx) and macro-enabled new formats (e.g. docm). All files with macros are blocked and all files without macros are allowed to pass.

Fixes #3266

### DIFF
--- a/data/conf/rspamd/local.d/external_services.conf
+++ b/data/conf/rspamd/local.d/external_services.conf
@@ -4,4 +4,6 @@ oletools {
   # needs to be set explicitly for Rspamd < 1.9.5
   scan_mime_parts = true;
   # mime-part regex matching in content-type or filename
+  # block all macros
+  extended = true;
 }
--- a/data/conf/rspamd/local.d/mime_types.conf
+++ b/data/conf/rspamd/local.d/mime_types.conf
@@ -4,13 +4,22 @@ bad_extensions = {
   scr = 4,
   lnk = 4,
   exe = 1,
+  msi = 1,
+  msp = 1,
+  msu = 1,
   jar = 2,
   com = 4,
   bat = 4,
+  cmd = 4,
+  ps1 = 4,
   ace = 4,
   arj = 4,
   cab = 3,
-  doc = 10,
+  vbs = 4,
+  hta = 4,
+  shs = 4,
+  wsc = 4,
+  wsf = 4,
 };
 
 # Extensions that are particularly penalized for archives


andryyy:
Looks good, thank you!

antipode-mj:
Thanks a lot for your great work, this indeed seems to me as the best solution to my question #3266  !

mritzmann:
Thanks andryyy and mkuron for the change. Works fine, documents with macros get a score of 20.

```
OLEFY_MACRO(20.00)[];
```

@mkuron I am trying to understand this pull request. In the file `mime_types.conf` you add more file extensions to be blocked. That seems logical to me. What I don't understand is this: In the second file `external_services.conf` you set `extended` to `true`. Is this the option that turns macro checks on and off? Because if I disable this option on my installation, macros are still rated with 20.

```shell
# Mailcow Version
# f5522a809cd03ceb0d6e05fac337a2c5a7380167

$ git diff data/conf/rspamd/local.d/external_services.conf
diff --git a/data/conf/rspamd/local.d/external_services.conf b/data/conf/rspamd/local.d/external_services.conf
index f05314b..3367eb2 100644
--- a/data/conf/rspamd/local.d/external_services.conf
+++ b/data/conf/rspamd/local.d/external_services.conf
@@ -5,5 +5,5 @@ oletools {
   scan_mime_parts = true;
   # mime-part regex matching in content-type or filename
   # block all macros
-  extended = true;
+  extended = false;
 }

$ docker-compose down && docker-compose up -d
[..]
```

Test E-Mail:

```
X-Spamd-Result: default: False [20.59 / 2000.00];
         [..]
	 OLEFY_MACRO(20.00)[];
         [..]
```

Or tl;dr: What is the easiest and best option to still not check macros?

mkuron:
In standard mode, it only produces the OLEFY_MACRO symbol when the macro does certain things (like executing automatically or calling certain suspicious functions). That catches some dangerous macros, but you could write macros such that they wouldn‘t get detected.

In extended mode, the symbol contains additional information about what suspicious behavior was found. And for us the important thing is that even if no suspicious behavior is found, the symbol is added as soon as a macro is found.

To disable macro checks, remove the entire oletools block from external_services.conf.

mcmufffin:
Hi @mkuron,

since this commits i get false positive from .csv files.

I talked to andre and heinlein support and they said that triggering on suspicious type only is not recommended.

To you think you can change to only trigger on suspicious + autoexec?

Little background:
olevba scanns the .csv file, but olevba does not actually have support for .csv so its triggering a false postive here ("suspicious hex string found")

mkuron:
I would rather keep all macros blocked. I‘ve seen documents that contain instructions how to run the (malicious) macro manually and some people actually do that, so I would consider triggering only on autoexec to be insufficient.

However, we could try sending only the old file extensions (doc, dot, xls, xlt, ppt, pps, pot) to olefy and block the new macro-enabled extensions (docm, dotm, xlsm, xltm, pptm, potm, ppsm) completely.

mkuron:
Or maybe just disable the hex string check? Looking at https://github.com/decalage2/oletools/blob/631a172e4325278991c5955699b575569296f6f5/oletools/msodde.py#L763, it seems like there are malicious things that could be embedded in a CSV.

mcmufffin:
Hi @mkuron,

the autoexec can still require user interaction to activate them (i guess the name is kinda missleading here), so only triggering on suspicious + autoexec should be fine and is the recommended setup by heinlein. 

Mybe you can check your files if the are marked as autoexec?

the hex string check seems legit to me in general, its just a bad for .csv files but mybe we can exclude .csv from olveba but keep cheking it by msodde??

mkuron:
Since macros in email attachments are almost always malicious nowadays, I'd much rather block all macros and have people manually whitelist trusted senders. Very few people need macros, so I think it's okay to err on the safe side.

I don't trust heuristics intended to detect "suspicious" behavior. There are always ways to write your macros so they don't trigger the detection. So, just block all macros and whitelist trusted senders as necessary to be on the safe side.

I was surprised to learn that CSV and RTF can too be exploited in Microsoft Office (https://github.com/decalage2/oletools/wiki/rtfobj, https://github.com/decalage2/oletools/wiki/msodde, https://payatu.com/csv-injection-basic-to-exploit), so we shouldn't just whitelist CSV and RTF. But we can definitely disable that hex check because obfuscation doesn't matter if there is no way to execute code. Looking at https://github.com/decalage2/oletools/blob/eb3300674609ac4f37d4a2072c2ea910af244a05/oletools/olevba.py, I can see that macro detection (and thus the hex check) is skipped for RTF files (line 2735), but not for CSV files.

I'll see if I can come up with a solution. The problem is that you cannot detect a CSV file via magic number and the file extension doesn't appear to get passed to olevba. olevba just reads the entire csv file, falls back to thinking it's a bare VBScript file and is alerted by strings that look like hex. We might be able to remove the OLETOOLS symbol if there are only csv attachments and only the hex heuristic got triggered.

mkuron:
The default config, which we use, actually doesn't even include csv or rtf files: https://github.com/rspamd/rspamd/blob/a1410d8d5e62886dd6a51784bd96c3d844b0d072/src/plugins/lua/external_services.lua#L53-L89. I don't know why they got passed to oletools. Could you please check what MIME type the attachment was sent as?

mcmufffin:
Hi,
first, thank you for your research and the input!!

Your idea of removing the olefy tag when its an csv attachments  + hex string only get triggered sounds good if skipping the csv file entirly is not possible.

I will post the mime type here when i get it (dont have access to the mailbox right now).

mcmufffin:
Hi @mkuron 

in the headers i only found
MIME-Version: 1.0
Content-Type: multipart/related;

is that what you ment? Im not sure if thats for the attachment.

mkuron:
There should be `Content-Type`s in multiple places, outside of the main headers.

PR #3278, opened 2020-01-19
[CI] Added automated testing using drone

ntimo:
This PR adds automated testing to the mailcow GitHub repo. Tests are run every time a new commit is pushed to the master branch. The test results are published as html files. 

[https://int-test.mailcow.email/results/](https://int-test.mailcow.email/results/)

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,0 +1,120 @@
+---
+kind: pipeline
+name: integration-testing
+
+platform:
+  os: linux
+  arch: amd64
+
+clone:
+  disable: true
+
+steps:
+
+- name: download-tests
+  pull: default
+  image: timovibritannia/ansible
+  commands:
+  - wget $TEST_DOWNLOAD_URL --quiet
+  - unzip -qq mailcow-integration-tests.zip
+  - rm mailcow-integration-tests.zip
+  environment:
+    TEST_DOWNLOAD_URL:
+      from_secret: TEST_DOWNLOAD_URL
+  when:
+    branch:
+    - master
+    event:
+    - push
+
+- name: lint
+  pull: default
+  image: timovibritannia/ansible
+  commands:
+  - ansible-lint ./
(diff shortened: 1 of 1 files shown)


mkuron:
That looks interesting. Where do the tests come from and what do they cover?

andryyy:
Would be nice if you moved the checks to the mailcow org. :)

Timo did an amazing job!

ntimo:
@mkuron 
The tests are downloaded from a private storage system of mine. Since they contain a lot of secrets and passwords. I don't want to have them in a public Github repo for now. Also this allows me to run the tests before uploading a newer version of them. This way I can test the tests you could say. The best way to see what the tests covers would be to check the results of one of them for example: https://int-test.mailcow.email/results/20200120T192813_f1f8a48fa3.html

Basically what happens when ever a new commit gets added to the mailcow repo a new Hetzner cloud server is started and setup. Then the mailcow repo gets cloned and the images are build from source then the stack is started. After that the tests are run, and the results get uploaded as a html to the web service provided by André. Thats also why the tests take on average 30 minutes to run. I plan on adding more and more tests in the future. If you have suggestions for new tests let me know.

The tests are then run on a small Drone server that I have setup.

mkuron:
> Since they contain a lot of secrets and passwords.

These could all go into private variables.

> I don't want to have them in a public Github repo for now.

If anyone makes a change or pull request and see that a test fails, they‘ll want to look at the test and understand what it tests and why it fails. If we only see the test name, debugging test failures is very difficult.

> Also this allows me to run the tests before uploading a newer version of them. This way I can test the tests you could say.

You could do that by having the tests in a separate repo that has its own CI job.

andryyy:
Yes, the checks should be public indeed. :)

ntimo:
@andryyy going to switch the check logs to public, once I had time to check something.

PR #3280, opened 2020-01-20
Added data/web/rc* to ignore list

Houtek:
When users add Roundcube (explained here: https://mailcow.github.io/mailcow-dockerized-docs/third_party-roundcube/), they have a lot of files in 'git status' showing as 'not tracked'.


andryyy:
Thanks :)

PR #3304, opened 2020-01-30
Change want_spam to list of symbols_enabled

dragoangel:
`want_spam` disable all mail processing by RspamD, so DKIM signing is not applied for any mail where rcpt is `/postmaster@*/`, including external one.
So `greylist = null` and list of `symbols_enabled` used for outgoing mail doing same job as `want_spam`, but not broke DKIM & ARC.


andryyy:
Thanks, I also removed the ratelimit stuff. :)

dragoangel:
Ok, but ratelimit used only for outgoing mail. So I think it fair to be here. If you think to better remove it - ok =)

andryyy:
Also bounces to postmaster for those who set lower rate limits. :) better safe than sorry

PR #3307, opened 2020-01-31
[Nginx] custom redirect from location / ie webmail.domain to /SOGo/

awsumco:
Added the new to add subdomain redirects for example if a domain is being moved from cPanel the end users where used to using webmail.DOMAIN. I did not use sites.*.custom as the redirect needs to go into location / to avoid any endless http loops. 

example redirect:
==> data/conf/nginx/redirects.webmail.custom 
if ( $host ~ ^webmail\.(?<domain>.+)$ ) {
        return 301 https://webmail.$domain/SOGo/;
 }

### DIFF
--- a/data/conf/nginx/includes/site-defaults.conf
+++ b/data/conf/nginx/includes/site-defaults.conf
@@ -46,6 +46,7 @@
 
   location / {
     try_files $uri $uri/ @strip-ext;
+    include /etc/nginx/conf.d/redirects.*.custom;	
   }
 
   location /qhandler {


andryyy:
Hi, create a new site, please. This is part of the docs. :) Webmail is even an example there. :)

awsumco:
Oh, that's odd, that was not in the docs a few days ago when I was looking. Only reason why I did it this way was to do a catchall for any host header. 

Anyway, you can remove the PR and I will fix my repo :D 

Who do I speak to about getting more involved with this project, i am keen to help.

PR #3309, opened 2020-02-01
Rename docker-compose.yml to docker-compose.tmpl

openemail:


ntimo:
May I ask why you renamed the docker-compose.yml to docker-compose.tmpl?

openemail:
> May I ask why you renamed the docker-compose.yml to docker-compose.tmpl?

I wanted to generate docker-compose.yml at the time of installation using docker-compose.tmpl which can contain docker image tags as variables with an a intention to maintain versions

andryyy:
Nice to see all our hard work is copied and sold under a different name...

PR #3311, opened 2020-02-03
Reduce Rspamd DNSBL false positives

mkuron:
- Ignore SORBS for forwarding hosts. Recently, one of gmx.net's servers was listed on SORBS. I have that whitelisted as a forwarding host with spam filtering enabled, so I got lots of false positives there. When you trust a server enought to list it as a forwarding host, you don't need to check SORBS (since that might tag all messages coming from there as spam).

- Disable Spamhaus XBL for `Received:` headers. XBL lists "compromised" IP addresses, but that can include addresses that ISPs assign to their customers. Even though these should actually be on PBL, there are cases where an address is on the "wrong" list. This can cause emails to get flagged as spam that simply got sent from an end-user ISP address (unless of course the sender's email server anonymizes its `Received:` headers).

### DIFF
--- a/data/conf/rspamd/local.d/composites.conf
+++ b/data/conf/rspamd/local.d/composites.conf
@@ -34,6 +34,6 @@ FORGED_W_BAD_POLICY {
   expression = "( ~g+:policies | ~R_SPF_NA) & ( ~FROM_NEQ_ENVFROM & ~FORGED_SENDER )"
   score = 3.0;
 }
-RBL_SORBS_EXCLUDE_FWD_HOST {
-  expression = "-WHITELISTED_FWD_HOST & ^RBL_SORBS_RECENT"
+RBL_EXCLUDE_FWD_HOST {
+  expression = "-WHITELISTED_FWD_HOST & ^g:rbl"
 }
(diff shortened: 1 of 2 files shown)


andryyy:
Shouldn't we disable the whole rbl group on fwd hosts then?

mkuron:
SORBS_RECENT is the only SORBS list that a reasonably well operated mail server should be able to get listed on by accident, so excluding that one should suffice. The other lists require serious server misconfigurations, while RECENT only needs one customer that sends spam. But we could disable the entire SORBS group for forwarding hosts; it shouldn‘t make any difference.

Maybe we should exclude UCEPROTECT too? I don‘t know what criterion they use for listing.

andryyy:
UCE and SORBS sounds reasonable then.

mkuron:
Done. According to https://github.com/rspamd/rspamd/blob/f5b203c98ace2eb818f8497af293bea2967ed83d/conf/groups.conf#L36-L40, all RBLs are in a symbol group called `rbl`.

andryyy:
Perfect. :) Thanks!

Issue #3018, opened 2019-10-08
Mailcow UI font is not multilingual friendly.


evilMouse:
**Prior to placing the issue, please check following:** *(fill out each checkbox with a `X` once done)*
- [X] I understand that not following below instructions might result in immediate closing and deletion of my issue.
- [X] I have understood that answers are voluntary and community-driven, and not commercial support.
- [X] I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

---

**Description of the bug**: What kind of issue have you *exactly* come across?

Some national (accented) characters in Mailcow UI look distinctly different than non-accented characters.
SoGo is not affected.

**Reproduction of said bug**: How *exactly* do you reproduce the bug?

Just change UI language to Czech or Polish - you will see that accented characters are notably different: a little bolder and bigger.

__I have tried or I do...__ *(fill out each checkbox with a `X` if applicable)*
- [X] In case of WebUI issue, I have tried clearing the browser cache and the issue persists.
- [ ] I do run mailcow on a Synology, QNAP or any other sort of NAS.

**System information**

Further information (where applicable):

| Question | Answer |
| --- | --- |
| My operating system | Debian 10 |
| Is Apparmor, SELinux or similar active? | No |
| Virtualization technlogy (KVM, VMware, Xen, etc) | Hyper-V |
| Server/VM specifications (Memory, CPU Cores) | 4GB,2 |
| Docker Version (`docker version`) |  18.09.1 |
| Docker-Compose Version (`docker-compose version`) | 1.24.1 |
| Reverse proxy (custom solution) | No |

As a side note: I'd like to contribute to Polish translation. How can I do this?

andryyy:
Hi, thank you. I will try to fix that. :)

You can find the lang files in data/web/lang.

andryyy:
Any idea for an alternative font, free to use, compatible etc?

evilMouse:
I'm not much UI guy, but maybe the one used in SOGo would be OK?

evilMouse:
> You can find the lang files in data/web/lang.

I know, I've already updated mine. :)
I'm asking if I should just paste new lines here or submit them somewhere (I'm also new to github :) ).

andryyy:
You should be able to edit the file on Github browsing the repo. It will 
create a PR then.

------ Originalnachricht ------
Von: "evilMouse" <notifications@github.com>
[...]

PR #3333, opened 2020-02-15
[Web] Add latin-ext to PT Sans font #3018

Hexaris:


andryyy:
Oh, cool, let me check this later. :) thank you very much!

PR #3383, opened 2020-03-03
Fix restore script not restoring MySQL database on restore

kodehat:
Yesterday I moved my mailcow to a new maschine by using the provided helper script. After restoring the made backup, I found out that the MySQL database hadn't been replaced and wondered why. So I tracked down the problem and saw that the MySQL database wouldn't get restored if it's a `mysql` folder instead of a `.gz` file. So I changed the behavior of the restore script as shown in this PR. I add two images that illustrate the problem.

This one is missing the `mysql` folder.
not-working-version

After the fix, the `mysql` folder is included.
working-command

### DIFF
--- a/helper-scripts/backup_and_restore.sh
+++ b/helper-scripts/backup_and_restore.sh
@@ -230,14 +230,14 @@ elif [[ ${1} == "restore" ]]; then
   echo
   declare -A FILE_SELECTION
   RESTORE_POINT="${FOLDER_SELECTION[${input_sel}]}"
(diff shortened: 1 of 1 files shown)


andryyy:
Thank you!

Lazy me just removed the type flag completly. Do you want to add it to the previous find command, too? :)

kodehat:
Sure. Additionally added the `maxdepth` option to the second command as shown in the image. Somehow I forgot to include it in the first commit.

Great project by the way! 👍 😍

andryyy:
Thank you! And thank you! 😄

PR #3386, opened 2020-03-04
Aliases in SoGO will be sorted this way

kic68:
The sender drop down list when writing a new email in SoGO will be sorted with this patch. Currently they are in a pretty random order. I had to manually drop and recreate the view, not sure how to trigger this in Mailcow.


andryyy:
Thanks. :)

PR #3391, opened 2020-03-05
add more signatures DBs

Clevero:
Add more signature databases that are listed with a low false-positive rate on https://sanesecurity.com/usage/signatures/

### DIFF
--- a/data/Dockerfiles/clamd/clamd.sh
+++ b/data/Dockerfiles/clamd/clamd.sh
@@ -73,6 +73,22 @@ while true; do
       --include 'rogue.hdb' \
       --include 'sanesecurity.ftm' \
       --include 'sigwhitelist.ign2' \
+      --include 'foxhole_generic.cdb' \
+      --include 'foxhole_filename.cdb' \
+      --include 'bofhland_cracked_URL.ndb' \
+      --include 'bofhland_malware_attach.hdb' \
+      --include 'bofhland_malware_URL.ndb' \
+      --include 'bofhland_phishing_URL.ndb' \
+      --include 'porcupine.hsb' \
+      --include 'porcupine.ndb' \
+      --include 'winnow_malware.hdb' \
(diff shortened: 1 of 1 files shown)


mkuron:
Good idea, but every additional signature database increases ClamAV's memory footprint. Please check how much additional memory consumption and how many additional signatures come with these databases.

andryyy:
False positives do also increase with these signatures, what pushes users to disable ClamAV or whitelist too many senders.
Not sure about this PR. Memory requirements will also increase, yes.
You cannot even test it properly without merging and collecting mulitple users feedback on this. I currently prefer to keep it as it is, hm.

Clevero:
So then I need to know if I should close this PR as a hole or remove most of the additions and leave only the ones where we personally know that these have very little false-positives rates

mkuron:
Can you give estimates on false-positive rate and memory consumption for the different databases?

Clevero:
The false-positive rate should be low as they are listed at sanesecurity with low. Also we have the following ones prodution for several months and had no false-positive so far with more than 30 corporate mailboxes.
```
bofhland_cracked_URL.ndb
bofhland_malware_attach.hdb
bofhland_malware_URL.ndb
bofhland_phishing_URL.ndb
foxhole_filename.cdb
foxhole_generic.cdb
porcupine.hsb
porcupine.ndb
spamattach.hdb
winnow.attachments.hdb
winnow_bad_cw.hdb
winnow_extended_malware.hdb
winnow_malware.hdb
winnow_malware_links.ndb
```

For the question about memory consumption I cannot give exact values for every DB. The clamd docker container on our side is using 1.154GiB of RAM at this moment. But its also using several other DBs from https://www.securiteinfo.com/

mkuron:
If you don't have before and after memory consumption numbers, could you please check how many signatures are in each new database and how many signatures are in the databases we already use? Memory consumption should be roughly proportional to that.

Issue #3129, opened 2019-11-06
Normalize line endings


MAGICCC:
**Is your feature request related to a problem? Please describe.**
We should enforce LF as line endings to avoid problems with 'invisible chars' as line endings

**Describe the solution you'd like**
Add a `.gitattributes` with entries for all types of files ending with `text eol=lf` ala `*.conf text eol=lf`

**Additional context**
https://help.github.com/en/github/using-git/configuring-git-to-handle-line-endings

ClundXIII:
Quick question: what is the preferred line ending if I want to contribute code right now? Can I just ignore the `^M` from git diff until this is resolved?

PR #3405, opened 2020-03-10
Normalize line-endings

mirorauhala:
This PR fixes issue #3129 

It introduces a new file `.gitattributes` that enforces proper LF line-ending. This change also fixes an issue with the nginx container being in a "restarting" state due to CRLF line-endings. I came across this when installing mailcow under Windows 10 with Docker.

Also this PR touches every file to update the line-endings.


MAGICCC:
Derp I totally forgot about this issue which was mentioned once in our telegram group :/

mirorauhala:
No worries! I had to fix this as I couldn't get mailcow to run under Windows 10. My mac instance worked just fine.

mirorauhala:
Also maybe touching every file was a bit unnecessary as they would have updated later. Come to think of this, me touching every file will most likely trigger a merge-conflict.

andryyy:
I don't think we should merge this for various reasons...

mkuron:
This is going to cause lots of merge conflicts. Is that really worth it?

Does Mailcow even work correctly on Windows 10? The networking is completely different, so you might be losing source IPs, fail2ban and IPv6 support.

MAGICCC:
Hm shouldnt the attribute file avoid merge conflicts due to line endings anyway?

mirorauhala:
Yeah in hindsight it was bad idea to touch every file. I will revert the file update so this PR will only add the `.gitattributes` file.

mirorauhala:
> Does Mailcow even work correctly on Windows 10? The networking is completely different, so you might be losing source IPs, fail2ban and IPv6 support.

Well I'm running it under Docker so everything built under the docker-compose should work just fine. At least that's the idea of Docker. Anyway this is my development machine right now and CRLF line-endings were the culprit as bash scripts should have LF.

mirorauhala:
Any update on this?

PR #3428, opened 2020-03-23
Quarantine notifications - don't send if sender is blacklisted

feldsam:

### DIFF
--- a/data/Dockerfiles/dovecot/quarantine_notify.py
+++ b/data/Dockerfiles/dovecot/quarantine_notify.py
@@ -97,7 +97,10 @@ def notify_rcpt(rcpt, msg_count, quarantine_acl):
       print('%s'  % (ex))
       time.sleep(3)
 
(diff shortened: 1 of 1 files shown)


andryyy:
Hi,

Hm, how will that work with `*@bla.com` or `asd*@bla.com`?
How about a sender `bla@bla.com` while only `asdasd@bla.com` is blacklisted?

feldsam:
Hi, thanks for pointing that. I fixed `*@domain.tld` senders, but `asd*@bla.com` is more complicated. In case of `asd*@bla.com` quarantine notification will be send.

feldsam:
Ok, I replaced sender part by REGEX and looks like all cases are working

andryyy:
Oh, cool. Let me check that tomorrow. :) Thanks!

andryyy:
@mkuron What do you think?

mkuron:
The query looks like it might work, but I'm not familiar with the quarantine mechanism as I don't use it.

PR #3440, opened 2020-03-31
[rspamd] Restore add header forced action

mkuron:
Revert 0474de88b175ef1cb638ef5525376033057be564. I have confirmed that the issue does not occur anymore since c3a4c6d311058103287a4ad95ab2cca0bc51585a.


andryyy:
Thank you, Michael.

PR #3445, opened 2020-04-03
Quarantine release - fix when sender is empty

feldsam:
There was check using `isset` function, which returns false only if an array key not exists or its value is `null`. In our case, key exists and value is an empty string.


andryyy:
Yes, indeed, thanks

PR #3446, opened 2020-04-03
Quarantine notifications - exclude blacklisted sender

feldsam:
The final version, fixed syntax. I see my prev PR was reverted and I also hit the error with syntax problem. In this version, the syntax is fixed and it is working.

### DIFF
--- a/data/Dockerfiles/dovecot/quarantine_notify.py
+++ b/data/Dockerfiles/dovecot/quarantine_notify.py
@@ -97,7 +97,13 @@ def notify_rcpt(rcpt, msg_count, quarantine_acl):
       print('%s'  % (ex))
       time.sleep(3)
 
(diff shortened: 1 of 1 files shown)


andryyy:
Thanks. :) Query works fine, but did you test some cases? :)

feldsam:
Hi, currently in production on two servers. This feature was requested by my client. I tested cases you wrote me in prev PR. We can wait a bit to see if client reports some issues or no.

andryyy:
Okay, sounds cool. Perhaps in 2-3 days? :)

vain90:
It would be gorgeous if a mail from a blacklisted sender would never go to the quarantaine. =)

feldsam:
@andryyy, no issues reported from customers, so I think you can merge this.

PR #3451, opened 2020-04-07
added force mode to update.sh

alxhlz:
Use `--force` or `-f` to prevent questions from update.sh


alxhlz:
i found a bug, will fix that quickly

PR #3453, opened 2020-04-07
added --force mode to update skript

alxhlz:
use `--force` or `-f` to prevent update skript asking questions

`exit 0` means the skript ran successfully
`exit 1` means that some error occurred
`exit 2` means that update.sh changed and the skript should be triggered again


andryyy:
Merci

PR #3465, opened 2020-04-10
[Web] Fixed read write API permissions

ntimo:
This fixes a issue where the read and write API key would not work.
Since the check if the API key is allowed to issue write calls, is using the wrong database table so the value ro and rw are never found and thus the else block just sets every api key to read only.


andryyy:
Merci

PR #3469, opened 2020-04-12
[Web] Translation + Fix

ghost:
- Did a 1:1 match line by line with the german translation file, and found a few strings missing in the swedish file (thety are now matching 1:1).
- Also found and fixed a duplicate entry in the german lang file.


MAGICCC:
Thank you very much!

PR #3477, opened 2020-04-14
[Helper] Added name to backup and restore containers

ntimo:
Name the docker containers used to create and restore backups mailcow-backup. 
This way they will show up better in monitoring tools like Grafana :D .


andryyy:
:D Merci

PR #3481, opened 2020-04-16
[Web] Update slovak translation

Hexaris:
Update lang.sk.json + sorting in dovecot


Hexaris:
After changing the language files it is necessary to recreate the PR.

PR #3482, opened 2020-04-17
Bugfix: chmod +x would fail if needed

copyrights:
`which` only  works for executable files in `PATH`. So `chmod +x $(which docker-compose)` will only work if docker-compose is already executable.

Here a test.
```shell
copyrights@kestrel:~$ which mytest
/home/copyrights/bin/mytest
copyrights@kestrel:~$ chmod -x $(which mytest)
copyrights@kestrel:~$ chmod +x $(which mytest)
chmod: missing operand after '+x'
Try 'chmod --help' for more information.
copyrights@kestrel:~$ 

```


andryyy:
Indeed! Thanks. :)

PR #3484, opened 2020-04-18
Don't delete folder which are not created...

duracell:
... in the script. And changed the order to be consistent.


andryyy:
Thanks, I will include it with the next update of the image. :)

Issue #3489, opened 2020-04-18
Update the base image from backup_and_restore.sh (stretch-slim -> buster-slim)


christianbur:
**Is your feature request related to a problem? Please describe.**
Since I regularly check the up-to-dateness of my used docker images I noticed that the helper-[scripts/backup_and_restore.sh ](https://github.com/mailcow/mailcow-dockerized/blob/a63bad41bf671610c0b748590ba69bae4d10fecf/helper-scripts/backup_and_restore.sh)still uses base image "debian:stretch-slim" but the other mailcow images use "debian:buster-slim".

**Describe the solution you'd like**
I would suggest to change the base image of the backup script to "debian:buster-slim". 

**Additional context**
no

andryyy:
Yep, I plan to do that every day. :(

Do you want to provide a PR?

christianbur:
OK, I create a PR.

christianbur:
PR create  https://github.com/mailcow/mailcow-dockerized/pull/3490/commits/88083757201d96359de09bd2b046e91711409c21

PR #3490, opened 2020-04-18
Update backup_and_restore.sh (debian:stretch-slim -> debian:buster-slim)

christianbur:
Upate debian docker image from "debian:stretch-slim" to "debian:buster-slim".
see also https://github.com/mailcow/mailcow-dockerized/issues/3489

I just tested the script in two of my mailcow instances with backup "all", unfortunately I don't have a test environment for restore.


andryyy:
Thank you. :)

PR #3493, opened 2020-04-19
Update backup_and_restore.sh

suleyman-sahin:
Fix error while "Arguments to -type should contain only one letter" on CentOS


andryyy:
Hi, cool. Is this tested on non-CentOS systems, too? :)

andryyy:
Now it is.

PR #3500, opened 2020-04-23
[Dovecot] fix error redirection at doveconf

chigix:
I find the error redirect from doveconf while dovecot instance is not launched is helpless in current https://github.com/mailcow/mailcow-dockerized/blob/731f5cb354e5015dd3c916749c966d7e60d3ff74/data/Dockerfiles/dovecot/docker-entrypoint.sh , though which seems a typo.


andryyy:
Thanks. :)

PR #3508, opened 2020-04-28
It is unnecessary to run update.sh as root

benoitjpnet:
If the current user is in the `docker` group it is unnecessary to run this
script with the `root` user.

Having `/opt/mailcow-dockeriwed` in `root:root` and executing everything
including `git pull` as `root` should be considered a bad practice for
security reasons.

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -1,9 +1,11 @@
 #!/usr/bin/env bash
 
 # Check permissions
-if [ "$(id -u)" -ne "0" ]; then
-  echo "You need to be root"
-  exit 1
+if groups | grep -vq docker; then
+  if [ "$(id -u)" -ne "0" ]; then
+    echo "You need to be root or in the docker group."
+    exit 1
+  fi
 fi
 
 if [[ "$(uname -r)" =~ ^4\.15\.0-60 ]]; then


mkuron:
Actually, the script updates docker-compose around line 466, which won't work as non-root if it is in the usual place of /usr/local/bin.

> bad practice for security reasons.

That's certainly true, but in a Docker world, it doesn't really matter. Any user in the docker group can take over the host by running a malicious container with `--privileged`, for example. So only put users into the docker group if you would trust them enough to put them into the sudo group.

chriscroome:
I used to always run `update.sh` as the `mailcow` user and manually update `docker-compose` as `root` but then some time ago the script was changed so it had to be run as `root`, I assumed there was a good reason for this at the time?

mkuron:
The only reason was probably to get more people to update their docker-compose. We do support older docker-compose versions though and nobody has to use update.sh (I don't). So consider the update script more of a suggestion that people can adapt to their specific requirements.

benoitjpnet:
> Actually, the script updates docker-compose around line 466, which won't work as non-root if it is in the usual place of /usr/local/bin.

This part of the update script should not be mandatory. Actually I'm using the docker-compose binary from my distribution packaging. We can ask or add a flag to manage this part. (Do you want to update your local docker-compose binary? [Y/n]).

Or, you can vendor the docker-compose binary in the project and add it in the PATH?
There are many ways to prevent using the root account...

andryyy:
Clone it as root, update it as root. Several reasons, several answers of mine previously. 

Doesn't matter to mailcow anyway, as Michael said. It is stupid to change these things, please use your time for open feature requests.

PR #3523, opened 2020-05-07
Minor change to Aliases handling

Programmierus:
Allows to get mailbox Aliases by both ID and address
Returns ID within the msg for newly created Aliases


Programmierus:
I updated my PR with a DB query, as we discussed in Telegram.

As about updating lang files for JSON messages - I don't think it's necessary. With the change I offer here the API reply looks like this:

```JSON
{
   "log":[
      "mailbox",
      "add",
      "alias",
      {
         "address":"login@domain",
         "goto_null":true
      },
      null
   ],
   "msg":[
      "alias_added",
      "login@domain",
      "214"
   ],
   "type":"success"
}
```

Message in the UI doesn't mention this ID, and I think it's perfectly fine. The ID might be useful for somebody interacting with the API and is indeed useless for the end-user. What do you think?

Programmierus:
@andryyy any update on this?

andryyy:
Perhaps update the langs and set %s (%d) there to reflect the id.

Programmierus:
From what I can see, it's already inconsistent (i.e., even %s is not mentioned in some localizations, see below). I still think ID information is useless for GUI user. 

But without arguing much, I'll go ahead and add it to those where %s exists, if you persist. Please confirm. 

image

andryyy:
It isn't really an argument to say "it's already inconsistent". Now that I have seen it I would fix it anyway. So you would need to adjust your PR, too. I don't quite get it. 

Leave it shitty because it already is?

Programmierus:
Due to updated master I've recreated this one here #3572

PR #3524, opened 2020-05-08
Fix pathes -> paths typo in two places

Cellane:
I think `paths` is the correct spelling in this context.

The `apple_connection_profile_complete` string appears in administration panel home page when logged in as a regular user, not admin.
The `oauth2_info` string appears in administration panel home page when logged in as a system administrator.


andryyy:
Thanks :)

PR #3531, opened 2020-05-11
PR: Rspamd default behaviour for rejected mails

AndyX90:
With this option it would be possible to change the current default behavior of rspamd (reject) to silently drop(discard) rejected mails.
In some circumstances with quarantine-enabled environments the reject-mails confuse senders so that they think their mails won't get delivered.
Afaik the action stays "reject" and should not affect other things.

### DIFF
--- a/data/Dockerfiles/rspamd/docker-entrypoint.sh
+++ b/data/Dockerfiles/rspamd/docker-entrypoint.sh
@@ -79,6 +79,10 @@ EOF
   redis-cli -h redis-mailcow SLAVEOF NO ONE
 fi
 
+if [[ "${RSPAMD_DEFAULT}" == discard ]]; then
+echo "discard_on_reject = true;" >> /etc/rspamd/override.d/worker-proxy.inc
+fi
+
 chown -R _rspamd:_rspamd /var/lib/rspamd \
   /etc/rspamd/local.d \
   /etc/rspamd/override.d \
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -82,6 +82,7 @@ services:
         - IPV6_NETWORK=${IPV6_NETWORK:-fd4d:6169:6c63:6f77::/64}
         - REDIS_SLAVEOF_IP=${REDIS_SLAVEOF_IP:-}
         - REDIS_SLAVEOF_PORT=${REDIS_SLAVEOF_PORT:-}
+        - RSPAMD_DEFAULT=${RSPAMD_DEFAULT}
       volumes:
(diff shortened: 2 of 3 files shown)


andryyy:
After x restarts you will have the following file:

```
discard_on_reject = true;
discard_on_reject = true;
discard_on_reject = true;
discard_on_reject = true;
discard_on_reject = true;
discard_on_reject = true;
discard_on_reject = true;
discard_on_reject = true;
discard_on_reject = true;
discard_on_reject = true;
discard_on_reject = true;
discard_on_reject = true;
```

andryyy:
Please add it to `data/conf/rspamd/override.d/worker-proxy.custom.inc` if you need it. I will add it to the docs.

andryyy:
And, furthermore, the quarantine IS for rejected mail. It **should** reject.

I think it is bad to have a quarantine system to accept mail and leave it in a subsystem to be never read and eventually deleted. I hate those systems. So many lost mails, that were accepted by the remote but never delivered. :-/

If we accepted probable-spam, we put it in the Junk folder. That threshold can be controlled. If it was crap, we reject it and inform the sender of the bounce and the rcpt, that a mail was rejected and a copy saved.

In my opinion, a quarantine should not be a grave for mails, that will never be read and the rcpt will actually never look into. The sender will not know his mail was quarantined and probably never read. It is better to reject and evaluate a copy of that message later. No pressure to read the quarantined message.

AndyX90:
Okay, sorry!
Then i will place it in data/conf/rspamd/override.d/worker-proxy.custom.inc, because i have the requirement to silently quarantine emails and there are only a few people who are authorized to decide if the mails get through or not.

mkuron:
> In some circumstances with quarantine-enabled environments the reject-mails confuse senders so that they think their mails won't get delivered.

That's what I was worried about when we introduced the quarantine feature... Traditionally, there are two kinds of messages:
1. Messages that have been rejected by the server. The sending server can act on the rejection accordingly and usually notifies the sender. If the sender receives such a notification, he can thus assume that nobody has received or stored a copy of that message.
2. Messages that have been accepted by the server and delivered to the user (either his inbox or spam folder, but always to the user himself). The sender can thus assume that if a message has been accepted by the server, it has been delivered to the user.

Quarantine has introduced an in-between state: Messages that have been rejected by the server but delivered to some quarantine database that the user may or may not have access to and that may or may not get checked regularly. If you really want that in-between state, it is still better to err on the side of caution and reject quarantined messages instead of accepting them -- that way, nobody can claim that a user received their message, even though it was only received by the quarantine admin.

PR #3548, opened 2020-05-18
Remove obsolete setting smtpd_use_tls.

floli:
See http://www.postfix.org/postconf.5.html#smtpd_use_tls. It is
controlled by smtpd_tls_security_level, which is set to may.


andryyy:
Good catch, thanks.

PR #3553, opened 2020-05-19
Update anonymize_headers.pcre

dragoangel:
Change Received by for Rspamd with Dmarc Reporting module enabled.


andryyy:
Thanks. :)

PR #3561, opened 2020-05-21
[Postfix] Set smtp_address_preference to any

### LINKED ISSUE #3560: Also allow IPv6 for delivering mail
After testing mailcow i noticed that outgoing mail never gets transported over IPv6 and I think thats a bad default.

A solution is to set
```
smtp_address_preference = any
```
in `data/conf/postfix/main.cf` instead of `ipv4` which would allow postfix to at least try to submit the email over ipv6

MyIgel:
This change allows the server to submit mail over IPv6

Closes https://github.com/mailcow/mailcow-dockerized/issues/3560
The corresponding doc change is https://github.com/mailcow/mailcow-dockerized-docs/pull/217

### DIFF
--- a/data/conf/postfix/main.cf
+++ b/data/conf/postfix/main.cf
@@ -172,7 +172,7 @@ non_smtpd_milters = inet:rspamd:9900
 milter_mail_macros = i {mail_addr} {client_addr} {client_name} {auth_authen}
 mydestination = localhost.localdomain, localhost
 # Prefere IPv4, useful for v4-only envs
-smtp_address_preference = ipv4
+smtp_address_preference = any
 smtp_sender_dependent_authentication = yes
 smtp_sasl_auth_enable = yes
 smtp_sasl_password_maps = proxy:mysql:/opt/postfix/conf/sql/mysql_sasl_passwd_maps_sender_dependent.cf


andryyy:
@mkuron Your thoughts?

mkuron:
Fine with me. Nobody _should_ be running without IPv6 these days anyway (people still _do_, but that doesn‘t mean we have to support them), so treating both equally or preferring IPv6 is the right thing to do.

That being said, I don’t think there is any reason not to prefer IPv4 — every mail server on the internet has native IPv4, so there is no performance penalty over IPv6. That is different on the client side, where avoiding NAT and tunneling protocols by using IPv6 makes sense (https://en.wikipedia.org/wiki/Happy_Eyeballs).

andryyy:
My concern was the slower delivery due to dual stack inside the container but ipv4-only in front. Delivery will be slow as it sees a working IPv6 stack and tries v6 first. 

As long as we update the docs it is absolutely fine, I agree. No excuse for v4-only nowadays. We should also push people to avoid providers with no v6 spaces. Having no v6 at home is one thing, but I think you should avoid servers without v6. 

It's not that you are unable to run without v6, it will work fine. Just needs a few (one-time) extra steps.

mkuron:
We could add an IPv6 check and a warning pointing to the documentation in the generate_config.sh script.

remkolodder:
The comment above the change should also be removed (it states, prefer ipv4 for v4-only users, that is no longer the case). In addition, the default is any.. so perhaps the entire statement should better be removed? :-))

Keep up the good work lads, much much appreciated!

andryyy:
Thanks

Issue #2545, opened 2019-04-19
Mem error with /usr/local/bin/quarantine notify.py


mister-man:
**Describe the bug, try to make it reproducible**
A clear and concise description of what the bug is. How can it be reproduced? 
If applicable, add screenshots to help explain your problem. Very useful for bugs in mailcow UI.

I installed mailcow for a few days and configured 4 domains. After a few days the Server runs out of ram (8GB Ram and 8 GB Swap) and cpu time (4 cores 100%).

In htop it seems the ressources are blocked by "/usr/local/bin/quarantine notify.py"

![image](https://user-images.githubusercontent.com/8549816/56435530-0e5e5800-62d9-11e9-8e7c-a1c5b3cd005b.png)

If I run update.sh, the server runs a few days without errors.

Further information (where applicable):
 - Your OS (is Apparmor or SELinux active?)
Opensuse Leap 15

 - Your virtualization technology (KVM/QEMU, Xen, VMware, VirtualBox etc.)
vServer, it seems KVM

 - Your server/VM specifications (Memory, CPU Cores)
4 Cores
8 GB Ram
8 GB Swap

 - Don't try to run mailcow on a Synology or QNAP NAS, do you?
No

 - Docker and Docker Compose versions

 - Output of `git diff origin/master`, any other changes to the code?
No

andryyy:
Hm. Interesting bug. 

Can you try to find the script in the dovecot-mailcow log output?

What happens when you run the script inside the Docecot container as user vmail? Does it finish?
Something like that (in mobile rn):

docker-compose exec dovecot-mailcow /bin/bash
su -s /bin/bash vmail
/usr/local/blabla/quarantine_bla.py

mister-man:
I try it later... the serevr is currently out of ressources. I can't connect via SSH :(
Most time after one or two h it workes

mister-man:
Same error with 

docker-compose exec dovecot-mailcow /bin/bash
su -s /bin/bash vmail
/usr/local/blabla/quarantine_bla.py


Whre can I find the right log?

mister-man:
@andryyy

Which log do you mean?

laurent-so-1:
Maybe this can help anyone, I encountered the same issue on my installation.

Apart having the quarantine system enabled, I implemented _smtpd_helo_restrictions_ on Postfix (`data/conf/postfix/main.cf`) as below:
```
smtpd_helo_restrictions = permit_sasl_authenticated,
[...]

PR #3591, opened 2020-06-03
add increment of count to prevent infinity loop

### LINKED ISSUE #3590: missing increment of count in /usr/local/bin/quarantine_notify.py
I'm not sure, but isn't there an increment of `count` missing in the loop?

https://github.com/mailcow/mailcow-dockerized/blob/e01c24aa051e07c5b666f262b8275b3f36a8977d/data/Dockerfiles/dovecot/quarantine_notify.py#L72

JanMalte:
fix #3590
might even prevent memory error in #2545


andryyy:
Thank you! :)

PR #3592, opened 2020-06-05
fixed backup failing

jonasled:
I had the problem that my backup is failing, becaus I added a container (acme.sh) which is not using the mailcow-network. So docker created the default network. The docker network command now has returned two Networks. I added now that mailcow only searches for the own network.


andryyy:
Looks good, no problem to include that.

PR #3593, opened 2020-06-05
romanian language

Razvan0925:
romanian language translation file


MAGICCC:
Cool!
Also the language needs to be added here: https://github.com/mailcow/mailcow-dockerized/blob/master/data/web/inc/vars.inc.php#L81

You can also add some lines here, https://github.com/mailcow/mailcow-dockerized/blob/master/data/conf/dovecot/dovecot.conf#L90

Anything else where you can/need to add localizations @andryyy?

andryyy:
Nice! :) Thank you very much!

I will add the missing changes.

PR #3595, opened 2020-06-07
Removed api_blueprint docs and use swagger

ntimo:
This PR switches the API docs from API Blueprint to swagger, to make it easier to use them. Since swagger is more widely used and has direct integrations with tools like Postman.


andryyy:
Mörciii

PR #3599, opened 2020-06-08
Update FR lang - acl, oauth, add and user

IsserTerrus:
Translation from english.
Added oauth part and more


andryyy:
Wow, nice! Thank you!

PR #3600, opened 2020-06-10
Removed cd

fabreg:
if you jump in directory and then back up to previous level, script is not able to find mailcow.conf file
```
vmi388707:/opt/mailcow-dockerized# ./install.sh -i
./install.sh: line 29: mailcow.conf: No such file or directory
Subdomain to run Nextcloud from [format: nextcloud.domain.tld]: 
```


andryyy:
That's not correct. The script works as intended. There is no install.sh script in the mailcow root folder. Just run `./helper-scripts/nextcloud.sh -i`.

The `..` part is necessary.

It will detect the correct full path in helper-scripts and jump up to its parent to source mailcow.conf.

The script should not be copied from helper-scripts to the mailcow base dir. :)

fabreg:
I ran the script from /opt/mailcow-dockerized with ./nextcloud.sh -i
But yes probably in helper-scripts folder, works fine.

Il giorno mer 10 giu 2020 alle ore 11:14 André Peters <
notifications@github.com> ha scritto:

PR #3608, opened 2020-06-11
[Web] Added spam-score api docs

ntimo:
Added spam-score api endpoint to docs


andryyy:
kannst du immer direkt mergen. :)

PR #3609, opened 2020-06-12
Make delete-days delete only files with mailcow-* in backup location

floli:
Makes it more secure, so that it doesn't delete files that are unrelated to mailcow. It now mirrors the find command used in restore.


andryyy:
Merci, looks good.

PR #3612, opened 2020-06-14
Updated nextcloud.conf

Oratorian:
Updated nextcloud.conf for nginx to fix the HSTS ( strict-transport-security )warning in the Admin menu.


andryyy:
I will not decide for others to activate HSTS. This PR is dangerous as it also includes subdomains. 

Please don't blindly copy "security" recommendations. :(

I prefer to not cover HSTS in the config file. :)

PR #3622, opened 2020-06-25
Allow activation of own S/MIME Certificates in iOS configuration profile

softcat:
When installing the provided .mobileconfig configuration profile it is impossible to enable (separately managed and installed) S/MIME certificates on the device for this email account. 

This PR enables the user to enable S/MIME signing and encryption, by default it is disabled.  Tested in iOS 13.


andryyy:
Merci! :)

PR #3623, opened 2020-06-26
Speed up Gziping of backup

Ry3nlNaToR:
Hi, currently the backup script is using compression level best (-9) which is good if the backup must take up the least disk space possible,the downside of using that level of compression is it uses quite a bit more cpu time and takes long to complete with very little space saving compared to the default level of 6.

I've done a comparison doing a backup on my server using the currently setting of --best and on the 2nd run editing the script to use Gzip default. 

The backup using best Gzip level took 40% longer to complete then Gzip default level the overall comparison ratio was only 0.1% more.

```
Best
ratio name
94.3% backup_crypt
87.1% backup_mariadb
96.7% backup_postfix
33.2% backup_redis
88.8% backup_rspamd
17.1% backup_vmail
33.3% (totals)

Default
 ratio name
94.1% backup_crypt
87.0% backup_mariadb
96.2% backup_postfix
33.1% /backup_redis
88.1% backup_rspamd
17.0% backup_vmail
33.2% (totals)
```

### DIFF
--- a/helper-scripts/backup_and_restore.sh
+++ b/helper-scripts/backup_and_restore.sh
@@ -81,32 +81,32 @@ function backup() {
       docker run --name mailcow-backup --rm \
         -v ${BACKUP_LOCATION}/mailcow-${DATE}:/backup \
         -v $(docker volume ls -qf name=${CMPS_PRJ}_vmail-vol-1):/vmail:ro \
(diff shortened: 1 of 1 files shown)


andryyy:
Yes, that's fine. Thanks.

PR #3629, opened 2020-07-03
Change autocomplete-behaviour on some forms

gbrune:
This PR adds `autocomplete="off"` and `autocomplete="new-password"` to all forms where it makes sense in my opinion (= create/change user and password forms).

This change prevents the autofilling of usernames and passwords by browsers (tested in chrome and firefox) and saves you from having to empty the input field manually. :)

The changes are only based on my own taste in usability, so it's fine with me if they are not merged for some reason.

Example:

| mailcow-autocomplete-1  | mailcow-autocomplete-2 |
| ------------- | ------------- |

Regarding `autocomplete="new-password"`:

https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion

> **Preventing autofilling with autocomplete="new-password"**
> If you are defining a user management page where a user can specify a new password for another person, and therefore you want to prevent autofilling of password fields, you can use `autocomplete="new-password"`.
> 
> This is a hint, which browsers are not required to comply with. However modern browsers have stopped autofilling `<input>` elements with `autocomplete="new-password"` for this very reason. For example, Firefox version 67 (see [bug 1119063](https://bugzilla.mozilla.org/show_bug.cgi?id=1119063)) stopped autofilling in this case; however, Firefox 70 (see bug [1565407](https://bugzilla.mozilla.org/show_bug.cgi?id=1565407)) can suggest securely-generated passwords, but does not autofill a saved password. See the [`autocomplete` compat table](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete#Browser_compatibility) for more details.

### DIFF
--- a/data/web/admin.php
+++ b/data/web/admin.php
@@ -302,13 +302,13 @@ if (!isset($_SESSION['gal']) && $license_cache = $redis->Get('LICENSE_STATUS_CAC
             <div class="form-group">
               <label class="control-label col-sm-3" for="rspamd_ui_pass"><?=$lang['admin']['password'];?>:</label>
               <div class="col-sm-9">
-              <input type="password" class="form-control" name="rspamd_ui_pass" required>
+              <input type="password" class="form-control" name="rspamd_ui_pass" autocomplete="new-password" required>
               </div>
             </div>
             <div class="form-group">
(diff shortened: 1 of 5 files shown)


andryyy:
Looks good to me, are you fine with it @mkuron ?

mkuron (approved):
Looks good.

PR #3643, opened 2020-07-08
[Web] Allow underscore and hyphen in DKIM selector

Ashus:
This change allows to have cleaner DNS zones as mail security related entries usually begin with _.
Also the default dkim selector was changed to _dkim.
I tested it and Google recognizes this selector without a single hitch.

### DIFF
--- a/data/web/inc/functions.dkim.inc.php
+++ b/data/web/inc/functions.dkim.inc.php
@@ -34,7 +34,7 @@ function dkim($_action, $_data = null, $privkey = false) {
           );
           continue;
         }
-        if (!ctype_alnum($dkim_selector)) {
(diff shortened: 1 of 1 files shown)


andryyy:
Please remove the changed default, we will merge then.

There is no need to make it `_dkim`. It is not a standard nor given by a RFC. I have actually never seen a DKIM like `_bla`.
Furthermore not all security related entries are `_` prefixed. What about SPF TXTs?

Please do not create a PR with a changed default without ever discussing it. :(

Ashus:
Ok, removed the default selector change, sorry for that.
Yes, SPF TXTs are usually without any third level domain, but _dmarc, _mta-sts, _smtp._tls, even _acme-challenge all begin with _.

andryyy:
Merci :)

PR #3645, opened 2020-07-09
Add hints for for private IP ranges

MAGICCC:
Added Wikipedia articles for private IP ranges, because some users can't figure out which IPs are pulic and which are private


andryyy:
\o/

PR #3646, opened 2020-07-10
[Web] Clean PHP code by removing unused variables

mirorauhala:
Small PR to clean the PHP code by removing local variables that are not used. Mainly `$lang` and `$stmt`.


andryyy:
Cool. :) Thanks. What do you think, @mkuron ?

mkuron:
Sure, go ahead and merge. I assume @mirorauhala has tested that none of these variables were needed.

PR #3651, opened 2020-07-11
[WEB] fix undefined elements

mirorauhala:
There's a function `rsettings` but `rsetting` was used. Also removed an undefined `$display` variable.


andryyy:
Thanks. 

The display: hidden attr was replaced by `<div class="panel panel-default <?=($_SESSION['mailcow_cc_role'] == "admin") ?: 'hidden';?>">`

PR #3655, opened 2020-07-14
Allow CIDR notation for API_ALLOW_FROM

monsterry:
The documentation mentions that currently the regex in `php-fpm/docker-entrypoint.sh` does not support CIDR notation for the `API_ALLOW_FROM` variable. As the CIDR notation is already supported via the webinterface I fixed the regex.

See also https://github.com/mailcow/mailcow-dockerized/pull/2855.


andryyy:
Looks good, thanks.

PR #3656, opened 2020-07-14
Allow SOGo branding via WebResources override

monsterry:
We overcame the need to brand the SOGo webapp in mailcow. The easiest way to achieve this is to override parts of the `WebResources` directory. Instead of mounting every single file as an overlay to the container, we used the existing mount `./data/conf/sogo/:/etc/sogo/` and used rsync to override all provided files.


andryyy:
Sorry, I plan to move away from rsyncing the content, it was just a workaround I plan to change in the future. 

Just bind-mount the files you want to change in a docker-compose.override.yml. See https://sogo.nu/support/faq/how-to-customize-the-html.html - you can override specific files in sogo users home.

PR #3661, opened 2020-07-16
backup_and_restore.sh: use pigz for faster compression

maltejk:
- use pigz for faster compression in backups, niced
- obviously need pigz. https://github.com/maltejk/buster-slim-pigz/

Can provide PR for data/Dockerfiles if this change would be considered

### DIFF
--- a/helper-scripts/backup_and_restore.sh
+++ b/helper-scripts/backup_and_restore.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-DEBIAN_DOCKER_IMAGE="debian:buster-slim"
+DEBIAN_DOCKER_IMAGE="maltejk/debian:buster-slim"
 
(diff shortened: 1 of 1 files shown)


xopez:
A compare would be nice.

maltejk:
> On 16. Jul 2020, at 18:24, Sascha Moser <notifications@github.com> wrote:
> 
> A compare would be nice.

pigz runs four times faster here as I have four cpu cores.

andryyy:
I appreciate this, but I don't think we should switch from the official packages for pigz. :/

maltejk:
> On 19. Jul 2020, at 08:10, André Peters <notifications@github.com> wrote:
> 
> I appreciate this, but I don't think we should switch from the official packages for pigz. :/

Understood, thanks for the quick feedback. 

I can certainly build an image via compose if that's helpful. 

Another option certainly would be to stop compressing the mail folder tar archive, as it is compressed on a file base already I guess. 

Will open another PR then.

PR #3664, opened 2020-07-17
Display app_links always

TiiFuchs:
I figured that app links configured in the backend will not get displayed if the array $MAILCOW_APPS in inc/vars.inc.php or vars.local.inc.php is empty...

I changed it so that $app_links does get displayed even if $MAILCOW_APPS is empty.

### DIFF
--- a/data/web/index.php
+++ b/data/web/index.php
@@ -87,23 +87,23 @@ $_SESSION['index_query_string'] = $_SERVER['QUERY_STRING'];
             if (!empty($MAILCOW_APPS)) {
               foreach ($MAILCOW_APPS as $app) {
                 if (getenv('SKIP_SOGO') == "y" && preg_match('/^\/SOGo/i', $app['link'])) { continue; }
-              ?>
+              ?>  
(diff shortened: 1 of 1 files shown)


andryyy:
Hi, 

Thanks. 

This adds some unnecessary white spaces. 

There is also the navbars dropdown left.

TiiFuchs:
I just checked: In the navbar it already shows the app_links, if MAILCOW_APPS is empty, doesn't it?

I also can't see any unwanted whitespace. Can you clarify what you mean exactly?

data/web/index.php
```diff
@@ -87,23 +87,23 @@
...
-                ?>
-                  <a href="<?= htmlspecialchars($val); ?>" role="button" style="margin-bottom:3pt" class="btn btn-primary"><?= htmlspecialchars($key); ?></a>&nbsp;
-                <?php 
-                  }
-                }
-              }
-            }
-          }
-          ?>
+              }
```
mkapra: I think he means whitespaces like behind this bracket.

TiiFuchs:
I finally found the time to remove those trailing spaces from githubs online ide...

andryyy:
Just curious and tired, so I cannot check it myself right now: What happens when both are empty? Links created in the UI and in vars?

TiiFuchs:
It just has the label/legend "mailcow Apps" (or what is configured in the UI modifications in the backend) and an empty space below... (Like before I think...?)

grafik

andryyy:
Okay, thanks. :)

PR #3669, opened 2020-07-22
[Nginx] Refresh cipher suites

Jellyfrog:
Also turn ssl_prefer_server_ciphers off.
"The cipher suites are all strong and so we allow the client to choose, as they will know best if they have support for hardware-accelerated AES" - https://wiki.mozilla.org/Security/Server_Side_TLS#Intermediate_compatibility_.28recommended.29


Jellyfrog:
Bump

andryyy:
Looks good to me.

@mkuron ?

mkuron:
Since it's the Mozilla recommendation, I think it's fine.

andryyy:
My thoughts too. Intermediate is also fine.

andryyy:
@Jellyfrog We had a discussion and we don't really understand why you lowered some of the ciphers. Can you please explain why you chose to replace the current template?

Please do never ever just copy things from somewhere without explanation. I think it was a mistake to blindly trust the PR and make that decision. :/ 

If you can explain each cipher replacement, we will keep it as it is.

Jellyfrog:
I've should have pointed out I've used https://ssl-config.mozilla.org since I noticed  ssl_prefer_server_ciphers was on to simply get the latest recommendations. 
But not really sure what the big fuzz is about, these are still safe ciphers

mkuron:
I think it's fine. It's the current Mozilla wiki recommendation. Over our old config, it adds DHE-RSA-AES128-GCM-SHA256 and DHE-RSA-AES256-GCM-SHA384. It removes ECDHE-ECDSA-AES128-SHA256, ECDHE-ECDSA-AES256-SHA384, ECDHE-RSA-AES128-SHA256, and ECDHE-RSA-AES256-SHA384, which are all CBC -- but we still have the GCM versions of these four ciphers. CBC is old and supposedly less secure than GCM. It has been dropped from TLS 1.3, while GCM is guaranteed available since TLS 1.2. 

Cipher order doesn't matter because that's now up to the client.

mcmufffin:
> Cipher order doesn't matter because that's now up to the client.

Generally yes but this also disables preferred ec curves (which are not set at the moment).

In telegram we came up with something like this:
```
  ssl_ciphers ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305;
  ssl_prefer_server_ciphers on;
  ssl_ecdh_curve X25519:X448:secp256r1:secp384r1;
```

On top we should consider "PrioritizeChaCha" as Openssl option for mobile devices for better battery life. This should be set in general and not for nginx only (e.g. dovecot since its probably profits most)

PR #3672, opened 2020-07-22
Watchdog: Allow sending notifications via Slack webhooks

jacobbaungard:
With this commit we allow watchdog notifications to be sent to a Slack
channel via a Slack webhook (https://api.slack.com/messaging/webhooks).

Notifications are sent to all defined methods. That if both watchdog
e-mail and slack webhook is defined, notifications are sent to both. If
only one of them are defined, only one notification method will be used.

In order to do so, a bunch of if statements were moved to the
`mail_error` function.

### DIFF
--- a/data/Dockerfiles/watchdog/watchdog.sh
+++ b/data/Dockerfiles/watchdog/watchdog.sh
@@ -90,7 +90,6 @@ function mail_error() {
   [[ -z ${1} ]] && return 1
   # If exists, body will be the content of "/tmp/${1}", even if ${2} is set
   [[ -z ${2} ]] && BODY="Service was restarted on $(date), please check your mailcow installation." || BODY="$(date) - ${2}"
-  WATCHDOG_NOTIFY_EMAIL=$(echo "${WATCHDOG_NOTIFY_EMAIL}" | sed 's/"//;s|"$||')
   # Some exceptions for subject and body formats
   if [[ ${1} == "fail2ban" ]]; then
     SUBJECT="${BODY}"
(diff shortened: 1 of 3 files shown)


data/Dockerfiles/watchdog/watchdog.sh
```diff
@@ -90,38 +90,46 @@ function mail_error() {
...
+        --hello-host=${MAILCOW_HOSTNAME} \
+        --ipv4
+        #--server="${RCPT_MX}"
+      log_msg "Sent notification email to ${rcpt}"
+    done
+  fi
+  # send slack msg, too
+  if [ ! -z "$WATCHDOG_SLACK_URL" ]; then
+    content="\"text\": \"${SUBJECT}\",\"blocks\": [{\"type\": \"section\",\"text\":{\"type\": \"mrkdwn\",\"text\":\"*${SUBJECT}*\"}},{\"type\": \"section\",\"text\":{\"type\": \"mrkdwn\",\"text\":\"${BODY}\"}}]"
+    curl -X POST --data-urlencode "payload={\"username\": \"mailcow-bot\", $content}" $WATCHDOG_SLACK_URL
```
ntimo: I haven't used Slack that much, but does this mailcow-bot user actually have to exist inside of your Slack organisation? And should it be a hardcoded value?

jacobbaungard: Thanks, good point. Turns out it is not needed to set the username at all, as it is set when the incoming webhook is created. I've amended the commit.

jacobbaungard:
Any thoughts whether this feature is something you'd like to include?

andryyy:
Hi,
Can we leave this open a bit longer? I can try to use this as base for a more general HTTP notification config. Is that fine for you? :)

jacobbaungard:
Sure sounds good. Happy to assist in building something more general as well, just let me know :-)

jacobbaungard:
@andryyy anything I can do to help? If you detail roughly what you are thinking I can try to implement something.

GrantBirki:
👍 This feature would be awesome. I am currently running into an edge case where I believe outgoing SMTP is not possible (Azure VM) and this would be an excellent workaround. https://github.com/mailcow/mailcow-dockerized/issues/4197

jacobbaungard:
Closing as https://github.com/mailcow/mailcow-dockerized/pull/4968 is merged.

PR #3679, opened 2020-07-29
Update lang.fr.json

cl-consulting:
French translation based on the English version


MAGICCC:
Cool thanks
Seems your editor changed spaces with tabs and the whole diff is actually the whole file now.
@andryyy It's not nice, but we could merge it...

andryyy:
Fixed it. :)

We should probably be consistent here and change to 2 spaces. But that's not for today.

Thank you! :)

PR #3680, opened 2020-07-29
Disable SMTPUTF8 in Postfix due Dovecot-LMTP isn't support it

dragoangel:
SMTPUTF8 to work correctly must be done end-to-end. Leaving it enabled now when LMTP can't receive such email gives more issues then profit.


dragoangel:
Without this patch mail to pynnucode domain from Gmail will be rejected with: Relay denied.
Mail from mailcow itself send via Thunderbird (or another UFT8 compatible client) with not decoded pynnucode domain will be rejected with: SMTPUTF8 is required, but was not offered by host.

andryyy:
Yes, we talked about that. Thanks.

PR #3683, opened 2020-07-30
Add urlhaus map with urls to rspamd

Clevero:
Add the list with recent online malware urls from URLhaus into rspamd to check against. In addition to the already included map with recent hashes from Malware Bazaar
https://urlhaus.abuse.ch/api/#retrieve


andryyy:
Merci 👍

PR #3684, opened 2020-07-30
Create mailcow-update-mailbox-acl.sh

dragoangel:
Helper script which can enable or disable ACL for for all users of mailcow instance


dragoangel:
If this will be added to mailcow UI then this script will be useless, but till this, it can help someone =)

PR #3685, opened 2020-07-31
Update dkim_signing.conf

dragoangel:
auth_only is deprecated, new settings which need to be used instead is sign_authenticated


andryyy:
Can you link the docs and commits in the Rspamd repository?

Was it ONLY a change of the parameter name?

dragoangel:
Q:Was it ONLY a change of the parameter name?
A: Yes

Q:Can you link the docs and commits in the Rspamd repository?
A: https://github.com/rspamd/rspamd/pull/3233/files

PR #3687, opened 2020-07-31
SOGo UI: change default filter criterion

### LINKED ISSUE #3221: SOGo UI: change default filter criterion
Hey guys!
Nice-to-have feature request to solve a usability issue.

It would be nice if the SOGo search function would search by "Subject or Sender" instead of only "Subject" by default.

A ticket with a comment with a solution which I did not test does exist: https://sogo.nu/bugs/view.php?id=4555 Is this request something you would consider? Who should be mentioned with code ownership for SOGo?

I'm not sure why "subject only" would even be the default as this is clearly not what a typical user would expect. If any I'd suspect "sender only" to be the more favorable among the two individual ones - with the combination being the intuitive one.

dragoangel:
This resolves #3221 but note: I can't check if this working so it need to be tested before merge first. And this fix is applied to everyone, doesn't think such changes need to be tuned via mailcow.conf actually, they too small.

### DIFF
--- a/data/Dockerfiles/sogo/bootstrap-sogo.sh
+++ b/data/Dockerfiles/sogo/bootstrap-sogo.sh
@@ -238,6 +238,11 @@ chmod 600 /var/lib/sogo/GNUstep/Defaults/sogod.plist
 #  fi
 #fi
 
+# Patch MailUI default search settings
+if patch -R -sfN --dry-run /usr/lib/GNUstep/SOGo/Templates/MailerUI/UIxMailFolderTemplate.wox < /search_by_subject_or_from.diff > /dev/null; then
+  patch -R /usr/lib/GNUstep/SOGo/Templates/MailerUI/UIxMailFolderTemplate.wox < /selected_subject_or_from.diff;
+fi
+
 # Copy logo, if any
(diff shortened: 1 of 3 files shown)


dragoangel:
@andryyy is something need to be adjusted to merge this branch?

andryyy:
I prefer this is changed upstream. You could also mount an override of the template file. 

I don't think we should merge this. :/

dragoangel:
What you mean by change upstream? Like change directly in Sogo original source? I was reused same method as it was done for change the ACLs, there negative impact on this type of modification? I see that this commented out, but I doesn't know why (I saw commit with this change but there no description of why it was commented out). I can try get pull request to their repo but wanted to know why diff method bad at least to not do such in future

By override file you mean override it totally via full file in docker-compose.yml like mount? But this will broke with updates. The diff will work fine or with big update of Sogo it will not be applied due to dry run.

dragoangel:
Ok I tried to update upstream, if they will accept this PR, then we will close this one. Link: https://github.com/inverse-inc/sogo/pull/287

dragoangel:
PR to update upstream https://github.com/inverse-inc/sogo/pull/287 is merged, so this patch no needed.

PR #3708, opened 2020-08-16
Prevent update.sh from crashing docker-compose under Alpine Linux

ghost:
Hi,
i run Mailcow on Alpine 3.12 and if i not pass the --no-update-compose flag while using the update.sh, my local docker-compose installation will be crashed and mailcow wont start correctly after updating.

The reason is, that the docker-compose release, which update.sh installs, is compiled against glibc. But Alpine Linux is using musl and does not support glibc.

To prevent this crash, i added a simple check to update.sh, which tests whether the OS is Alpine Linux and then says, the user should manage his docker-compose via apk (packet manager of the OS). No worry, the docker-compose release in the apk repos aint old stuff, its (normally) up to date.

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -480,6 +480,8 @@ fi
 
 if [[ ${NO_UPDATE_COMPOSE} == "y" ]]; then
   echo -e "\e[33mNot fetching latest docker-compose, please check for updates manually!\e[0m"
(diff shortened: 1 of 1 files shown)


ghost:
Push?
The problem still exists and could crash mailcow instances...

andryyy:
Thank you.

PR #3715, opened 2020-08-20
Use separate .conf file to store secret values

0x326:
This allows downstream users to keep `mailcow.conf` under some version control without worrying about password leakage.


mkuron:
I believe the documentation contains `source mailcow.conf` in a couple places (and maybe some of the helper scripts do too), so this is not an ideal solution. In general, it shouldn't be necessary to keep mailcow.conf in version control though as it contains rather few settings and does not need to be changed often.

andryyy:
Hm. I agree with Michael. :/

But thank you very much for contributing. :)

0x326:
Ok, no problem

PR #3741, opened 2020-09-07
Removed typo when adding a row

mascali33:
Hi there !

When adding row the name of the button is  `"> lang.remove_row` instead of just `lang.remove_row`

Can be showed in App links section of config in the dashboard

It's showing this
image

Instead of
image

Have a nice day !


andryyy:
Oh, thanks! :)

PR #3766, opened 2020-09-18
[Compose] Added SELinux support / volume labeling

shiz0:
Added SELinux labeling info (z/Z) to the volumes.

Having those changes already shipped in the compose file makes it easy to run mailcow on SELinux enabled systems, as only minimal adjustments (which should then be added to the docs) need to be done apart from this. Also, the labeling does not need to be added in an override file, which is error prone and potentially not upgrade safe.

Labeling changes the SELinux context of the mounted files on the host. 
```
/lib/modules
/var/run/docker.sock
``` 
Can/should therefore not be labeled (that's why those are unchanged).

Containers mounting those need to run privileged (netfilter already is), or with SELinux protection disabled (added for dockerapi and ipv6nat) via:

```
security_opt:
  - label=disable
```

If run like this on systems without SELinux, the labeling info seems to be ignored altogether. 
I did a quick test on Ubuntu 18.04 LTS: Fresh installation, added a domain and mailbox, which was also accessible in Sogo, could not spot anything unusual in the logs.

My production mailcow (Centos Atomic Host, SELinux enforcing) has also been running fine like this since yesterday.

Maybe could use another few confirming tests before merge though.


shiz0:
Resolved conflict due to https://github.com/mailcow/mailcow-dockerized/commit/3e457e03eef4a73ad748965db8d2097288ec006d

andryyy:
Merci, will check this tomorrow!

andryyy:
I don't see a problem. What do you think @mkuron - your final words? Any concerns?

mkuron:
Looks okay, but please check what minimum versions of docker-compose and docker are required for the new options to be supported. If those are reasonably old, we can go ahead and merge.

shiz0:
These are not actually new. :-)

Labeling like this has been supported in docker since version 1.7:
https://www.projectatomic.io/blog/2015/06/using-volumes-with-docker-can-cause-problems-with-selinux/
Compose seem to have added support around that time as well:
https://github.com/docker/compose/issues/643

I have bee running mailcow like this (labeled volumes, SELinux enforcing) for over a year now, but with the labels added in an override, instead of the main compose file. :-)

andryyy:
Yes, it is quite "old". :)

karimelc:
I had one issue with the labels on a SELinux enabled system. It's in the volumes section in acme-mailcow:

      volumes:
        - ./data/web/.well-known/acme-challenge:/var/www/acme:rw,Z

I needed to change this to:

      volumes:
        - ./data/web/.well-known/acme-challenge:/var/www/acme:rw,z

Otherwise the LetsEncrypt HTTP validation would fail. I guess this makes sense since Nginx needs to serve this file and with capital Z it's not shared, right?

andryyy:
@shiz0 ? :)

shiz0:
@karimelc 
Correct, thanks for reporting!
Paths/files should not be labeled for exclusive access, if a part of it's path is used by other containers as well.
Sorry, I must have missed/overlooked this conflict (and two other possible) when creating the labeling!
Fixed by https://github.com/mailcow/mailcow-dockerized/pull/3825

PR #3767, opened 2020-09-18
Update quarantine.js

prey87:
changed created formatter like in site/user.js, site/mailbox.js & site/admin.js to get month and day in 2-ditgit format


andryyy:
I think the seconds are fine in this overview. :/

prey87:
The time format don't change.
before:
2.6.2020, 19:41:52
after:
02.06.2020, 19:41:52

andryyy:
Ah, sorry. Got it wrong. Yes, that's fine. But there are other places, too. We should change all of them.

prey87:
found some more in the debug.js and changed them too

andryyy:
Thank you very much!!

PR #3770, opened 2020-09-19
Support mailcow.conf being a symlink

max-baz:
When `mailcow.conf` is a symlink, on every update `sed -i` destroys it as it overwrites the file in place.

I found it useful to keep `maicow.conf` in a separate "infra dotfiles" git repo and use symlink in mailcow folder.

This PR adds an option that lets `sed` preserve such symlink.

### DIFF
--- a/helper-scripts/backup_and_restore.sh
+++ b/helper-scripts/backup_and_restore.sh
@@ -251,10 +251,10 @@ function restore() {
         fi
         echo "Modifying mailcow.conf..."
         source ${RESTORE_LOCATION}/mailcow.conf
-        sed -i "/DBNAME/c\DBNAME=${DBNAME}" ${SCRIPT_DIR}/../mailcow.conf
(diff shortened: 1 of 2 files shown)


andryyy:
Hi Maxim,

Thanks for the PR. We probably need to add a check for busybox sed then as we do with grep for example as I think this option is only available in the gnu version. Or, as alternative, we add a sub command to resolve the symlink via readlink. But we need to check for readlink then.

max-baz:
Thanks, great catch! I added to `update.sh`, but I'm not sure if I should add the same check to `generate_config.sh` (we don't use `sed` there, but maybe that script is used to help people catch missing dependencies early on?), and to `helper-scripts/backup_and_restore.sh` (there GNU `sed` is only needed when you restore DB config and only on an already running instance, but the file has no checks for busybox `grep` and `cp`). What do you think?

andryyy:
Yes, probably better to add it there. :)

PR #3774, opened 2020-09-23
[Web] Add autocomplete to OTP

irgendwr:
This adds support for automatically filling in the TOTP field.

(see https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete, https://developer.apple.com/documentation/security/password_autofill/enabling_password_autofill_on_an_html_input_element)


andryyy:
Cool. :) Thanks.

PR #3633, opened 2020-07-04
DNS: add link for downloading zonefile

mkuron:
Fixes #3567

I know it's not a particularly elegant solution, but I didn't want to re-write the entire DNS diagnostics page.

PR #3777, opened 2020-09-25
Correct syntax for MX and SRV Records

e1mo:
Per spec, both of those records need to have a priority / preference[^1]. And, in the
case of the SRV-Record[^2], it also needs to provide a weight and the
Port **infront** of the domain name. The Validating of the DNS-Entries
is also adjusted to pass as long as the target matches without the need
for a matching priority.

As an example for the domain domain.tld the current behaviour is:

```zone
@	IN	MX	mailsrv.domain.tld.
_autodiscover._tcp	IN	SRV	mailsrv.domain.tld. 443
```

This is bad syntax, the correct on, and as this patch introduces it is
like this:

```zone
@	IN	MX	10 mailsrv.domain.tld.
_autodiscover._tcp	IN	SRV	10 10 443 mailsrv.domain.tld.
```

[^1]: https://tools.ietf.org/html/rfc5321#page-70
[^2]: https://tools.ietf.org/html/rfc2782

### DIFF
--- a/data/web/inc/ajax/dns_diagnostics.php
+++ b/data/web/inc/ajax/dns_diagnostics.php
@@ -161,7 +161,7 @@ if ($_SESSION['mailcow_cc_role'] == "admin") {
 $records[] = array(
   $domain,
   'MX',
-  $mailcow_hostname
+  '10 '.$mailcow_hostname
 );
 if (!in_array($domain, $alias_domains)) {
   $records[] = array(
@@ -172,8 +172,9 @@ if (!in_array($domain, $alias_domains)) {
   $records[] = array(
     '_autodiscover._tcp.'.$domain,
     'SRV',
-    $mailcow_hostname.
-    ' '.$https_port
+    '10 10'. // Priority and Weight
+    ' '.$https_port.
+    ' '.$mailcow_hostname
   );
   $records[] = array(
     'autoconfig.'.$domain,
@@ -208,7 +209,9 @@ if (!in_array($domain, $alias_domains)) {
       $records[] = array(
         '_pop3._tcp.' . $domain,
         'SRV',
-        $autodiscover_config['pop3']['server'] . ' ' . $autodiscover_config['pop3']['tlsport']
+        '10 10'.  // Priority and Weight
+        ' '.$autodiscover_config['pop3']['tlsport'].
+        ' '.$autodiscover_config['pop3']['server']
       );
     }
   }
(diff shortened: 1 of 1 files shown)


andryyy:
I don't think we should print a priority for such custom values. There are valid reasons for having a different priority. Since you cannot set an MX without priority and we don't check against it (and don't need to), I would prefer not to show any priority at all. 

I'd be fine with a hint about priorities for these records. 

If I'm missing crucial points let me know.

When someone tries to force-add an MX without priority (which will never work) and then complains, they should not run a mail server. :/

andryyy:
I'm open for votes. If others prefer your way, we will merge it.

MAGICCC:
I would go with a hint for such records, even when most of the users will run one single mailserver on their domain

e1mo:
> Since you cannot set an MX without priority and we don't check against it (and don't need to), I would prefer not to show any priority at all.

I wasn't sure about the default either, so I'm going to agree with you in that point.

> If I'm missing crucial points let me know.

Since #3633 there is the option do download a zonefile, which I like. But the generated content is, as mentioned in the PR, is invalid. That kind defeats the purpose. While I get your point, I don't think we should create an artificial barrier.

I am, by no means, experienced with operating Mailserver. But I feel as if I have enough knowledge and ability to feel comfy & secure with mailcow. When adding a new domain, it always bothers me a bit. Maybe the question is, should we force users to get even deeper with SRV records, which isn't a direct requirement for your mail-server. I'm sure most users will just not add if if they don't know what it's for. And maybe, at least in my case, they're just going to start entering some random numbers in their DNS-Zone Web-UI. Have they learned anything? Chances are, they haven't. By adding the correct record we would just take out some frustration. And add a bit of comfort for those, who just want to copy & paste into their own Zonefile when operating their own nameservers.

And when we'll be adding some hints about the priority and weight, I think we should start explaining every single record. Operating a Mailserver, even with Mailcow, requires a basic understanding of the stuff you're working with. And I don't think, please excuse me when it sounds supercilious, of a Mail-Server to force sysadmins to educate them about sub specs of minor minor features while withholding some comfort for the experienced. And overall, those inexperienced Sysadmins are probably only running a single mail server.

andryyy:
Hm. :)

It was @mkuron s implementation, he should have the final word. I think he will agree with you. I totally get the zone file part and missed that.

mkuron:
The problem is that we are feeding both the DNS check and the zonefile download from the same data. The two require slightly different information, like in the present case. I would rather not show irrelevant information in the DNS check, but the priority should definitely be added to the zonefile.

andryyy:
Yes, it should be part of the zone file indeed but as easy as possible in the DNS check. The DNS check is not to be seen as visual zone file, it should be as easy as possible. :)

mkuron:
> The DNS check is not to be seen as visual zone file

I agree, but that‘s part of the problem. The zonefile was added later and basically just transforms the DNS check table into something that roughly looks like a zonefile. It misses some special cases and sometimes checkmarks or notes leak into the zonefile.

PR #3783, opened 2020-09-28
[Dovecot] Quota template - better compatibility

Hexaris:
MS Outlook, Thunderbird, eMClient, SOGo, Roundcube


andryyy:
Looks good, thanks!

PR #3784, opened 2020-09-29
Add simplified Chinese language translations

EHfive:
Generally, we use language code 'zh-cn' to represent simplified Chinese, 'zh-tw' in the case of traditional Chinese.
Due to the limitation of https://github.com/usrz/bootstrap-languages used by mailcow UI, there is only code 'zh' avaliable for Chinese. That's why I use 'zh' instead.

Maybe we should switch from bootstrap-languages in future so that we can add support for traditional Chinese.


MAGICCC:
Wew nice!
You can also add some strings here:
https://github.com/mailcow/mailcow-dockerized/blob/master/data/conf/dovecot/dovecot.conf#L90

EHfive:
Oh, I missed that.

EHfive:
> Wew nice!
> You can also add some strings here:
> https://github.com/mailcow/mailcow-dockerized/blob/master/data/conf/dovecot/dovecot.conf#L90

Added.

EHfive:
I think it's ready to be reviewed.

MAGICCC:
LGTM @andryyy

andryyy:
Very cool. Thank you!
Did you check Dovecot is ok with the chinese chars in the config file?

MAGICCC:
It's starting and does not crash for me

EHfive:
> Very cool. Thank you!
> Did you check Dovecot is ok with the chinese chars in the config file?

I don't think it would be a problem because `dovecot.conf` stored as 'UTF-8'.
And, yes, I did't observed any errors.

andryyy:
Let me check that too and then merge it. :)

andryyy:
Thank you!!!

PR #3787, opened 2020-10-02
[Update] Add new args to update.sh -h output

MAGICCC:
Added `--no-update-compose`, `--prefetch` , `-f|--force` arguments to `update.sh --help` output


andryyy:
Mörcii

PR #3796, opened 2020-10-08
Make mailcow.conf in backups accessible to other users

blitzdesigner:
To process the backups further on, like rsync them to another system, one should be able to do this w/o root/sudo, hence giving 644 for mailcow.conf as it has originally 600.

As the datafiles are already accessible by other users, I don't see a security issue here.

### DIFF
--- a/helper-scripts/backup_and_restore.sh
+++ b/helper-scripts/backup_and_restore.sh
@@ -81,6 +81,7 @@ function backup() {
   mkdir -p "${BACKUP_LOCATION}/mailcow-${DATE}"
   chmod 755 "${BACKUP_LOCATION}/mailcow-${DATE}"
   cp "${SCRIPT_DIR}/../mailcow.conf" "${BACKUP_LOCATION}/mailcow-${DATE}"
+  chmod 644 "${BACKUP_LOCATION}/mailcow-${DATE}/mailcow.conf"
   while (( "$#" )); do
     case "$1" in
     vmail|all)


shiz0:
I don't think this is a good idea, there's database credentials in there..
Also, sensitive stuff in the data/ folder is 0600 as well. SSL private keys for example.

blitzdesigner:
Right, but why are then the actual backup files itself 644?
I mean, those are also sensitive stuff more or less.

ntimo:
I think we should then change the permissions of the backup files and not the permissions of the mailcow.conf

shiz0:
> Right, but why are then the actual backup files itself 644?
> I mean, those are also sensitive stuff more or less.

While the .tar.gz files themselves may be 644, sensitive files within them ARE NOT.
See crypt for example:
```
$ tar -tvf backup_crypt.tar.gz
tar: Removing leading `/' from member names
drwxr-xr-x root/root         0 2020-10-05 14:18 /crypt/
-rw------- 401/root        241 2020-10-05 14:18 /crypt/ecprivkey.pem
-rw-r--r-- 401/root        178 2020-10-05 14:18 /crypt/ecpubkey.pem
```
Sorry, but this discussion is ridiculous, I'm gonna close this.
If you feel like you need the mailcow.conf in your backups to be world readable (for whatever reason),
feel free to change it after the backup is completed, but merging this as default behavior is out of question.


> I think we should then change the permissions of the backup files and not the permissions of the mailcow.conf

This, on the other hand, might be worth a thought

blitzdesigner:
all good, I actually didn't checked in the backups, mea culpa.

PR #3806, opened 2020-10-13
Add option to reset a password to nextcloud.sh

heavygale:
This PR adds a option (-r) to set a new password for a given username. There's a hint included, not to use this option if the Nextcloud is configured to use mailcow for authentication. (I added a hint to this new option in the mailcow docs: https://github.com/mailcow/mailcow-dockerized-docs/pull/235)

I also added a missing exit to line 27 ("Cannot use -p and -i at the same time").


andryyy:
Thanks! :)

PR #3810, opened 2020-10-16
[Rspamd] Add DMARC check to Spoofed Unauth

dragoangel:


andryyy:
I made some adjustments, please let us keep this PR open in case something is still not ok.

We quarantine with a reject, so 50.0 for a policy failure should be ok if the header from field was spoofed.

andryyy:
The change I made will also allow forwardings without specifying the fwd host first **as long as no policy is broken**.

Only fully trusted and prefiltering fwd hosts should be added to the fwd hosts. Or those to not align with a policy when those servers cannot be added to the policy for some reason.

dragoangel:
`POLICY_FAILURE` is 10, while it not remove `policies_groups` maybe too big, `policies_group` has already much bigger score then needed. So something like:
```
symbols = {
    "ARC_REJECT" {
        score = 0.1;
    }
    "R_SPF_FAIL" {
        score = 6.0;
    }
    "R_SPF_PERMFAIL" {
        score = 6.0;
    }
    "R_DKIM_REJECT" {
        score = 6.0;
    }
    "DMARC_POLICY_REJECT" {
        weight = 14.0;
    }
    "DMARC_POLICY_QUARANTINE" {
        weight = 6.0;
    }
    "DMARC_POLICY_SOFTFAIL" {
        weight = 1.0;
    }
}
```
and
```
POLICY_FAILURE {
  expression = "-g+:policies & !MAILCOW_WHITE";
  score = 3.0;
}
```
can be more better, as we with not set too big scores and provide chance to other symbols change the game little bit, like `BAYES` or `FUZZY` or `NEURAL` for mail out from mailcow domains, and at same time with big `SPOOFED_UNAUTH` we 100% sure that nobody can to fake local domains without `WHITELISTED_FWD_HOST`.

We also missing interesting AUTH_NA score which by default:
```
AUTH_NA {
    expression = "R_DKIM_NA & R_SPF_NA & DMARC_NA & ARC_NA";
    score = 1.0;
    policy = "remove_weight";
    description = "Authenticating message via SPF/DKIM/DMARC/ARC not possible";
}
```
I not sure what score it must have. I even not sure, do it ok to get such mail at all? I know there people who haven't configured even SPF, but this their fault they not configured their domain to send mail, no? So, maybe 12.0 is ok...

andryyy:
AUTH_NA is absolute fine. Please don't increase its score. That would be bad for many system mails. 

But yes, POLICY_FAILURE is bad. Can you remove it?

andryyy:
Or perhaps add a score of 2.0 okay? So we slightly increase the score for all failed policies.

dragoangel:
You not like rebalance I provided (decrease POLICY_FAILURE and each POLICY)? Also doesn't understand how to trust mail with `AUTH_NA` 🤷‍♂️. I mean that any domain without `SPF nothing -all` can be used for spam.

andryyy:
Because there are a ton of mails without DKIM and SPF. I'm not stupid. That's why we don't increase the score to 12. oO

dragoangel:
Ok, forget about what I said AUTH_NA, yep you are right :( in general, it only in my case maybe I not have many such mail. What you think about other parts?

PR #3817, opened 2020-10-19
[Rspamd] Rebalance group policies

dragoangel:


dragoangel:
Hi @andryyy & @mkuron I lowered `POLICY_FAILURE` from 10 to 1 and some group scores, as it now provide extremely big score.

In summary (if sum 1 group score and `POLICY_FAILURE` score): 
- `DMARC_POLICY_REJECT` will have `17`. and before be0ec8e it was `20`. Now it have `30` score
- All symbols that was has `10` (before be0ec8e) will have `9` score. Now they have `20` score
- `DMARC_POLICY_SOFTFAIL` will be same as before be0ec8e - `2` score, but now it has `12`
- `ARC_REJECT` will be `1.01` while it was `0` before be0ec8e and now it has `10` score

What you think?

andryyy:
lgtm

PR #3820, opened 2020-10-23
[Rspamd] Not use prefilter for GLOBAL_FROM_BL

dragoangel:


dragoangel:
Point of this change to provide Rspamd ability to correctly learn emails (such as `IP_REPUTATION` etc). Vsevolod said that `prefilters` is evil in case of learning.
I think `2000` is ok. It lower then `GLOBAL_(SMTP|MIME)_FROM_WL` by `50` and at same time it greater by `1001` then `MAILCOW_WHITE` which can be set per-user setting. From my view I think that `GLOBAL_(SMTP|MIME)_FROM_BL` must be greater that per-user settings `MAILCOW_WHITE` and at same time good when `GLOBAL_(SMTP|MIME)_FROM_WL` is greater then `GLOBAL_(SMTP|MIME)_FROM_BL`.

andryyy:
I have this change commited in my repository. We talked about it. It is just not pushed yet. :)

dragoangel:
Oh, ok

PR #3821, opened 2020-10-23
[Rspamd] Allow skip quarantine for from regex maps

dragoangel:


dragoangel:
This will allow people to not blacklist annoying TLD for example, but simply ignore spam from them if any.

andryyy:
That's too hacky for me

dragoangel:
I not see any better way to **not ban** for example `XYZ` and some other noisy TLD, but at same time exclude them from getting in Quarantine in case of spam as 99.99% from them is real spam.
In my case they totally blow quarantine. In case if one email from `COM` will be good in bunch of emails from `XYZ` people can miss them simply or start to ignoring or totally disable spam notifications.

dragoangel:
@andryyy also want to note: now `MAILCOW_BLACK` (per-user setting blacklist) isn't exclude email from quarantine.
This is this desired flow or it missed?
UPD: they not excluded, but at same time user will not be notified for email from his own user defined blacklist. This maybe not bad, but not mentioned anywhere in docs or feature description.

P.s. `global_(mime|smtp)_from_exclude_quarantinelist.map` is tested on my mailcow and it works as expected.

Issue #1019, opened 2018-02-10
Switch to a safer password hash


klausenbusk:
[A single round](https://github.com/mailcow/mailcow-dockerized/blob/55586f4cb13ff0f197f6bff4f7646c0428c9bb7f/data/web/inc/functions.inc.php#L4) of sha256 provides close to no security. Off-the-shelf hardware these days can hash [2000 millions hash/s](https://gist.github.com/epixoip/a83d38f412b4737e99bbef804a270c40).

We should switch to something safer (ex bcrypt), which both [Dovecot](https://wiki.dovecot.org/Authentication/PasswordSchemes) and [PHP](http://php.net/manual/en/function.password-hash.php) seems to support.

andryyy:
It is salted sha256.

There were no better alternatives compatible by all used software when we started mailcow with SOGo.

May consider this now.

klausenbusk:
> compatible by all used software

I assume that last missing piece is Postfix? Right now we use a simple SQL query (https://github.com/mailcow/mailcow-dockerized/blob/079170682c5905c56938789c2f4bab3b644c037e/data/Dockerfiles/postfix/postfix.sh#L77-L91) which won't work with bcrypt, but it should be possible to use Dovecot for the SASL auth (https://wiki2.dovecot.org/HowTo/PostfixAndDovecotSASL).

andryyy:
SOGo

Please check the current config. We already use Dovecot as sasl provider. 

Any migration to a new cipher must not destroy current configurations.

klausenbusk:
> We already use Dovecot as sasl provider.

Oh, correct. I'm new to all this :)

> SOGo

SOGo does not support blowfish at the momemt, and there seems to be no plan to implement it: https://sogo.nu/bugs/view.php?id=2682

> Any migration to a new cipher must not destroy current configurations.

Definitely! :)

jkaberg:
From [Dovecot](https://wiki2.dovecot.org/Authentication/PasswordSchemes):
> MD5 based schemes:
>     PLAIN-MD5: MD5 sum of the password stored in hex.
>     LDAP-MD5: MD5 sum of the password stored in base64.
>     SMD5: Salted MD5 sum of the password stored in base64. 
> 
> SHA based schemes (also see below for libc's SHA* support):
>     SHA: SHA1 sum of the password stored in base64.
>     SSHA: Salted SHA1 sum of the password stored in base64.
[...]

PR #3832, opened 2020-10-30
Support of different default pass schemes + support of BLF-CRYPT

monofox:
In #1019 (closed) it was requested to support stronger algorithm (like bcrypt). As stated after a while within the ticket, it was informed, that SOGo support was meanwhile implemented as well.

In urge needing to have this functionality as well prior using mailcow, i've made an implementation here with some points:
- Default is for now still SSHA256 (ensured during update script as well as in docker-compose)
- In mailcow.conf it is possible to change the corresponding default pass scheme.
- Using as default pass scheme, only SSHA256, SSHA512 as well as BLF-CRYPT is supported.

In order to have it working, the corresponding images of mailcow/sogo as well as mailcow/dovecot is required in order to use the new environment variable `${MAILCOW_PASS_SCHEME}` (do we need to increase the tag version numbers here?).

I choose to use an environment variable for the default pass scheme in order to give the admins control to choose stronger or weaker algorithm by themself (e.g. due to hardware / resources, etc.).

For my own tests i created own docker images in `monofox/sogo` respective `monofox/dovecot`.
In my tests, i tested:

- All three default pass schemes mentioned above.
- Tested, that it is still possible to login with old password even the default pass scheme was changed
- Tested, that in all cases, a login is possible with Dovecot (IMAP), SOGo as well as management frontend of mailcow.

This is my first contribution here. Kindly give it a review and a feedback if this is fine, in your mind or changes required.

Thank you!


mkuron:
Thanks for the contribution! I actually don‘t think this needs to be configurable. If BLF-CRYPT is currently the best algorithm and it works with SOGo, Dovecot and our web UI, we can and should use it as the default. I see no reason why anyone would want to downgrade to a worse algorithm, and there is no point to having and maintaining a config option that nobody needs. What is your opinion, @andryyy?

andryyy:
Was thinking the same and want to run some tests. :)

andryyy:
I think we can leave it configurable but mention the supported methods. Blowfish should be standard if it works good.
We should probably make it the default for updated installations, too, so updated mailboxes will generate a new hash.

monofox:
How do you do such password updates here? Through login post hook?

andryyy:
I would keep existing users as they are and only migrate when a password is changed. 

Updates via postlogin hook might be tricky. Not sure it will work with all special chars for example. Better soft migrate.

generate_config.sh
```diff
@@ -117,6 +117,9 @@ cat << EOF > mailcow.conf
 
 MAILCOW_HOSTNAME=${MAILCOW_HOSTNAME}
 
+# Password hash algorithm
```
andryyy, line 120: We could add a link to a new docs page to inform about supported methods, what do you think?

andryyy:
@mkuron - merge? Tested it successfully.

Default can be set to BLF-CRYPT imo. Only new and updated mailboxes will use BLF-CRYPT.

@monofox Can you update docker-compose.yml and increase dovecot-mailcow and sogo-mailcow versions? I already built and push them.

mkuron:
I have not tested it, but it looks good to me. We should default to the new algorithm since it‘s considered "more secure".

andryyy:
https://mailcow.github.io/mailcow-dockerized-docs/model-passwd/

monofox:
Ok, thanks for your feedback.
I did now:

1. Updated default scheme to BLF-CRYPT for new config generation as well as updates
2. In both situations, link to the documentation is embedded to the configuration file.
3. docker-compose.yml increased version for mailcow/sogo to 1.92, mailcow/dovecot is 1.136.

Did i got everything?

andryyy:
Looks good. 

Ready to merge, @mkuron ?

data/web/inc/functions.inc.php
```diff
@@ -84,8 +84,24 @@ function ip_acl($ip, $networks) {
...
+	switch (strtoupper($default_pass_scheme)) {
+    case "SSHA256":
+      $salt_str = bin2hex(openssl_random_pseudo_bytes(8));
+      $pw_hash = "{SSHA256}".base64_encode(hash('sha256', $password . $salt_str, true) . $salt_str);
+      break;
+    case "SSHA512":
+      $salt_str = bin2hex(openssl_random_pseudo_bytes(8));
+      $pw_hash = "{SSHA512}".base64_encode(hash('sha512', $password . $salt_str, true) . $salt_str);
+      break;
+    case "BLF-CRYPT":
```
mkuron, line 100:
```suggestion
    case "BLF-CRYPT":
    default:
```
This needs a default case. Returning an empty password is dangerous. Please also change the comment accordingly.

mkuron (changes requested):

data/web/inc/functions.inc.php
monofox: Done. Thanks for your feedback @mkuron

mkuron (approved):

andryyy:
Yes, looks good. Thank you!

mcmufffin:
Awesome!

Thank you very much.

Two questions as a non-developer:
Would it be hard to implent argon2? Sogo and Dovecot do have support for it now and it would be great alternative.
See: https://github.com/inverse-inc/sogo/commit/4c27826fb57d4b30465e6fb51fddcbc466b6147c

Then also, would it make sense to set the cost option as mentioned here:
https://www.php.net/manual/de/function.password-hash.php
(See example 3)
BCrypt seems to default to 10 otherwise, IMHO it would make sense to set the option according to the systems performance.

Thanks!
Regards

monofox:
Hey  @mcmufffin,

from technical point of view, its not difficult to implement the Argon2i / Argon2Id functionality, as the necessary base was implemented to support further pass schemes.

Implementing configuration of the cost is indeed a bit more of work, as it must be passed to the containers and supported in proper way for all applications. Correct me if i'm wrong, SOGo has implemented 5 rounds by default, am i right? In the documentation i can't find such an option.

mkuron:
To add Argon2 support, check whether our Dovecot, SOGo and PHP images were built with the necessary library. Then, you only need to add a few lines at https://github.com/mailcow/mailcow-dockerized/blob/e1d328eb1a9c2aa5088355387b6d4af3e0bcf60c/data/web/inc/functions.inc.php#L100-L104.

BLF-CRYPT is currently considered sufficiently secure. The number of rounds is determined when the password is set, i.e. by SOGo or PHP. I expect they chose a reasonable number and will increase it in the future when hardware allows and security demands to do so.

mkuron:
SOGo has 5 rounds (non-configurable) according to https://github.com/inverse-inc/sogo/blob/2f75c389f7ebf340558a4dc9107458cc7f71fc17/SoObjects/SOGo/NSData%2BCrypto.m#L71. PHP defaults to 10 according to https://www.php.net/manual/de/function.password-hash.php.

PR #3841, opened 2020-11-05
Add upstream DNS server resolve DNS issues

g0rbe:


andryyy:
That's not a fix at all. :-( It will break all blacklist filtering. Do **not** use a public resolver. It even says so in the docs. Please never do that.

andryyy:
If your network is not able to talk to the root DNS servers, setup a DNS recursor whereever you like and use that server as forwarder. But DON'T use a public DNS as forwarder in that "recursor" (it would not be a recursor then anymore anyway).

PR #3842, opened 2020-11-06
Hungarian translation

balping:
It covers most of the user UI, some of the admin UI.


andryyy:
Nice \o/ Thank you!

Should we wait with merging? Do you want to work on it?

balping:
I was only planning to translate what my clients would encounter, maybe I'll work on the admin panel later; but for now please merge.

andryyy:
Thank you!

PR #3849, opened 2020-11-09
[WEB] Let admins log into Roundcube without password

balping:
**What does it do?**

This PR implements a similar feature as the password-less SOGo admin login: admins and domain admins can login to Roundcube by clicking a button on the admin interface, without having to enter any password.

rc

**Prerequisites**

Admins need to install Roundcube and the plugin [dovecot_impersonate](https://github.com/corbosman/dovecot_impersonate/).

**How does it work?**

PHP reads the `/etc/sogo/sieve.creds` file which contains the password for dovecot's master user. Then it logs in with the master password by sending curl HTTP requests to Roundcube, basically mimicking a browser. Finally it sets the retrieved roundcube authentication cookies for the user and redirects him/her to Roundcube.

Note: if a user is in the *Force password update at next login* state, even admins can't access emails by this method.

---

I'm not sure if you want to include this as Roundcube is not installed by default. Nonetheless too many things need to be modified if you want this feature, so I think it is easier to include the code than writing a manual patch in the docs.

### DIFF
--- a/data/web/inc/lib/RoundcubeAutoLogin.php
+++ b/data/web/inc/lib/RoundcubeAutoLogin.php
@@ -1,0 +1,129 @@
+<?php
+
+/**
+ * Class to automatically login on a Roundcube installation
+ * @compatibility RoundCube 1.0.2+
+ */
+
+// original script by Philipp C. Heckel https://blog.heckel.io/2008/05/16/roundcube-login-via-php-script/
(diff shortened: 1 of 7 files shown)


andryyy:
Hi, thanks, but Roundcube is not part of mailcow. Please move that to the docs.

PR #3851, opened 2020-11-11
[Rspamd] Add xlsm to bad extensions

dragoangel:


mkuron:
We don't have docm, pptm etc. blacklisted either, so I see no reason why .xlsm should be. We use oletools to detect macros inside MS Office files, so no explicit blacklisting is needed.

dragoangel:
Hi @mkuron. If oletools would detect VBS correctly I would not push this change 😄 .
I don't know olefy to troubleshoot it and more to fix it, I can send you mail sample which wasn't been marked as attachment with macros. To note this was virus attachment which clamav also don't see as evil, https://www.virustotal.com/gui/file/c089c5fb3a7bb03adb83957f53ce463c1380e5b7bd0b60ed36c0fefb1e6e72c9/detection

mkuron:
Okay, we prefer to actually fix bugs at the source (i.e. oletools) instead of workarounds like this. I believe @andryyy has previously fixed bugs in oletools (ed9daeb849ea91cfbc8ac3847dcc8f11835e7e1c), so perhaps he could have a look at your sample file?

dragoangel:
Okay, I agree that this more correct way.

dragoangel:
@mkuron have update on issue:
zip archive named as blabla.7z or blabla.rar, etc. with any docs that need to be scanned by oletools can't be understated by olefy. If send pure xlsm it will find marco. I afraid it can be abused more by spammers with any docs types.

mkuron:
Ah, that‘s a good find. We had not considered this scenario. Oletools does scan inside zip files by default (https://github.com/decalage2/oletools/commit/a299cc6d941fe7cf99a999eb081420d0ca6a9eab), but does not support rar, 7z, tar, etc.

I see two possible solutions: we either implement unpacking of the other archive types, or we blacklist all macro-enabled Office formats (a long list, doc, dot, docm, dotm, xls, ...) when found **inside an archive** (via `bad_archive_extensions`). Rspamd support zip, rar, gzip and 7z for this (https://github.com/rspamd/rspamd/blob/b715c7d3ad0cf8d898c4a0246d531c2f43ea6aff/src/libmime/archives.c#L1963). That‘s more than oletools supports, but not comprehensive.

For example, Windows can natively extract cab files and macOS supports bzip2 and a few others. Not even ClamAV supports all exotic archive formats. So there will always be ways to send macros inside an archive in a way that hides them from both rspamd and oletools. For this reason, we might as well ignore macros inside archives entirely. You basically need an on-access virus scanner on the client machine with oletools-like functionality to achive full protection.

dragoangel:
Yeah. Also there is scam that use password protected zips with text or image that show password to user :D. CR said that they plan to create some brute force module for such zips in future with OCR function even maybe to get text from images and try to unzip them. But in general I agree that local AV have much more chance to find bad attachment in any case, no matter you get mail from browser or via desktop client.

mkuron:
One could also consider simply blocking all password-protected archives. I think Gmail does that too. That doesn‘t solve the problem that the mailserver needs to be able to detect all archive formats though. And of course an attacker can just send a download link instead of the archive itself (that‘s again somethig that would need to be detected by a scanner on the client machine). If you really want to be safe, you have to disable macros completely in MS Office.

andryyy:
Blocking macro files in archives is not a good idea imo.

andryyy:
Or... perhaps we can block unsupported archives with macros in them. Every thing that cannot be unpacked by oletools.

Better: fix it upstream.

dragoangel:
Point that any zip which renamed to other archive format is unsupported =) we can't block it, this bug

PR #3879, opened 2020-11-30
[WEB] Optional HTML in system mails

balping:
Adds a textbox to send HTML mass system emails, but plain text emails remain default.

s


andryyy:
Thanks. :)

PR #3283, opened 2020-01-21
Add support for pre hashed passwords via API

cybersholt:
@mkuron this is the patch for #3282 I think my editor ( PhpStorm ) had a field day with the line endings and line spacing, if this is an issue let me know.

PR #3884, opened 2020-12-02
support pre-hashed passwords

drpuur:
PR was closed instead of merged. See https://github.com/mailcow/mailcow-dockerized/pull/3283

### DIFF
--- a/data/web/inc/functions.mailbox.inc.php
+++ b/data/web/inc/functions.mailbox.inc.php
@@ -1054,7 +1054,12 @@ function mailbox($_action, $_type, $_data = null, $_extra = null) {
               );
               return false;
             }
-            $password_hashed = hash_password($password);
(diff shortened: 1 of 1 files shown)


andryyy:
Merci. Perhaps we should add this here: https://mailcow.github.io/mailcow-dockerized-docs/model-passwd/ - can you do that, too? :)

drpuur:
> Merci. Perhaps we should add this here: https://mailcow.github.io/mailcow-dockerized-docs/model-passwd/ - can you do that, too? :)

I am a new user and missed the possibility to import mailbox-passwords from my current setup. I only have them in hashed form, of course. So I'd rather put a note in the API doc.

andryyy:
That's fine.

andryyy:
@ntimo Do you think you can add that hint there? :)

PR #3885, opened 2020-12-03
Korean translation

MangoCubes:
Originally supposed to be focused on what user would encounter, but decided to do a bit more.


MAGICCC:
Wew!
That's very nice. I just tested it in a dev enviroment and looks fine for me. Do you want to get it merged now or wait until you translated every string (I saw some strings not translated yet)?
You may look into this file too: https://github.com/mailcow/mailcow-dockerized/blob/master/data/conf/dovecot/dovecot.conf#L86

MangoCubes:
Thank you! I'd like for it to be merged for now. I do use mailcow on my VPS which is why I decided to do some contributions, but I'll be honest, I'm not quite sure where I've some of the strings I haven't translated. Of course, if I do get enough time in the future, I'll fix the rest of the 'sections' (as in like "success" and "edit" and stuff). But by then, it would be a bit too big. Not to mention others would be able to fix the holes I wasn't able to fill. It would be great if someone else could help me with that.

andryyy:
Thank you! I love language contributions. :)

PR #3887, opened 2020-12-04
[WEB] Set domain to filtered one by default when creating new mailbox

balping:
This is a little convenience feature for admins: if one selects a domain in the dropdown to filter mailboxes, it is reasonable to assume that if they want to add a new mailbox, they want to add it to the selected domain.

mm


andryyy:
Cool, thank you! :)

PR #3891, opened 2020-12-06
[Config] Add Wikipedia link for TZ values

MAGICCC:
Add a Wikipedia link for TZ values because users tend to insert any random value for the TZ variable and thus have errors


andryyy:
Thanks. Perhaps we should add a check in the future to verify the selection. Until then, the provided link is fine, thanks!

PR #3923, opened 2021-01-06
Use tmpfs for watchdog's /tmp, reduce disk writes

ValdikSS:
Watchdog container writes data to /tmp quite frequently, which lead to unnecessary (excessive) disk writes, which is an issue for SSD.
Use tmpfs for /tmp inside the container to make it write to RAM.


ValdikSS:
@andryyy, please take a look at this pull request. It's very simple and doesn't break anything.

dragoangel:
+1 vote

andryyy:
Thanks

PR #3931, opened 2021-01-11
Add postscreen whitelist syntax and examples

ValdikSS:
In addition to documentation update, also add examples into whitelist file.


andryyy:
Thanks, it will be included with the next rebuild.

PR #3934, opened 2021-01-14
 [Rspamd] Sort & add infos for bad ASN map

MAGICCC:
- Sort ASNs in ascending order
 - Add comments for each ASN (AS name using https://ipinfo.io/)


andryyy:
Merci :)

PR #3946, opened 2021-01-20
Unsafe network allow on Postfix

fnzv:
Hello there!
I'm contacting you in order to report that after running the mailcow-dockerized setup and used it for a few days (just a few changes from the UI - add user/domain etc.) the server became an Open relay allowing anyone the send emails without authentication because of the transport allowing the entire Docker network 172.16.0.0/12 .

My system setup:
- Docker version 19.03.13, build 4484c46d9d
- Ubuntu 20.04.1 LTS

I think removing this network from the postfix networks is more safe and all the users by default should use authentication.
Allowing the entire docker network could give the opportunity also to other unsafe containers to reach this and exploit the NAT

Thank you for your time
Sami


andryyy:
No, that's when you misconfigured your network. mailcow is **not and never** an open relay in a proper configuration. If you masquerade all inbound connections, it is up to you to fix your network.

It is a **huge** error in your configuration when you masquerade the wrong connections. Removing the network from mynetworks does **not** fix that...

Whatever else runs on that machine or whatever you configured with iptables/netfilter, check it and check it again. Fix it.

andryyy:
Just to make it clear: Stop your installation and fix it. Please do not keep it running in that state. :/

fnzv:
@andryyy It was a test installation and i do not consider this tool safe if by design, it's up to you if you want people spawn dockerized open relays just keep it

And YES removing it from mynetworks does fix it, because the NATed connections hitting the VM with the ip 172.x.x.x will  be denied and can only send emails if they are authenticated

svengo:
192.168.0.0/16 is a private network (see RFC 1918) where the docker containers hang. IMHO it is okay to trust traffic from this network by default. If you run untrusted containers (not a good idea because root in the container is root on the host), you can change the setting.

andryyy:
The whole issue is about a dangerous misconfiguration.

The 1918 networks are part of the docs with a short information about how to remove them. Read the docs. :)

fnzv:
Wow what a great and original reply!  "Read the docs. :)" ... chill out dude you are just mantaining a bunch of scripts to run a postfix not the kernel of linux

I was just suggesting that it might be a good idea to opt out those networks by default to make it more secure and then if some one wants to add them it's just a few chars to add.

RFC 1918 networks are also used in cloud envs and natted (e.g. Openstack public clouds not using flat public network  and many major cloud providers) so this is why you should keep them out of mynetworks

ghost:
> chill out dude you are just mantaining a bunch of scripts to run a postfix not the kernel of linux

Jesus, what a prick you are.

patschi:
> Read the docs. :)

Docs don't write itself, so I appreciate how serious and thankful you are for available docs, what people help writing and updating them. Anyway: https://mailcow.github.io/mailcow-dockerized-docs/firststeps-rfc-1918/

> ... chill out dude you are just mantaining a bunch of scripts to run a postfix not the kernel of linux

Feel free to write your own "bunch of scripts".

This is just so disrespectful.

> I was just suggesting that it might be a good idea to opt out those networks by default [...]

Then simply remove those "few chars" on your setup. Internal networks should be secured by all means anyway.

> RFC 1918 networks are also used in cloud envs and natted

If you're running something like OpenStack public clouds and other customers can access customer's systems from someone else, you definitely have some other knobs you might want to adjust. Useable hosters and setups do separate by using other techniques, or at least disallow routing between internal networks from different users. So that's not a valid argument.

Anyway. We got your point, there are no plans so far changing it. And if you would like to change it on your setup, see linked docs above.

Locking this down.

PR #3956, opened 2021-01-30
[Api docs] Add sogo_visible property to alias endpoints

derJD:
`sogo_visible` is missing in  `{add,get}/alias` endpoint documentation.


andryyy:
Merci :)

PR #3967, opened 2021-02-06
[Dovecot] Add Russian and Ukrainian folders

dragoangel:


andryyy:
Merci. :) It is tested, I guess? Feel free to merge, if it is.

ValdikSS:
Should alternative variants with `ё` for удаленные and удаленные элементы also be listed?

dragoangel:
I don't saw that any software try use such folders, so no, or I missed something?:)

PR #3968, opened 2021-02-06
[docs] /api/v1/edit/mailbox Changed type of items parameter to list

myOmikron:
As the Swagger UI displays the items object in the request body as a String instead of a list, requests made with this method would lead into empty responses.


ntimo:
Thank you :)

PR #3974, opened 2021-02-10
[Web] Implement all supported dovecot password schemas

kaechele:
When migrating from other Dovecot based installations it can be very
convenient to just copy over existing hashed passwords.
However, mailcow currently only supports a limited number of password
schemes.

This commit implements all password schemes that do not require
challenge/response or OTP mechanisms.

A convenient way to generate the regex with all supported schemas is
`docker-compose exec dovecot-mailcow doveadm pw -l | awk -F' ' '{printf
"/^{("; for(i=1;i<=NF-1;i++){printf "%s%s", sep, $i; sep="|"}; printf
")}/i\n"}'`

Note that this will also include unsupported challenge/response and OTP
schemas.

Furthermore this increases the vsz_limit for the dovecot auth service to
2G for the use of ARGON2I and ARGON2ID schemas.

Signed-off-by: Felix Kaechele <felix@kaechele.ca>

### DIFF
--- a/data/conf/dovecot/dovecot.conf
+++ b/data/conf/dovecot/dovecot.conf
@@ -382,6 +382,7 @@ service auth {
     mode = 0600
     user = vmail
   }
+  vsz_limit = 2G
 }
 service managesieve-login {
   inet_listener sieve {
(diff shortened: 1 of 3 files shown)


andryyy:
Amazing! Thank you very much!

andryyy:
Updated https://github.com/mailcow/mailcow-dockerized-docs/blob/master/docs/model-passwd.md accordingly.

PR #3975, opened 2021-02-11
move white/blacklist form above the table

tobikris:
I would like to propose a UI change for the whitelist and blacklist forms. I had to add many entries to the blacklist and it is a little unfortunate that the form field is below the table. I always need to scroll back down to add the next one.
As it is very easy from my point of view to make the usability better in this aspect here is my PR.

It is simply moving the forms above the table.

mailcow-blacklist-form-reorder


andryyy:
That's fine with me. :) Can you do that in edit.php, too?

tobikris:
Good idea - did not think of that. Thanks!

andryyy:
Thank you. :)

PR #3980, opened 2021-02-13
minor ejabberd changes

FriedrichAltheide:
- add 5223 and 5270 ports
- set c2s_protocol_options/s2s_protocol_options
- set c2s_ciphers/s2s_ciphers
- improved omemo support
- s2s_use_starttls required


andryyy:
You can't set the port like this without using a macro. Setting the port to basically anything else will not change it in ejabberd. Furthermore I like to keep it as it is for now, while accepting the change in s2s tls required and the OMEMO stuff if you want to explain it to us a bit. :)

Please also don't make major changes in a PR without A) explanation and B) without discussing it. :(

FriedrichAltheide:
XMPP Ports:
- 5222 will use startTLS while 5223 will only allow TLS
- 5223 will therefore improve the user experience on mobile clients as less RTT will be required
- 5223 will require an additional SRV record _xmpps-client
 
OMEMO: not all clients set the access permissions for the omemo bundle and devicelist correct. With the patch ejabberd ensures that all clients can access these entries (cmp. https://www.freie-messenger.de/dateien/xmpp/perfect-xmpp-setup.PDF, slide “OMEMO aktivieren”).
 
Sorry where would you like to see a macro?

PR #3981, opened 2021-02-13
Unset Postfix smtpd_tls_session_cache_database, reduce disk writes

ValdikSS:
Postfix may update smtpd_tls_session_cache_database quite frequently even on not busy server, which leads to unnecessary (excessive) disk writes, which is an issue for SSD.
Postfix documentation suggests not to use this parameter anymore since there's another, better TLS session resumption method available.

>As of Postfix 2.11 the preferred mechanism for session resumption is RFC 5077 TLS session tickets, which don't require server-side storage. Consequently, for Postfix ≥ 2.11 this parameter should generally be left empty.

http://www.postfix.org/postconf.5.html#smtpd_tls_session_cache_database


andryyy:
Thank you, good catch. I didn't know.

PR #3987, opened 2021-02-16
[Rspamd] Increase map watch interval

### LINKED ISSUE #3929: HTTPS requests to invaluement.com are performed every 5-6 seconds on idle server
**Prior to placing the issue, please check following:** *(fill out each checkbox with an `X` once done)*
- [X] I understand that not following or deleting the below instructions will result in immediate closure and/or deletion of my issue.
- [X] I have understood that this bug report is dedicated for bugs, and not for support-related inquiries.
- [X] I have understood that answers are voluntary and community-driven, and not commercial support.
- [X] I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

## Summary
Mailcow performs HTTPS queries to `www.invaluement.com` domain (to `https://www.invaluement.com/spdata/sendgrid-id-dnsbl.txt` and `https://www.invaluement.com/spdata/sendgrid-envelopefromdomain-dnsbl.txt` URLs) every 5-6 seconds.
Full TCP connection is established and closed for this query (from SYN to FIN), this is not a keep-alive ping.
This creates unnecessary load to `invaluement.com` server.
The queries are performed by [ivm-sg.lua](https://github.com/mailcow/mailcow-dockerized/blob/master/data/conf/rspamd/lua/ivm-sg.lua) script.

## Logs
I found nothing in logs regarding these requests.
Tcpdump log with timestamps is attached.

[**tcpdump-log-invaluement-com.txt**](https://github.com/mailcow/mailcow-dockerized/files/5793254/tcpdump-log-invaluement-com.txt)

## Reproduction
1. Run tcpdump on `www.invaluement.com` IPv4 and IPv6 addresses. Command for current address set: `tcpdump host 104.22.15.144 or host 172.67.14.207 or host 104.22.14.144 or host 2606:4700:10::6816:f90 or host 2606:4700:10::6816:e90 or host 2606:4700:10::ac43:ecf`
2. Observe new TCP connections every 5-6 seconds.

I've tried to intercept the data by replacing the URL to my HTTP mocking server which just returned HTTP 200 OK, but in this case there are no repetitive requests are performed.

## System information

| Question | Answer |
| --- | --- |
| My operating system | Linux Ubuntu 20.04 |
| Is Apparmor, SELinux or similar active? | Yes, AppArmor. No issues with it in audit logs. |
| Virtualization technlogy (KVM, VMware, Xen, etc - **LXC and OpenVZ are not supported** | Bare metal |
| Server/VM specifications (Memory, CPU Cores) | 4 cores, 16 GB RAM |
| Docker Version (`docker version`) | 20.10.1 |
| Docker-Compose Version (`docker-compose version`) | 1.27.4, build 40524192 |
| Reverse proxy (custom solution) | Custom configuration, did not touch Mailcow configs, irrelevant |

- Output of `git diff origin/master`, any other changes to the code? **No**.
- All third-party firewalls and custom iptables rules are unsupported. *Please check the Docker docs about how to use Docker with your own ruleset*. Nevertheless, iptabels output can help us to help you: `iptables -L -vn`, `ip6tables -L -vn`, `iptables -L -vn -t nat` and `ip6tables -L -vn -t nat`.
- DNS problems? Please run `docker exec -it $(docker ps -qf name=acme-mailcow) dig +short stackoverflow.com @172.22.1.254` (set the IP accordingly, if you changed the internal mailcow network) and post the output.

MAGICCC:
May fixes https://github.com/mailcow/mailcow-dockerized/issues/3929

We can also set it to 45s because since the actual check interval is jittered, it'll get executed every 1x to 2x of the set value (45s to 90s).
https://rspamd.com/doc/configuration/options.html#global-options

### DIFF
--- a/data/conf/rspamd/local.d/options.inc
+++ b/data/conf/rspamd/local.d/options.inc
@@ -1,7 +1,7 @@
 dns {
   enable_dnssec = true;
 }
-map_watch_interval = 30s;
+map_watch_interval = 1min;
 dns {
   timeout = 4s;
   retransmits = 2;


andryyy:
This does also affect when settings are applied (blacklist and whitelist entries made via UI). Not sure if that's good. Perhaps we better remove the map we seem to hammer.

MAGICCC:
Also I didn't found a setting where you can change intervals for different maps in a first glance. Or maybe it's possible with a different approach

andryyy:
Hmmm, perhaps we should disable that specific map instead of increasing the refresh interval.

PR #3996, opened 2021-02-21
Fix for listing mailboxes and aliases from multiple domains

gjerek:
Fix for API call to list multiple mailboxes and aliases from multiple domains.

### DIFF
--- a/data/web/json_api.php
+++ b/data/web/json_api.php
@@ -845,7 +845,7 @@ if (isset($_GET['query'])) {
                   $domains = mailbox('get', 'domains');
                 }
                 else {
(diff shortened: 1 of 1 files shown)


andryyy:
Why exactly "fix"?

I don't think we allow for a comma in the URL. It's also missing the change in the API docs. 

I think there should be at least (!) a discussion before proposing changes to the API. :)

andryyy:
We will do a better implementation with an APIv2 :) We chatted in Telegram about that.

andryyy:
@ntimo Can you update the API? :)

PR #4010, opened 2021-03-03
[PHP-FPM] Increase PHP memory limit for "cli" to 512M

teissler:
The CLI of Nextcloud (occ) complains about the memory limit which is by default 256 MB.
So please consider changing the php cli memory limit to 512 MB like for the PHP FPM web-worker pool.

### DIFF
--- a/data/conf/phpfpm/php-conf.d/other.ini
+++ b/data/conf/phpfpm/php-conf.d/other.ini
@@ -1,2 +1,3 @@
 max_execution_time = 3600
 max_input_time = 3600
+memory_limit = 512M


andryyy:
But you just changed FPM config.

teissler:
No, the file `data/conf/phpfpm/php-conf.d/other.ini` is mapped to `/usr/local/etc/php/conf.d/zzz-other.ini` in php-fpm-mailcow container and thus loaded by the php cli.

```
root@mail01:/opt/mailcow-dockerized# cat data/conf/phpfpm/php-conf.d/other.ini 
max_execution_time = 3600
max_input_time = 3600
memory_limit = 512M
root@mail01:/opt/mailcow-dockerized# docker exec -it -u www-data $(docker ps -f name=php-fpm-mailcow -q) bash -c "php -i | grep -i memory_limit"
memory_limit => 512M => 512M
```

andryyy:
Cool. :) I think we could then remove the mem limit from the pool config. Right?

andryyy:
Hm, I forgot we have different values for the workers. That's fine then, I guess.

Thank you for your contribution!

PR #4012, opened 2021-03-03
Added missing french parts (XMPP and Nginx service)

guiguir68:
Comparing the lang_en.json file, I added the translations regarding the XMPP and Nginx service.


MAGICCC:
Merci!

Issue #2507, opened 2019-04-07
Regex in aliases


Athan1:
Like in issue #1787. I would appreciate if add support for regex in alias addresses. Why is it needed? I'm using used postfix regex-maps for several years by now and switching to mailcow would be much easier, when I simply could use the same entries. What I always liked about the regex-aliases is, they are very flexible. e.g. special e-Mail addresses for every online login. Of course there is the recipient_delimiter, but every body knows that and if I use foo+bar@example.com everybody knows I'm foo@example.com. Okay I can switch the delimiter, but that's not a big bargain. With regex you can build nearly every pattern. Who to do it? I hope it shouldn't be a big deal to implement. Maybe replacing the normal query in /opt/postfix/conf/sql/mysql_virtual_alias_maps.cf (in the postfix container) by
 ```
SELECT goto FROM alias
  WHERE address='%s' REGEXP CONCAT('^',alias.address,'$') AND SUBSTRING(alias.address,1,1) != '@'     AND active='1';
```
 and allow entering REGEXP as aliases in the UI might do most of the job.

**Comment by timm0e** [comment]
+1

**Comment by timo-reymann** [comment]
+1

**Comment by axello** [comment]
+1 (also coming from self-hosted postfix server where I used the regex. Now I have to edit the regex file in the postfix docker container, which is sub-optimal.

**Comment by BorysVrublevskyi** [comment]
For everyone who was looking for a workaround with some limitations. This will allow you to match incoming mail by regex and redirect it to any other mail but **will break the Catch-ALL rule in the Aliases**.

(Assume that you are `root` and repo was cloned into the `/root/mailcow-dockerized`)

Override main.cf

```bash
nano /root/mailcow-dockerized/data/conf/postfix/extra.cf
```

Add new line

```
virtual_alias_maps = regexp:/opt/postfix/conf/aliases-regexp
```

Add regex rules

```bash
nano /root/mailcow-dockerized/data/conf/postfix/aliases-regexp
```

```
/^prefix1(.*)@mydomain.com$/    catch@mydomain.com
/^prefix2(.*)@mydomain.com$/  catch@mydomain.com
```

Restart postfix container

PR #4015, opened 2021-03-05
Allow regex in email address aliases

chaserdoe:
Implemented #2507


dragoangel:
Hm, aliases in mailcow not only allow incoming mail, but also and outgoing. Does this work here? Also regex aliases not break alias expanding in rspamd?

chaserdoe:
Do you mean address rewriting? I only implemented regex for incoming mail aliases. What exactly do you mean by alias expanding? The global filter maps seemed to work fine.

dragoangel:
1. Any user that has alias allowed to use this address in from header (identity).
2. Rspamd per user settings such as tagged mail or whitelist on user only applies on user itself, but mailcow has script to expand alias to original user to apply this settings correctly in case with alias also.
Does your PR not break this futures?

chaserdoe:
1. Works now, but sogo prevents from headers, which are unknown to it. Works fine in other mail clients though.
2. Seems to work fine.

dragoangel:
Hi @andryyy what you think?:)

mbretter:
any chance getting this merged?

Levivb:
this looks like a very decent feature to support. any chance of getting this merged? I might need it to create a `<some-prefix>.<whatever-here>@domain.com`

for example`mb.*@domain.com` (or maybe `mb\.[\w-]+@domain.com` regex style) which would allow `mb.facebook@domain.com`

gnasch:
If I understand this correctly this would allow to bind aliases to the services they are used for. This is great because they will refrain from selling traceable addresses, and they can selectively be revoked.

patschi:
That's indeed a neat feature! 
@chaserdoe would you please so kind solving the conflicts on your PR?

I'd like to wait for a quick check and OK from @andryyy before merging it.

retoo:
This is awesome and absolute critical for my migration path. I'm looking forward to try this!

apharp:
> 
> 
> This is awesome and absolute critical for my migration path. I'm looking forward to try this!

Me too! Really looking forward to this feature!

Ulrar:
I was just writing a feature request for this, great to see someone already did the work !
Also would be very happy to this see merged, that's what I'm missing to migrate.

retoo:
@chaserdoe FYI there was some refactoring in upstream

As far as I can tell these changes were done:
- the form field got refactored from edit.php to alias.twig
- it now uses twig notation (instead of pure php) 

I'm not very fluent in php, maybe you have a chance to merge in the latest changes. Otherwise I will give a try sooner or later.

Beside that: the regexp alias feature has worked very well in the last two months at my installation.

retoo:
I've just created an offical ticket asking for the merge of this PR. I offered buying another SAL, maybe some of the watchers are also interested and could chime in. Merging this feature probably is just a few hours of work (plus maintaining it of course). 

Cheers
Reto

andryyy:
This will have so many bad effects on routing, policies and spam filter settings that I’m against merging it. I will not veto it, feel free, I will not support it though.

retoo:
@andryyy I understand that, and I had some corner cases myself in the past 9 months, but on the other hand it's a feature I which is extremely useful I cannot give up, I have 15 years of 'reto-company-.*' email addresses ;). So maybe it just needs a very fat disclaimer with all the corner cases that are not supported (i.e. sending mails from SOGo).

apharp:
Not sure what the status of this is, or if it will be supported but I found a quick, dirty, unsupported workaround that's been working for me for the last few months.

## Step 1
Create regex virtual map file for postfix

`nano data/conf/postfix/regex.cf`

This is where you will create your regex rules.

```
/^someprefix\.(.*)\@(yourdomain\.com)/ recipient@domain.com
/^(reg|regex)\.(.*)\@(yourdomain\.com)/ recipient1@domain.com,recipient2@domain.com
```

## Step 2

Next configure the extra.cf to look for this file. NOTE: since you are replacing *virtual_alias_maps* from the main config (main.cf) you MUST include the original alias maps + your new alias map.

`nano data/conf/postfix/extra.cf`

Add this line

```
virtual_alias_maps = proxy:mysql:/opt/postfix/conf/sql/mysql_virtual_alias_maps.cf,
  proxy:mysql:/opt/postfix/conf/sql/mysql_virtual_resource_maps.cf,
  proxy:mysql:/opt/postfix/conf/sql/mysql_virtual_spamalias_maps.cf,
  proxy:mysql:/opt/postfix/conf/sql/mysql_virtual_alias_domain_maps.cf,
  regexp:/opt/postfix/conf/regex.cf
```

## Step 3

Run postmap

```
docker-compose exec postfix-mailcow postmap /opt/postfix/conf/regex.cf
```

## Step 4

Reload postfix

`docker-compose restart postfix-mailcow`

aHcVolle:
> Not sure what the status of this is, or if it will be supported but I found a quick, dirty, unsupported workaround that's been working for me for the last few months.
> ## Step 1
> ...

I Just tried it and this is working fine. 
Thanks for this tip :)

RafaelKr:
Thank you for this workaround!

BTW my use case would be to have an alias for `dmarc@*` which forwards this for every domain to `dmarc@example.com`. `dmarc@example.com` is the inbox which is scanned by https://github.com/patschi/parsedmarc-dockerized and currently I have to create an alias for every domain manually.

axello:
+1 (also coming from self-hosted postfix server where I used the regex. Now I have to edit the regex file in the postfix docker container, which is sub-optimal. (but doable, as the regexes do not change often.)
Thanks for the detailed guide above!

IGLOU-EU:
I've taken a keen interest in this PR and after reading through the comments, I understand some of the concerns that are raised.

Instead of using full regex, we could limit ourselves to wildcard support. We can implement this by using the PHP glob() function, which is simpler and less risky compared to a full regex implementation.

@andryyy Before proceeding, I'd appreciate your thoughts on this. Could this approach address some of your concerns? Could you elaborate a bit more on the potential problems you foresee with this use case so that I can create some test scenarios?

--- 

EDIT:

After some discussions with the mailcow team on the telegram chan. This feature is not expected and never be merged.
But Algoo company need a simple pattern matching like wildcard *-me@you.us, so we have made a fork.
Any information in our implementations here [\#5](https://github.com/algoo/galae/pull/5)

retoo:
@andryyy hey could you imagine working/accepting on solution that works like the "workaround" described by andy ([apharp](https://github.com/apharp))  in https://github.com/mailcow/mailcow-dockerized/pull/4015#issuecomment-1216045911

I.e. no gui support, and just accepting a new file at a certain path (which is empty by default installation). the difference to the workaround would just be, that we don't have to overwrite the built-in 'virtual_alias_maps'

What do you think?

PR #4020, opened 2021-03-09
[Rspamd] Sign Disposition-Notification Headers

dragoangel:
Add more Headers to DKIM signing


andryyy:
Is this a good idea for every added header?
Some may require further discussion. Perhaps you should propose this upstream. :)

dragoangel:
We  can discuss this in telegram :) but I don't think it bad to list sensitive headers that must be signed.

andryyy:
Signing the wrong headers can be problematic. A patch to upstream would be much better.

PR #4024, opened 2021-03-10
Refactor support for pre-hashed passwords

SnailShea:
Currently working on a migration from another mail service and wanted to carry over the existing hashes for our domain administrators. This code consolidates the check for an existing hash to the hashing function, rather than repeating the check throughout the code to achieve this functionality.


andryyy:
Thank you! :)

PR #4027, opened 2021-03-13
[Update, Config] Add subject for watchdog emails

the-voidl:
Adding option to customize the subject of mails that the watchdog sends.

This is useful if you maintain multiple instances of cows (a so called cow herd) or are just terryfied by the phrase "Watchdog: ALERT".


andryyy:
Merci, looks good.

Issue #1007, opened 2018-02-06
Use sdbox/mdbox


Lennix:
I was wondering if it would be possible to use sdbox or mdbox instead of maildir? I've already fiddled around with the config but it didn't seem to work.

The idea is that I'd like to use mail_attachment_dir to put larger attachments onto an external storage, but that only works for dbox. Also there's a nice performance boost when using dbox (on larger installations).

extremeshok:
removed

extremeshok:
removed

tgmedia-nz:
I second this - I have used mdbox on large scale systems, maildir would have killed it with IO. Also, there is no coding required for moving mail to another storage layer, dovecot has a built in "alternative storage" with dbox/mdbox: https://wiki2.dovecot.org/MailboxFormat/dbox.

All my deploys of mailcow I have updated maildir to mdbox. For compression I use ZFS :)

@Lennix - you'll have to update the config in

mailcow-dockerized/data/Dockerfiles/dovecot/docker-entrypoint.sh:

`SELECT CONCAT('mdbox:/var/vmail/',maildir)  ....`

(change the first maildir to mdbox)

and add

`mail_location = mdbox:~/`

to dovecot.conf

and rebuild your image (https://mailcow.github.io/mailcow-dockerized-docs/u_e-docker-cust_dockerfiles/)

Another way is to copy your docker-entrypoint.sh to the conf dir and update it, then link it into the image via docker-compose.yml:

```
      volumes:
        - ./data/conf/dovecot/docker-entrypoint.sh:/docker-entrypoint.sh:ro

```
I know it's hacky but works for me.

stevesbrain:
@tgmedia-nz Out of curiosity, what sort of size were you at to notice the difference in IO?

extremeshok:
Ok, ive been doing allot of research and testing regarding DBOX and have completely changed my opinion with it.

The biggest benefit is with the following: **MDBOX + SIS + LZ4**

SBOX reduced the load on needing to scan an entire directory on loading an inbox. There is also a vastly reduced IO as the dbox index files contain the message flags, so if a user marks 50emails as viewed, only 1 files is updated. (maildir, each file would be renamed).

[...]

PR #4028, opened 2021-03-13
Enable maildir_very_dirty_syncs by default

Daniel15:
`maildir_very_dirty_syncs` can significantly improve performance of very large mailboxes. 

From Roundcube's IMAP logs when loading one email from a mailbox containing 250k emails:
Before: 1.719s to fetch:
```
A0008 OK Fetch completed (0.860 + 0.000 + 0.859 secs).
```

After: 0.025s to fetch:
```
A0008 OK Fetch completed (0.013 + 0.000 + 0.012 secs).
```

What this option does is it avoids rescanning the entire `cur` directory whenever loading an email. With this option disabled, Dovecot takes it safe and scans the ENTIRE `cur` directory (essentially running an `ls`) to check if that particular email was touched (renamed, etc), by looking for all files whose names contain the correct ID. This is very slow if the directory is large, even on filesystems optimized for such use cases (eg. ext4 with dir_index enabled). 

With this option enabled, Dovecot will still notice changes if the file's mtime is changed, but otherwise it will not scan the directory and just assumes the index is up-to-date. This is essentially the same as what sdbox/mdbox do, and with this option you can get some of the performance increase that would come with sdbox/mdbox (see #1007) while still using maildir.

This is safe to use as long as you do not manually touch files under `cur` (as then Dovecot may not notice the changes). Delivering new emails using LMTP (which Mailcow does) is fine, and dropping new emails into the `new` directory is also fine, it's just modifying stuff under `cur` that's a bit risky. If you do manually modify files under `cur`, you can just delete the index files and let Dovecot rebuild them.

This option seems to be safe to enable with Mailcow, as nothing appears to directly modify files in the `cur` directory other than Dovecot (Docker is very useful to verify this - the volume is not mounted in any other Docker containers). Some other systems enable it default - For example, cPanel has it enabled by default. I did not want to enable it by default in Mailcow because we have no way of telling what weird things people are doing with their maildirs :)

### DIFF
--- a/data/conf/dovecot/dovecot.conf
+++ b/data/conf/dovecot/dovecot.conf
@@ -27,7 +27,7 @@ mail_attachment_min_size = 128k
 # Significantly speeds up very large mailboxes, but is only safe to enable if
 # you do not manually modify the files in the `cur` directories in
 # mailcowdockerized_vmail-vol-1.
-# https://mailcow.github.io/mailcow-dockerized-docs/u_e-dovecot-perf/
+# https://docs.mailcow.email/manual-guides/Dovecot/u_e-dovecot-performance/
 maildir_very_dirty_syncs = yes
 
 # Dovecot 2.2
--- a/data/conf/dovecot/dovecot.conf
+++ b/data/conf/dovecot/dovecot.conf
@@ -28,7 +28,7 @@ mail_attachment_min_size = 128k
 # you do not manually modify the files in the `cur` directories in
 # mailcowdockerized_vmail-vol-1.
 # https://mailcow.github.io/mailcow-dockerized-docs/u_e-dovecot-perf/
-maildir_very_dirty_syncs=yes
+maildir_very_dirty_syncs = yes
 
 # Dovecot 2.2
 #ssl_protocols = !SSLv3
--- a/data/conf/dovecot/dovecot.conf
+++ b/data/conf/dovecot/dovecot.conf
@@ -25,9 +25,10 @@ mail_attachment_fs = crypt:set_prefix=mail_crypt_global:posix:
 mail_attachment_dir = /var/attachments
 mail_attachment_min_size = 128k
 # Significantly speeds up very large mailboxes, but is only safe to enable if
-# you do not manually modify the files in the `cur` directories in 
-# mailcowdockerized_vmail-vol-1. Disabled by default.
-#maildir_very_dirty_syncs=yes
+# you do not manually modify the files in the `cur` directories in
+# mailcowdockerized_vmail-vol-1.
+# https://mailcow.github.io/mailcow-dockerized-docs/u_e-dovecot-perf/
+maildir_very_dirty_syncs=yes
 
 # Dovecot 2.2
 #ssl_protocols = !SSLv3
--- a/data/conf/dovecot/dovecot.conf
+++ b/data/conf/dovecot/dovecot.conf
@@ -24,6 +24,10 @@ mail_plugins = </etc/dovecot/mail_plugins
 mail_attachment_fs = crypt:set_prefix=mail_crypt_global:posix:
 mail_attachment_dir = /var/attachments
 mail_attachment_min_size = 128k
+# Significantly speeds up very large mailboxes, but is only safe to enable if
+# you do not manually modify the files in the `cur` directories in 
+# mailcowdockerized_vmail-vol-1. Disabled by default.
+#maildir_very_dirty_syncs=yes
 
 # Dovecot 2.2
 #ssl_protocols = !SSLv3


andryyy:
I thought about adding this as default a while ago. What do you guys think?

Daniel15:
@andryyy  I'd be fine with it being the default. cPanel has it enabled by default, meaning tens of thousands of mail servers (if not more) use this config option without problems, so the risk should be fairly minimal.

A standard Mailcow installation would have no problems with this setting being enabled - It's only if people manually modify the maildir files (which is pretty rare) that there'd be any problems. This could just be documented in the Mailcow docs though.

andryyy:
Yes, should be in the docs indeed. :)

andryyy:
Still testing it, it feels good.

Can you create a PR for the docs? Only if you like and find time. If you don't, I will do it.

Daniel15:
Sure, I'll send a PR when I get a chance.

Daniel15:
I updated this PR to actually enable the option rather than just adding a comment, and created a PR for the docs site (https://github.com/mailcow/mailcow-dockerized-docs/pull/275) to add some docs :)

sschwetz:
The only reason that I could see to not enable this by default would be that I it doesn't place nice with NFS. But I don't see this as being a major issue as many other portions of mailcow (mariadb) do not work at all when stored on NFS.

Daniel15:
> The only reason that I could see to not enable this by default would be that I it doesn't place nice with NFS.

@sschwetz  It's actually fine over NFS as long as you follow the rule that nothing other than one Dovecot should touch the maildir files. In particular, running multiple Dovecot instances that use the same maildir, like you might with a high-availability configuration, may not work properly. I wouldn't expect most Mailcow users to want to do that though. 

The person that implemented `maildir_very_dirty_syncs` says that it actually works fine if the maildir is on NFS and shared with other Dovecot instances as long as the index is also on NFS and shared the same way (https://dovecot.org/list/dovecot/2009-March/037944.html), but I'm not sure how well that works in terms of locking.

Daniel15:
@andryyy Could you please consider merging this? The bot closed it :(

patschi:
I like the change and I think it might be beneficial performance-wise indeed. However wondering, if dovecot doesn't scan the whole `cur/` folder each time, is there a way forcing dovecot to rescan if manual changes were done? (I guess restarting dovecot should do the trick?) There might be users doing changes in the folder, so we should at least provide a comment how to force a rescan if required.

@andryyy You've been testing this one since quite a while. Assuming you haven't seen any issues, I think it might be ready to merge.

However we might want to note somewhere that manually manipulating files in `cur/` isn't something we recommend/support?

Daniel15:
> if dovecot doesn't scan the whole cur/ folder each time, is there a way forcing dovecot to rescan if manual changes were done?

You need to shut down Dovecot, delete the relevant indexes for the folder that was modified (in `vmail-index-vol-1`), and then start Dovecot again. Deleting the index is a safe operation as Dovecot will rebuild the indexes on the fly as needed. The maildir is the source of truth for the mail; the index is just so Dovecot has fast access to some data like email subjects and received dates.

Note that this is similar to Dovecot's "mbox" format (it has exactly the same limitation around manually modifying the backing files), however deleting the index is NOT SAFE in that case as some data is ONLY stored in the index when using mbox. This is not the case when using maildir. Mailcow currently only supports maildir.

Having said that, I don't think people would be manually modifying the maildir files for a Mailcow instance very often, given they're encrypted so you can't just open them in a regular text editor and modify them.

patschi:
Makes sense. IMHO it might be good to add a quick side-note to the DocPR of yours, just for completeness.

I'm adding the option on my installations, let's see how it works out.

Daniel15:
Sure, I'll do that when I get a chance.

For what it's worth, I've been using this option on my mailserver since the start of the year and haven't hit any issues with it 😄

patschi:
Sure, take your time.

Just one more thing, to be picky, changing:
```
maildir_very_dirty_syncs=yes
```
to
```
maildir_very_dirty_syncs = yes
```
(with spaces, just like the other settings)

Ry3nlNaToR:
Hi, Just wondering is there any plans to merging this, enabling maildir_very_dirty_syncs shouldn't break anyone server, the only case where this would be issue if someone was doing manual edits in the maildir for 99.9% of users that would very unlikely since the files in maildir are lz4 compressed and encrypted.

Enabling it significantly speeds up imap fetch especially helpful for users with large inboxes.

I've been using it on my server for nearly a year zero issues with it disable using Roundcube for one of my large inbox over 200k email it's slow at loading inbox takes over 3-4 secs with it enable near zero load time in Roundcube within a sec.

ricardoamaro:
Plus one for getting this merged, please! :)

DerLinkman:
Will look into it.

dasn4pp3l:
+1 for merge

Turab:
Would this improve things with SoGO too? Or we have to use only Roundcube?

ghost:
> Would this improve things with SoGO too? Or we have to use only Roundcube?

Any email client would benefit from this.

data/conf/dovecot/dovecot.conf
```diff
@@ -24,6 +24,11 @@ mail_plugins = </etc/dovecot/mail_plugins
 mail_attachment_fs = crypt:set_prefix=mail_crypt_global:posix:
 mail_attachment_dir = /var/attachments
 mail_attachment_min_size = 128k
+# Significantly speeds up very large mailboxes, but is only safe to enable if
+# you do not manually modify the files in the `cur` directories in
+# mailcowdockerized_vmail-vol-1.
+# https://mailcow.github.io/mailcow-dockerized-docs/u_e-dovecot-perf/
```
DerLinkman: Please change the docs link to: https://mailcow.github.io/mailcow-dockerized-docs//manual-guides/Dovecot/u_e-dovecot-performance/

DerLinkman (changes requested):
Only that small change, the rest looks good!

Many thanks

data/conf/dovecot/dovecot.conf
MAGICCC: You mean https://docs.mailcow.email/manual-guides/Dovecot/u_e-dovecot-performance/ for the domain?

DerLinkman: Ah yes sorry....

DerLinkman (approved):
Looks very good!

I'm doing some final tests and will merge it then.

PR #4040, opened 2021-03-26
[netfilter] Use exit code 2 if an error occurs

monsterry:
At the moment the python server always exits with exitcode `0`. This is problematic if you change the container restart policy from `always` to `on-failure`. This PR changes `server.py` to exit with exitcode `2` if an error occured.


andryyy:
Looks good to me.

PR #4043, opened 2021-03-31
Removed --database-table-prefix switch

fabreg:
Looks like in NC 2.0 this option is no more supported. During installation I got this error:

The "--database-table-prefix" option does not exist.

### DIFF
--- a/helper-scripts/nextcloud.sh
+++ b/helper-scripts/nextcloud.sh
@@ -110,7 +110,6 @@ elif [[ ${NC_INSTALL} == "y" ]]; then
     --database-name ${DBNAME} \
     --database-user ${DBUSER} \
     --database-pass ${DBPASS} \
-    --database-table-prefix nc_ \
     --admin-user admin \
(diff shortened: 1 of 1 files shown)


andryyy:
Is there an alternative?

fabreg:
No @andryyy 
https://docs.nextcloud.com/server/20/admin_manual/installation/command_line_installation.html
As per documentation they mention only 

--database \
"mysql" --database-name "nextcloud"  --database-user "root" --database-pass \
"password" --admin-user "admin" --admin-pass "password"

andryyy:
Hm. :( It's okay, but messy for us. We may think about a new database for that. I will merge and create a PR later to create a new db for it, ok?

fabreg:
We have two options: asking to upstream to add again that option for >2.x versions or, as you proposed, create a new db.
Second way is faster.

realizelol:
This could be an alternative:

https://help.nextcloud.com/t/change-the-dbtableprefix-on-a-running-docker-instance/81662/2

best regards
realizelol

PR #4046, opened 2021-04-02
Respect Redis REPLICA

Der-Jan:
For replicated REDIS setup, servers_read&_write should be used ( which is already setup in local.d/redis.conf )

### DIFF
--- a/data/conf/rspamd/local.d/reputation.conf
+++ b/data/conf/rspamd/local.d/reputation.conf
@@ -2,9 +2,7 @@ rules {
   ip_reputation = {
     selector "ip" {
     }
-    backend "redis" {
-      servers = "redis";
-    }
+    backend "redis";
     symbol = "IP_REPUTATION";
   }
 }


patschi:
I don't really understand this change, or what this is supposed to address specifically? According to the [rspamd documentation for reputation](https://rspamd.com/doc/modules/reputation.html) the correct way to specify redis as the backend is what we're using as of today:
```
    backend "redis" {
      servers = "redis";
    }
```
Which points to the redis server shipped with the mailcow docker stack.

Der-Jan:
When you make use of REDIS_SLAVEOF_IP & PORT, redis server within the stack becomes read only replica (s. https://github.com/mailcow/mailcow-dockerized/blob/2a2857ea24b2f65f1b5c9e2ac7ceac49387f575e/data/Dockerfiles/rspamd/docker-entrypoint.sh#L55) . Therefore specifying servers = "redis" - will overwrite the read & write servers from local.d

andryyy:
Thanks, that's a good catch.

PR #4047, opened 2021-04-04
[API docs] Fix password property spelling in add,edit/mailbox endpoint

derJD:
This MR replaces `pasword` with `password`


andryyy:
Thanks

PR #4068, opened 2021-04-20
[ClamAV] Update to 0.103.2

svengo:
Security patch, see https://blog.clamav.net/2021/04/clamav-01032-security-patch-release.html


andryyy:
Thanks :)

PR #4073, opened 2021-04-26
Add server_tokens off to the default site settings

maxileith:
Hello,

I think it makes sense to hide the NGINX version in the default settings for sites. For example, if someone configures a redirect from `webmail.example.com` to `mail.example.com/SOGo/` as a new site, they probably won't think of setting `server_tokens off`. So the version can be leaked easily.

### DIFF
--- a/data/conf/nginx/includes/site-defaults.conf
+++ b/data/conf/nginx/includes/site-defaults.conf
@@ -3,6 +3,8 @@
   charset utf-8;
   override_charset on;
 
+  server_tokens off;
+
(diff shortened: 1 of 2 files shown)


andryyy:
It is off by default in mailcow.

andryyy:
The "webmail" site in mailcow does not include the defaults. So this would not help.

Including it in the defaults _does_ help with SNI enabled configs I think. So we can merge it.

I will add the server_token stuff to the docs.

andryyy:
Done. :) Thank you!

PR #4082, opened 2021-05-04
ACME_CONTACT: Fix broken quotation mark in update.sh

the-voidl:
The escape sequence here was broken so that update.sh will always fail when trying to insert ACME_CONTACT.

https://t.me/mailcow/148987


the-voidl:
Closed due obsolet

Issue #4087, opened 2021-05-06
XMPP writes users password to log in plaintext in logfile


ghost:
**Prior to placing the issue, please check following:** *(fill out each checkbox with an `X` once done)*
- [x] I understand that not following or deleting the below instructions will result in immediate closure and/or deletion of my issue.
- [x] I have understood that this bug report is dedicated for bugs, and not for support-related inquiries.
- [x] I have understood that answers are voluntary and community-driven, and not commercial support.
- [x] I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

## Summary
When users authenticate to the XMPP contianer the [custom authtication script](https://github.com/mailcow/mailcow-dockerized/blob/master/data/Dockerfiles/ejabberd/authenticator) does write the users password in plaintext to a logfile on the container, the password is written to this location: /var/www/authentication/auth.log on the xmpp container. When I do a cut on that file I do see all users password in plaintext.

Line of intrest in /data/Dockerfiles/ejabberd/authenticator:
`$stdoutHandler = new Monolog\Handler\StreamHandler('/var/www/authentication/auth.log', Logger::DEBUG);`

I really don't like to have xmpp auth script log my users password in plain text like that. Maybe the best would be for this logging be turn off. 

## Logs
From XMPP container (inside):

These four lines of code is written to log when a user authnticates, /var/www/authentication/auth.log
I have replaced the plain text information with `<USER>` and `<PLAIN TEXT PASSWORD HERE>` for my own privacy.
```
[2021-05-06T10:31:29.390379+00:00] ejabberdAuth.DEBUG: Input detected, reading 37 bytes... [] []                                                                                                                                                                                
[2021-05-06T10:31:29.390522+00:00] ejabberdAuth.DEBUG: Input read was: auth:<USER>:<PLAIN TEXT PASSWORD HERE> [] []                                                                                                                                                              
[2021-05-06T10:31:29.592977+00:00] ejabberdAuth.DEBUG: Successfully wrote output [1] to stdout [] []                                                                                                                                                                            
[2021-05-06T10:31:29.593103+00:00] ejabberdAuth.DEBUG: Clearing previous command data [] [] 
```
The information in log file is persistent as long as the container isn't stop, restarted or killed. And add more, and more users as they authticate on there xmpp clients.

## Reproduction
1. Open xmpp client and logon with a user
2. ssh to the xmpp/jabber container
[...]

PR #4089, opened 2021-05-06
[Ejabberd] Stop plain text password/user from being logged to file in auth.log (#4087)

ghost:
Ejabberd did log the users username/password in plain text to a logfile location.
Read more in issue, #4087


andryyy:
Thanks! I will rebuild the container.

Issue #4097, opened 2021-05-15
backup_and_restore.sh breaks when there are more then one name matching, e. g.


jatty:
## Summary

Script breaks when there are more then one name matching, e. g.

## Logs

docker network ls -qf name=mailcowdockerized_mailcow-network
1884c0299a88
ad4b985980b4

It doesn't break anymore when I use ^ and $:
root@mail01vp:/opt/mailcow-dockerized# docker network ls -qf name=^mailcowdockerized_mailcow-network$
1884c0299a88

## Reproduction

helper-scripts/backup_and_restore.sh.orig backup mysql
bash: warning: setlocale: LC_ALL: cannot change locale (de_DE.UTF-8)
Backup location (absolute path, starting with /): /mnt
Using /mnt as backup/restore location.

Found project name mailcowdockerized
Using SQL image mariadb:10.5
mysql://mailman:${DBPASS}@database/mailmandb?charset=utf8mb4&use_unicode=1
mysql://mailman:${DBPASS}@database/mailmandb?charset=utf8mb4&use_unicode=1
mariadb:10.5, starting...
Unable to find image 'ad4b985980b4:latest' locally
docker: Error response from daemon: pull access denied for ad4b985980b4, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
See 'docker run --help'.

I've made a patch using a more specific grep with ^ and $ (beginning and end of string) which works for me.

andryyy:
Fixed with #4098

PR #4098, opened 2021-05-15
Update backup_and_restore.sh

jatty:
Script breaks when there is more then one name matching,  see Issue #4097


andryyy:
Thanks, looks good. :)

PR #4099, opened 2021-05-19
Add pre and post update hook

alxhlz:
This pr adds a pre and post update hook to mailcow which can be helpful if you run some kind of reverse proxy in front of your mailcow installation that has to be restarted due to several reasons.

It simply checks if the required files exists and runs them.

`pre_update_hook.sh`
`post_update_hook.sh`

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -6,9 +6,11 @@ if [ "$(id -u)" -ne "0" ]; then
   exit 1
 fi
 
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
 # Run pre-update-hook
-if [ -f "$PWD/pre_update_hook.sh" ]; then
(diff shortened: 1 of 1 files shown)


andryyy:
Hi,
Can you please add it to the docs?

alxhlz:
Done in https://github.com/mailcow/mailcow-dockerized-docs/pull/284

alxhlz:
image

MAGICCC:
What about to put them in `helper-scripts` folder?

alxhlz:
Hmm, I am open to that change but isn't the helper-scripts folder for already working and existing scripts that ships with mailcow instead of custom scripts by the user?

MAGICCC:
Yes indeed. Maybe it's just me and my head that I like having a clean root folder and putting scripts in subfolder(s) :)

PR #4100, opened 2021-05-19
Remove old SOGo Admin Login hint

SnailShea:
Currently in "Mail Setup" page of Mailcow UI, a help message states "Administrative SOGo logins are only available for mailboxes with an existing SOGo profile. A user must have been logged in to SOGo at least once.". This is no longer true, so I have created a pull request to remove the string from relevant language files as well as the web UI.


andryyy:
Merci. :) As discussed this seems to be fine.

Issue #3892, opened 2020-12-08
Extend random alias generation for permanent alias


schemen:
## Summary
For privacy concerns it random aliases are great. "Sign in with Apple" already uses this system. Once a service is no longer needed and/or abuses the email you can delete the alias.

Currently, you can only create a time constrained alias. I would like the ability for a user to create permanent aliases (maybe also per domain or per user restrictions?) with a comment field to note what service this alias is being used for.

## Motivation

This will improve privacy for a enduser and make it harder for data kraken like websites to milk email addresses.

patschi:
Why not using Aliases in Mail Setup?
![image](https://user-images.githubusercontent.com/2029878/101530758-3efd7480-398a-11eb-8a2c-003dbbf4d661.png)

schemen:
Endusers can't do them, so I thought why not expand the available enduser interface.
![image](https://user-images.githubusercontent.com/5279155/101531796-01054e00-3994-11eb-903b-8352cece5776.png)

patschi:
Ah, I see. Forgot that bit. Now it makes sense, thanks!

schemen:
A push for visibility \o/ would still be great feature! I already deleted a forwarding email of a very spamy service that I created by hand :)

schemen:
From discussions within the Telegram group:
* The database entry for permanent aliases is already here as well as the infrastructure to serve them
* It's mostly UI that needs to be done for them
* Each user should be able to create a random alias (automatically generated) and add a optional comment to describe for what service it's being used
* Optionally, a via admin defined and select privacy domain (for example  @protect.mymail.io) where users can create the alias with. The options for a user can be a privacy alias with their standard domain or the privacy domain

schemen:
Created https://github.com/schemen/privacycow as a CLI tool to cover the requirements here. Serving me as long as this is not implemented in the UI :)

Uses the Mailcow API.

thxrben:
Is there any update on this yet?
Bumping :)

Handrail9:
I would also like this feature, a CLI tool is handy for myself but I'm actually extending my Email service to my family and would like them to utilize private emails as well :)

andrebnc:
I support this proposal.

ar1ja:
it has been 4 years, still no progress lol

PR #4101, opened 2021-05-19
WIP: Add essentially unlimited aliases

schemen:
For users that want "permanent" aliases that last a 100 years.
Done this was so it won't require a major overhaul.

Not sure if you want it implemented that way but this could be a "quickfix" for #3892

Let me know if I missed anything, my PHP is very, very rusty!

### DIFF
--- a/data/web/inc/functions.mailbox.inc.php
+++ b/data/web/inc/functions.mailbox.inc.php
@@ -35,7 +35,7 @@ function mailbox($_action, $_type, $_data = null, $_extra = null) {
           else {
             $username = $_SESSION['mailcow_cc_username'];
           }
-          if (!is_numeric($_data["validity"]) || $_data["validity"] > 672) {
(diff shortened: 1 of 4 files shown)


schemen:
I did find an error, after figuring out how to setup a test env

`MySQL error: PDOException: SQLSTATE[22003]: Numeric value out of range: 1264 Out of range value for column 'validity' at row 1 in /web/inc/functions.mailbox.inc.php:63 Stack trace: #0 /web/inc/functions.mailbox.inc.php(63): PDOStatement->execute(Array) #1 /web/json_api.php(174): mailbox('add', 'time_limited_al...', Array) #2 {main}`

I'll come back with a fix.

andryyy:
Hi, thanks, but as you see these are limits we set in the code. :)

If it's unlimited it is not really time limited anymore. We could perhaps make this configurable. But let's not add workarounds in the code.

schemen:
Yeah I came to the same conclusion. Consider this PR WIP, I'll try to implement it the full way.

PR #4105, opened 2021-05-23
[Web] Fix admin ui html label references

TophEvich:
Many of the labels referenced inputs that did not have the IDs they were supposed to have.
As some of those inputs on the page have the same name but in different forms, the ID could
not always be the same as the name.
For these cases, to avoid conflicts, the forms data-id is used as a prefix for the id attribute.

Also a rogue " was removed.

--- 

I'm new to the project, and in the 'contributing guidelines' I didn't find anything about PRs or code contributions.
So this is a first little PR as an introduction to your project and your processes.


andryyy:
Merci, looks good. :) Thanks!

PR #4107, opened 2021-05-24
[Web] Change include/require/file-read paths to use __DIR__

TophEvich:
Changes the include/require/file-read paths to use the recommended `__DIR__` instead
of $_SERVER['DOCUMENT_ROOT'].
In addition of being more robust, it also helps with static code analysis.


TophEvich:
The diff is way bigger than intended, which I only saw after the fact, because some of the file still had CRLF instead of only LF.
If this is an issue I can split the changes. The question here would be if you would want a seperate CRLF -> LF pull request, or just a separate commit.

andryyy:
Can you separate it? :) That would be really nice!

Furthermore, is the __DIR__ parameter really worth the change?

TophEvich:
> Can you separate it? :) That would be really nice!

Separate PRs or commits? 

> Furthermore, is the `__DIR__` parameter really worth the change?

This depends on your future plans you have for the code-base this commit touches. If you plan to extend or support it for a longer time, I'd say it is worth it. As mentioned in addition to it being more robust, development should be easier as static code analyzers can understand the paths more easily.
At least for me that helps with development, it's less error prone and thus as a by-product also faster.

andryyy:
Thanks for the explanation!

A separate PR would be great. If it's too much hassle, we can merge it like this.

It has been an open discussion for a while and we tried not to break too many installations with custom changes. But it's a necessary step indeed.

TophEvich:
Regarding the separate PR for fixing the line feed. This would impact a lot more files in /data/web.
The questions is now, if I should limit it to .php files or if js and css would also be OK to fix.
For PHP I just would change them for all, as the same issue would just come up for change of a different file.

The issue would be the same for .js and .css, but I just want to ask before I make the commit, as this is a bigger scope as the one we discussed before. 

(75 out of the 2541 files in total in /data/web)

TophEvich:
This PR is now depended on #4108

TophEvich:
Rebased of new version of #4108

PR #4108, opened 2021-05-24
[Web] Fix line feed of 75 files

TophEvich:
This fixes the line feed (CRLF to LF) in 75 files.
(To avoid problems with future changes regarding these files.)

This is a per-requisit for #4107


MAGICCC:
Somewhat also related to https://github.com/mailcow/mailcow-dockerized/issues/3129

andryyy:
Why not CRLF?

TophEvich:
> Why not CRLF?

In general the industry/communities have converged on using LF and not CRLF (I guess because as the Internet runs on Linux).
Specifically one of those communities would be the PHP community: https://www.php-fig.org/psr/psr-12/

In reality the bigger issue is that you want to have one and the same for all files in your project, CRLF or LF, but not both, as it becomes harder to maintain, he linked PR is an example of one type of the headaches it brings up.

andryyy:
Thanks! No, I'm really just trying to understand things. So it seems to be the way to go then.

Do you want to implement Peters FR in your PR?

TophEvich:
> Do you want to implement Peters FR in your PR?

This could have bigger side-effects, so maybe I do it in a separate PR?
Below is what I would add to the project. That's what I found as recommended.

``` .gitattributes
###############################
# Git Line Endings            #
###############################

# Set default behavior to automatically normalize line endings.
* text=auto

# Force batch scripts to always use CRLF line endings so that if a repo is accessed
# in Windows via a file share from Linux, the scripts will work.
*.{cmd,[cC][mM][dD]} text eol=crlf
*.{bat,[bB][aA][tT]} text eol=crlf

# Force bash scripts to always use LF line endings so that if a repo is accessed
# in Unix via a file share from Windows, the scripts will work.
*.sh text eol=lf

###############################
# Git Large File System (LFS) #
###############################

# Archives
*.7z filter=lfs diff=lfs merge=lfs -text
*.br filter=lfs diff=lfs merge=lfs -text
*.gz filter=lfs diff=lfs merge=lfs -text
*.tar filter=lfs diff=lfs merge=lfs -text
*.zip filter=lfs diff=lfs merge=lfs -text

# Documents
*.pdf filter=lfs diff=lfs merge=lfs -text

# Images
*.gif filter=lfs diff=lfs merge=lfs -text
*.ico filter=lfs diff=lfs merge=lfs -text
*.jpg filter=lfs diff=lfs merge=lfs -text
*.pdf filter=lfs diff=lfs merge=lfs -text
*.png filter=lfs diff=lfs merge=lfs -text
*.psd filter=lfs diff=lfs merge=lfs -text
*.webp filter=lfs diff=lfs merge=lfs -text

# Fonts
*.woff2 filter=lfs diff=lfs merge=lfs -text

# Other
*.exe filter=lfs diff=lfs merge=lfs -text
```

This is not exactly what is requested in the FR, but git should do the right thing here (convert it as needed on the platforms), thus `auto` in general for text.
(Sorry if this inflammatory, but most problems will come up if you use a gaming OS for work :P)

MAGICCC:
@topelrapha My FR was just a overview what options you have using the `.gitattributes` file

topelrapha:
@MAGICCC I think i am the wrong one :)

TophEvich:
@MAGICCC & @andryyy How do you want to proceed?

andryyy:
I'm a bit afraid some files will fail in scripts when the line ending changes; these worries are not unfounded. I had run into problems previously. Perhaps we should apply that to the data/web directory first?

TophEvich:
I rebased on current master, as there were already conflicts, and added the `.gitattributes` in `data/web`.

PR #4112, opened 2021-05-26
[Web] Fix admin ui title_name update

geektoor:
In #4105 there was a small error that renamed the input name to
input id, so we lost the name and the title name couldn't be
changed anymore. This should fix it.


andryyy:
Merci

PR #4118, opened 2021-06-04
Option to update if updated code is available

74k1n984ckmyp21v4cy:
If the script notices that updates are available, it gives the user the option to update right away.

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -123,9 +123,13 @@ while (($#)); do
         exit 99
       fi
       if [[ -z $(git log HEAD --pretty=format:"%H" | grep "${LATEST_REV}") ]]; then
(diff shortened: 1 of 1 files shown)


geektoor:
Hi,

this might break "monitoring" of mailcow updates. For example you run just the check to see if something is available or not, but with your change it will wait for an input.
I, as a user, would not be too happy about it, as i use -c to check for an update, not to actually do an update.

Cheers,
Sven

patschi:
That's also not a good idea, when the update script is updated. When you run the bash script straight after updating via git, it will run the old version and not the new code.

74k1n984ckmyp21v4cy:
Didn't think about that, but valid arguments against the idea.

Thank you for the feedback.

PR #4127, opened 2021-06-06
Remove left smtpd_last_auth statement

waja:
Removes smtpd_last_auth the which was missed in 05f6e28.


dragoangel:
Wow, missed one

PR #4142, opened 2021-06-11
[Rspamd] Fix FREEMAIL_POLICY_FAILURE with SPF_SOFTFAIL

dragoangel:
Add really low negative score to SOFTFAIL policy symbols to get FREEMAIL_POLICY_FAILURE triggered correctly


andryyy:
Couldn't you just modify the composite for the free mail rule?

dragoangel:
I can, but* this change will display in rspamd that `softfail` isn't neutral and by itself will not change general scores.

Issue #3904, opened 2020-12-18
Automation improvements to update script.


keslerm:
## Summary
The update script is hard to automate because the update script checks for an update itself and asks for a restart. The force mode also seems to not work in conjunction with `--gc` and as such still asks a question about cleaning up.

## Motivation

What are you about to solve or improve with this idea?

I have main things:

1. Add a argument that just checks for an update for the update script itself `--update-script` a script can run that first to ensure the update process itself runs the actual update
2. Ensure that `--gc` respects the `-f` flag

What would be the benefit for most users?

You would be able to automate the process more efficiently but running the update script update script update in a different step and run the garbage collect step in a different step without it asking yes or no.

I can update the script to do both of these things if it sounds okay.

andryyy:
Hi, thanks! Hm. That's a difficult topic. I personally am against automatic updates. Sometimes an update prints important information or warnings that we want the user to read. 
I know there is a mechanism that kind of works, but IMO we should even remove that.

I will leave this open for the community to decide. Okay?

keslerm:
TBH I think Automation is a more of a misnomer here because it's not happening automatically, more just allowing external tools to some additional control over the update process.

I bring it up because I use ansible to handle updating most of my VMs, but it's not a process that kicks off automatically - i manually trigger it.

It's really more about having finer grained controls around individual steps in the update process to allow an external actor to use it's abilities to run it in a controlled manner.

andryyy:
I see. I will add it to my list, feel free to remind me if I forget it. You can also join our group in telegram by the way.

keslerm:
Hey you do fantastic work, I can easily help here - I'll open a PR with some changes for this. I just like opening an issue first to open any discussion before i make any changes in case there is any objection.

andryyy:
Thank you. :) That would be great!

lpmartineau:
I would like to add my vote for this feature. I also use Ansible with some Jenkins to manage my infrastructure and being able to use your excellent upgrade script with external tools would be fantastic. 
Thank you for all that you do. Have a great day.

clauso:
Came accross the problem that gc was not run and my disk was full today. Obviously `-gc -f` was doing nothing as described above. I put a fix in the update.sh script and created a pull request: https://github.com/mailcow/mailcow-dockerized/pull/4147

PR #4147, opened 2021-06-16
Run garbage collection with -f without asking questions

clauso:
This PR addresses one of the problems described here: https://github.com/mailcow/mailcow-dockerized/issues/3904
In the update.sh script "Ensure that --gc respects the -f flag". With this change `./update.sh --gc -f` will run the garbage collection without asking for confirmation.


andryyy:
Looks good, thank you.

Issue #3456, opened 2020-04-08
Not required die in backup script


mkapra:
**Is your feature request related to a problem? Please describe.**
I get the following error when using the backup_and_restore.sh script:
`/backup/mail is not write-able for others, that's required for a backup.`

**Describe the solution you'd like**
I commented out the following if-condition and the script was working.
So my question is why the directory has to be write-able for others?

https://github.com/mailcow/mailcow-dockerized/blob/59b38c1111ff52036084ae5039605f6651041348/helper-scripts/backup_and_restore.sh#L43-L46

andryyy:
Some backups need this. Not everything is written as the user you are running the script.

I implemented it for a good reason. I should have documented it, but my lack of memory today does not mean it isn't there for a good reason. 😄

mkapra:
Hm okay. A diff on a directory with the write permissions for others enabled and without this permissions and without the check said there are no missing files or something.
But if there were some errors I will only comment this out in my repository.
It is essential for me, that not everybody can read the backup.

clauso:
I made a pull request for this and put some time into the reasoning there, why this check should be removed: https://github.com/mailcow/mailcow-dockerized/pull/4148

PR #4148, opened 2021-06-17
Backup Script should not require backup location to be world writable

clauso:
See this issue:
https://github.com/mailcow/mailcow-dockerized/issues/3456

I removed the unnecessary check. I tripped over it while setting up a backup and maintained my own modified copy of the backup script like many others.

Reasoning:

1. The check does not help: The user running the backup of course needs to be able to write to the backup location, sure. But the check only verifies that anyone is allowed to write to the backup. I cannot come up with any scenario where this check might be useful.
2. The check seems to be flawed as it will accept permissions ??5-7, which means read permission for others and at least one of write or access permission or both. That seems to be wrong. 
3. The implementation seems to be overly complicated: The `$(echo $(stat` is weird it executes a command built from the stat output of a world writable directory. I cannot come up with a good way to exploit this right now, but it seems strange not to write a much simpler check like: `stat -Lc %a ${BACKUP_LOCATION} | grep -qE '^[0-9][0-9][5-7]$'`
4. The check is harmful: This pushes users to make their backup location writable/readable to all users, which is very bad practice. Remember the backup contains all data, passwords etc. No need to hack anything if you can get the backup. Also a malicious user might corrupt some data but should not be able to corrupt the backup. I have worked many years as a pentester and believe me: We got into so many machines/services just because we could access an unprotected backup lying around.


clauso:
Correction on point 2. The check will show the error on the mentioned conditions, but those are still strange.

* No error on only read permission,eg. 004
* No error on only write permission, eg. 002
* No error on only access permission, eg. 001
* Error on write and read permission combo, eg. 006
* Error on read and access permission combo, eg. 005
* Error on read, write and access permission combo, eg. 007

So the check does not really verify the the backup dir is actually writable? I am either misreading this or the check is broken and never worked as advertised in the error message.

i5hi:
This issue should not have been closed. 

The script does not verify the backup dir is writable for others. My backup dir has permission 722 and still throws the same error. It only resolves if permission is 777 which is not safe.

clauso:
Just noticed that I am still maintaining my own copy of the script since the last two years...

milkmaker:

do4bmw:
Still a issue ...

DerLinkman:
Did a restore works if the permissions of the archives and/or the backup folders are not 777. Is the backup consistent and will it actually be done?

milkmaker:

PR #4150, opened 2021-06-18
Remove X-XSS-Protection

xopez:
X-XSS-Protection has been removed in modern browsers:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection
So this header doesn't do anything anymore and is useless.

### DIFF
--- a/data/conf/nginx/includes/site-defaults.conf
+++ b/data/conf/nginx/includes/site-defaults.conf
@@ -15,7 +15,6 @@
 
   add_header Strict-Transport-Security "max-age=15768000;";
   add_header X-Content-Type-Options nosniff;
-  add_header X-XSS-Protection "1; mode=block";
   add_header X-Robots-Tag none;
   add_header X-Download-Options noopen;
   add_header X-Frame-Options "SAMEORIGIN" always;


ghost:
How about backwards compatibility?

mkuron:
@AlexBeakes, nobody should be running outdated browsers, it is just too dangerous these days. We should only keep `X-XSS-Protection` if there are any browsers that support it and are still receiving security updates. According to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection that would only leave Safari (though looking at the WebKit source code, that is probably no longer the case) and Internet Explorer 8 (which might still be supported on Windows Embedded POSReady 7, Windows Embedded 8 Standard, or Windows Embedded 8.1 Industry/Pro). So I'd say it's safe to remove.

xopez:
Here is another overview with more browsers: https://caniuse.com/mdn-http_headers_x-xss-protection

mkuron:
Okay, so IE 11 and Safari support it and are definitely still within their lifecycle. The Samsung browser even just added this feature two months ago. So let's close this pull request for now.

PR #4155, opened 2021-06-23
allow send mail to catchall mailboxes for non-existent addresses covered

macwinnie:
Hi everyone =)

I use one regular mailbox, that handles most of my email addresses – by now the incoming part. From now, I also want that mailbox to handle outgoing mails for addresses mapping the catchall but not being another alias / mailbox.

For that, I've made some improvements to two SQL statements for Postfix.

The first change (`mysql_virtual_alias_maps.cf`) is, that the alias-lookup resolves unknown addresses to mapping catchall addresses.

And the second change (`mysql_virtual_sender_acl.cf`) is to allow sending mails from undefined alias / mailboxes under the catchall address. _So if there is given the domain `example.com` with the two maiboxes `one@example.com`and `two@example.com` where the last one is defined as catchall, `one@` only can send mails as itself and `two@` can send mails with any other address that is not `one@`._

It would be great, if that improvement could get part of the official mailcow dockerized package =)

Best,
Martin


andryyy:
Hi, can you explain the mysql_virtual_alias_maps change?

macwinnie:
Hi @andryyy,

the change to `mysql_virtual_alias_maps.cf` is used in `virtual_alias_maps` of Postfix. For Postfix, the catchall addresses are now "known aliases" after the change – for being able to send emails as the not specifically defined alias, that change could be ignored, I think. But I also think, that being clear / clean with the alias definitions is a good thing =D

The change within the MySQL statement is re-used below with `mysql_virtual_sender_acl`. The main query selects the first non- `NULL` -value (`COALESCE`) that is returned. And for the options to be checked if `NULL`, we use two sub-queries:

1.  the already known one, where we check if there is a fully mapping entry for the alias and
2. the new query, that checks if there is a catchall entry

The first non-NULL value is used to be returned.

I hope, that explanation is the one, you need =)
Best,
Martin

andryyy:
The catch-all is already working. Why would the change to the alias map be necessary? Also “I think” is pretty dangerous when changing the alias map. 

     the change to mysql_virtual_alias_maps.cfis used in virtual_alias_maps of Postfix.

I understand what function the map has. :P 

Postfix queries sequentially by domain, mailbox etc. - I don’t understand this specific change.

andryyy:
Furthermore: why “reused below”? No. They are independent maps. Perhaps the content is copied to the other map, besides that they don’t interfere. :/

You explained the query itself and that the alias map is an alias map. But why exactly is this necessary now?

And why does it improve security? And what does “allow send mail to catchall mailboxes for non-existent addresses covered” mean? I don’t get any of it. :/ These changes can have a **huge** impact and should be explained in detail. Use postalias for example to explain what works now and what didn’t work before, please. Keep in mind that Postfix queries the same map multiple times, where %s varies.

macwinnie:
### Security

`improve security` means the specific change in the specific commit:

* my first version allowed catchall users to send as any mail with the specific domain, even if there existed a mailbox or another alias
* that was insecure
* for that, the specific commit increased the security of my pull request

### Alias Map

For the change with the alias map – the alias maps represent ALL alias, the mailserver knows. So if one has `a@b.c` as a catchall defined for `@b.c`, there is a non-written alias `d@b.c` and even `z@b.c`. The current alias map (point of discussion above) does not reflect those non-written alias. My change would have reflected them.

Since I cannot tell you the functionality in specific, where that is an improvement, I've reset that change so we do not have to discuss it further.

### Main change with the Virtual sender ACL

That file / query is used to check, if a specific mailbox is allowed to send as the specific mail address.

#### Current mailcow state

Currently, one can only send emails as clearly written alias addresses / mailboxes. So if I have the mailbox `a@b.c` as catchall for `@b.c` and another mailbox `d@b.c`:

* `a@b.c` can send mails as `a@b.c`
* `a@b.c` cannot send mails as any other address unless one defines the alias
* `d@b.c` can only send mails as itself

#### Change by the pullrequest

given the same scenario:

* `a@b.c` can still send mails as itself
* also `a@b.c` can send mails as all undefined mail addresses `*@b.c`, e.g. `foo@b.c` or `bar@b.c`
* `a@b.c` cannot send mails as `d@b.c`
* `d@b.c` can still send mails only as itself

Hope, that does explain, what did not work until the change request.

PR #4157, opened 2021-06-24
Update lang.it.json

towerman86:
Translated more voices


MAGICCC:
Cool thanks. Do you plan to create more commits? If yes then we dont need to merge the PR now

towerman86:
I'm still translating as much as I can, but I prefer to save them here for not losing something 👍 
Do I have to tell you when I'm done?

towerman86:
> Cool thanks. Do you plan to create more commits? If yes then we dont need to merge the PR now

I'm still translating as much as I can, but I prefer to save them here for not losing something 👍 
Do I have to tell you when I'm done?

MAGICCC:
Sure no problem, just tell us here when we can merge :)

PR #4161, opened 2021-06-29
Update lang.it.json

towerman86:
@MAGICCC You could upload it.
I have to stop translating for a bit 😔


MAGICCC:
Oh ok.
But thanks a bunch!

PR #4162, opened 2021-06-29
use random password for mailcow-reset-admin.sh

### LINKED ISSUE #3921: Reset admin script with random password
At the moment, the script to reset the admin password, sets it to default value ( admin:moohoo ).
For security reason should be set to a random value.
Even if it is highly unprobable that someone unauthorized ( or a bot ) could access the admin account before changing the default password, it's still possible.

Best regards

prey87:
Closes #3921


andryyy:
Taking a quick glance I doubt it will work. Did you test it?

Please also use ${bla} just for integrity. :)

prey87:
Curly braces added.
I have also tested the script ;D

andryyy:
It’s strange it would expand the var in the single quotes.

prey87:
what do you mean exactly?

andryyy:
The `'$password'` stuff. But I'm wrong, it works indeed.

Thanks. :)

PR #4171, opened 2021-07-06
[web] Responsive Design

feldsam:
Complete styling for mobile devices. Tested on iPhone 11

### DIFF
--- a/data/web/admin.php
+++ b/data/web/admin.php
@@ -12,7 +12,7 @@ if (!isset($_SESSION['gal']) && $license_cache = $redis->Get('LICENSE_STATUS_CAC
 ?>
 <div class="container">
 
-  <ul class="nav nav-tabs" role="tablist">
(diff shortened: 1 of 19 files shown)


feldsam:
maildev feldhost cz_mailbox_lang=en

ghost (approved):
🎉

andryyy:
bootstrap-tabcollapse was last updated 6 years ago. My plan is to migrate to BS5.

feldsam:
Hi, but it is working good. When you update to BS5, then we can replace it with something new.

feldsam:
@andryyy what about this? update to BS5 is a big step and takes time. There are many changes and new functions, like flex grid. I think we should implement a templating system like Twig during the update to BS5. I am ready to help, but in meantime, we can have at least this responsive design on old code.

btw, I added two more lines to `tabcollapse` script to make it work better with Mailcow, it is a simple script, which I can maintain.

PR #4175, opened 2021-07-06
[web] dns diag spf record syntax link updated

feldsam:
There was a long time this SPF record syntax helper until the site was down. It is again live on new domain.


andryyy:
Even though it's http only, it still seems to be much more informative.

PR #4176, opened 2021-07-06
[redis] raise container net.core.somaxconn

feldsam:
On busy installations, default somaxconn is not sufficient. This raises it directly in the container because raising it in the host doesn't do it.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -49,6 +49,8 @@ services:
         - "${REDIS_PORT:-127.0.0.1:7654}:6379"
       environment:
         - TZ=${TZ}
+      sysctls:
+        - net.core.somaxconn=4096
       networks:
         mailcow-network:
           ipv4_address: ${IPV4_NETWORK:-172.22.1}.249


patschi:
It might make sense indeed on busier setups, but I'm unsure if we should set this manually.

According to [here](https://levelup.gitconnected.com/linux-kernel-tuning-for-high-performance-networking-high-volume-incoming-connections-196e863d458a) it's however set to a default of 4096 starting Linux 5.4:
```
Accept Queue Default
The default value for net.core.somaxconn comes from the SOMAXCONN constant, which is set to 128 
on linux kernels up through v5.3, while SOMAXCONN was raised to 4096 in v5.4. However, v5.4 is the 
most current version at the time of this writing and has not been widely adopted yet, so the accept 
queue is going to be truncated to 128 on many production systems that have not modified net.core.somaxconn.
```

So Ubuntu 20.04, shipping with Linux 5.4, should not be affected here. However in this case, we would limit the value to a lower one than the Linux kernel has...

To just keep special settings to a minimum, which might bite us in the future, I'd suggest documenting the issue (with log excerpts when this occurs) somewhere in the docs and how it can be increased if necessary. What's your opinion here, @andryyy?

feldsam:
What about using config var in mailcow.conf ?

patschi:
I think we should set `net.core.somaxconn = 4096` as per Linux 5.4 default.

andryyy:
Let's set it to 4096 globally. Can you edit the PR? :)

feldsam:
Done

patschi:
Thanks, @feldsam!

pquan:
Hello,
This is a breaking change for all of us that use mailcow in a container environment. I know, this is not supported, officially.
However, introducing "features" that fix a problem in one environment and create problems in other environments shall be "optional" and not by default.
BTW: This setting is 4096 by default in my debian/ubuntu servers. No need to set a 4096 "default" by default :)
Thanks.

PS: net.core.somaxconn is not available under LXC/LXD.

andryyy:
That’s a valid concern. @mkuron ?

feldsam:
@pquan users doesn't use only Debian/Ubuntu distros, also there are older where default is only 128? What does this breaks?? when it sets it to default from the latest kernel? What does this project have with LXC/LXD?

andryyy:
He means that this parameter cannot be used in LXC systems.

It is not supported anyway, but it’s probably a breaking change for those to run it on these platforms anyway. 

I’m unsure about it. It is probably not a big deal for those to run their cow on LXC to remove that parameter. They should be aware it is an unsupported platform.

pquan:
> He means that this parameter cannot be used in LXC systems.

Yes, LXC isolates most kernel parameters from the container, as a **security measure**. Docker inside LXC/LXD is supported and it actually works very nicely on **unprivileged containers** (including mailcow). This means **no root access for the whole mailcow** and isolation between different mailcow instances running on the same server. With all the kernel hacks and hardware hacks we have lately, this is only a good thing. And you still get to share some memory and other available resources on the same server, which you cannot do with vmware/kvm. 

> It is not supported anyway, but it’s probably a breaking change for those to run it on these platforms anyway.
> I’m unsure about it. It is probably not a big deal for those to run their cow on LXC to remove that parameter. They should be aware it is an unsupported platform.

It does not look nice to introduce a breaking change and limit the possible use cases of mailcow in general. And this change will likely break existing installations on update.

I also question the utility of this parameter for the future. This is the number of pending tcp/ip connections (backlog). As said, currently this is 4096 on debian/ubuntu and appears more than enough. Older systems with 128 connection are the exception and perhaps should not be used, specially for such (hosing/service) loads and they need extra tuning beyond the soconnmax.

Ultimately, I understand LXC is not supported, but perhaps is should be? Mailcow runs very nicely in it. And like I said, with LXC you get a perfect mailcow isolation, security, sharing of resources where available, private networking without the need to touch docker-compose. A lot of other things get easier when using a container outside docker. LXC supports nested containerization and that is all you need to use mailcow.

andryyy:
Oh come on. There is so much wrong in your post. Don’t try to turn things to make it look like it is a bad decision, insecure and all that. That’s bullshit. 

You are running it on a not more and not less secure environment, that is not supported. Deal with it. 

Unsubbed.

PR #4177, opened 2021-07-07
[web] helper script for detecting missing translations

feldsam:

### DIFF
--- a/helper-scripts/add-new-lang-keys.php
+++ b/helper-scripts/add-new-lang-keys.php
@@ -1,0 +1,63 @@
+<?php
+
+function array_diff_key_recursive (array $arr1, array $arr2) {
+  $diff = array_diff_key($arr1, $arr2);
+  $intersect = array_intersect_key($arr1, $arr2);
+
+  foreach ($intersect as $k => $v) {
+    if (is_array($arr1[$k]) && is_array($arr2[$k])) {
+      $d = array_diff_key_recursive($arr1[$k], $arr2[$k]);
+
+      if ($d) {
+        $diff[$k] = $d;
+      }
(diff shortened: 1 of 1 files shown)


andryyy:
Cool! Could you add this to the docs? I'd just change some style in the PHP, ok? Not because it is wrong or bad, just for uniformity. :D :)

MAGICCC:
ah nice, I always checked the CI results :)

feldsam:
Hi @andryyy, no problem feel free to change the code style. Regarding docs, in which section should I add it? There is nothing about translations.

ctrl-i:
That seems like a lot of code...  Would something like this one-liner be easier to maintain?

`comm -23 <( jq -r 'paths(scalars) | map(.|tostring)|join(" > ")' lang.en.json | sort ) <( jq -r 'paths(scalars) | map(.|tostring)|join(" > ")' lang.es.json | sort )`

This utilizes jq (downside, might need to be installed) to create a list of all the keys in lang.en.json,  We do this again for lang.es.json.  Finally we use comm to output all the keys that are in lang.en.json but not in lang.es.json. 

Obviously this could be automated further.

feldsam:
Hi, looks great, but I don't think so. This is easy to maintain for folks working with jq, for regular people working with frontend technologies it can be hard to understand.

andryyy:
We could add both I guess.

ctrl-i:
> We could add both I guess.

Let me know if you would like me to expand my one-liner into a script and I can work on it.

andryyy:
That would be very much appreciated :)

andryyy:
We should also add it to the docs.

PR #4178, opened 2021-07-07
[web] translated selectbox "nothing selected" str

sp8c3:
added german translation for "nothing selected"


andryyy:
Thanks. :) What do you think about making it an uppercase "Nichts"?

sp8c3:
oh of course everything else is capitalized too :)

PR #4180, opened 2021-07-07
Remove redundant password hash checks

SnailShea:
In a previous patch, the hash matching was moved into the hash_password() function. I thought I had cleaned up all the extra checks, but apparently not.


andryyy:
Thanks!

PR #4187, opened 2021-07-11
[web] enhaned responsive design

feldsam:
Fixed buttons with long titles (mostly other than EN and DE langs)

Would you please check the Apps buttons on the login page? After your commit, I had it jumping about 5 px when I add the second and third buttons. So I also added `margin-top: 0`;


andryyy:
I really appreciate your help. :) Thanks!!!

PR #4189, opened 2021-07-11
[rspamd] add soft reject on task timeout

geektoor:
As we have seen issues in DNS processing actually stops rspamd from
processing a message, which leads to missing tag insertion for example,
we turn on soft reject on task timeout. Behavior is the same as with
greylisting for example, so the mail will be delayed/soft rejected, but
as DNS issues usually are most likely temporarily, it should get delivered
on the second try.

### DIFF
--- a/data/conf/rspamd/local.d/options.inc
+++ b/data/conf/rspamd/local.d/options.inc
@@ -7,3 +7,6 @@ dns {
   retransmits = 2;
 }
 disable_monitoring = true;
+# In case a task times out (like DNS lookup), soft reject the message
+# instead of silently accepting the message without further processing.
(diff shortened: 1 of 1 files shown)


mkuron:
Good idea, especially since b5bf97eec9047511211b48344e6590ae2cd26e3b undoes #1965. I‘ll  test this for a while before we should merge it.

mkuron:
I have been running with this patch for a few days now and haven't seen any timeout-related soft rejects so far. I did get a few `query timed out` that resulted in `RBL_SORBS_FAIL` symbols, but evidently these are not considered fatal.

andryyy:
Cool, let's merge it then. :) Thank you!

PR #4192, opened 2021-07-19
[Web] Update restart button

Hexaris:
I remove repeat arrows from the initial proposal, because it looked complicated and the button was unnecessarily long.


andryyy:
Merci :)

PR #4195, opened 2021-07-24
[web] Fix bcc map deactivates when type is changed

feldsam:


andryyy:
Hi,

The `$_attr` I agree on.

But the problem with toggling active is actually related to line 138. I forgot to remove it at some point.

Thanks for the hint!

andryyy:
https://github.com/mailcow/mailcow-dockerized/commit/773383cacd61f3bb1a09bdc27d0bb3727e708610

PR #4196, opened 2021-07-24
[web] various i18n strings updated or added, sk,cs and en

feldsam:


andryyy:
Really nice, thank you, Kristian!

PR #4198, opened 2021-07-25
Slack Integration with Watchdog ✉️

GrantBirki:
# Slack Integration with Watchdog ✉️

A simple refresher of https://github.com/mailcow/mailcow-dockerized/pull/3672 to enable Slack webhooks to post Watchdog alerts directly into Slack for ease of monitoring.

Credits to @jacobbaungard

Fixes #3672

## Usage

To use this feature, all you need to do is create a Slack webhook ([docs](https://api.slack.com/messaging/webhooks)) and add the following line to your `mailcow.config` file:

```ini
WATCHDOG_SLACK_URL=<SLACK_URL>
```

You may also take advantage of the `WATCHDOG_SLACK_PING` variable to ping a user or group directly in Slack: `@username`, `@here`, `@channel`.

```ini
WATCHDOG_SLACK_PING=<@username>
```

### Examples

Docker Logs:

```console
# docker-compose logs --tail=200 -f watchdog-mailcow
watchdog-mailcow_1   | Waiting for containers to settle...
watchdog-mailcow_1   | Uptime: 1284  Threads: 12  Questions: 3198  Slow queries: 0  Opens: 59  Open tables: 50  Queries per second avg: 2.490
watchdog-mailcow_1   | Resolved MX for <redacted>
watchdog-mailcow_1   | Sun Jul 25 05:33:12 UTC 2021 Sent notification email to <redacted>
watchdog-mailcow_1   | okSun Jul 25 05:33:13 UTC 2021 Sent Slack notification
```

Slack Screenshot:

image

With `WATCHDOG_SLACK_PING` set to `@here`:

image

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -496,8 +496,8 @@ services:
         - OLEFY_THRESHOLD=${OLEFY_THRESHOLD:-5}
         - MAILQ_THRESHOLD=${MAILQ_THRESHOLD:-20}
         - MAILQ_CRIT=${MAILQ_CRIT:-30}
-        - WATCHDOG_SLACK_URL=${WATCHDOG_SLACK_URL}
-        - WATCHDOG_SLACK_PING=${WATCHDOG_SLACK_PING}
+        - WATCHDOG_SLACK_URL=${WATCHDOG_SLACK_URL:-}
+        - WATCHDOG_SLACK_PING=${WATCHDOG_SLACK_PING:-}
       networks:
         mailcow-network:
           aliases:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -496,6 +496,7 @@ services:
         - OLEFY_THRESHOLD=${OLEFY_THRESHOLD:-5}
         - MAILQ_THRESHOLD=${MAILQ_THRESHOLD:-20}
         - MAILQ_CRIT=${MAILQ_CRIT:-30}
+        - WATCHDOG_SLACK_URL=${WATCHDOG_SLACK_URL}
       networks:
         mailcow-network:
           aliases:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -497,6 +497,7 @@ services:
         - MAILQ_THRESHOLD=${MAILQ_THRESHOLD:-20}
         - MAILQ_CRIT=${MAILQ_CRIT:-30}
         - WATCHDOG_SLACK_URL=${WATCHDOG_SLACK_URL}
+        - WATCHDOG_SLACK_PING=${WATCHDOG_SLACK_PING}
       networks:
         mailcow-network:
           aliases:
--- a/data/Dockerfiles/watchdog/watchdog.sh
+++ b/data/Dockerfiles/watchdog/watchdog.sh
@@ -137,7 +137,7 @@ function mail_error() {
 
     # Send Slack Message
     if [ ! -z "$WATCHDOG_SLACK_URL" ]; then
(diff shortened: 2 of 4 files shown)


feldsam:
Hello, a nice feature. You should also add it to the `update.sh` script.

GrantBirki:
@feldsam Done! Added with the following commit https://github.com/mailcow/mailcow-dockerized/pull/4198/commits/602b63cc4c694bab2ea7744f276b5b2c46fc6244. 😃

andryyy:
What do you think, @mkuron ?

We should merge and cleanup/split the functions in the future.

docker-compose.yml
```diff
@@ -496,6 +496,8 @@ services:
         - OLEFY_THRESHOLD=${OLEFY_THRESHOLD:-5}
         - MAILQ_THRESHOLD=${MAILQ_THRESHOLD:-20}
         - MAILQ_CRIT=${MAILQ_CRIT:-30}
+        - WATCHDOG_SLACK_URL=${WATCHDOG_SLACK_URL}
+        - WATCHDOG_SLACK_PING=${WATCHDOG_SLACK_PING}
```
mkuron: Needs default values (specified with `:-` like on the other variables). Then you don‘t even need to add it to update.sh.

mkuron (changes requested):

docker-compose.yml
GrantBirki: What if the defaults should be blank @mkuron? Using the email notification variable as an example: https://github.com/mailcow/mailcow-dockerized/blob/master/docker-compose.yml#L465

Since we leave that variable blank and made changes to the `update.sh` file I would assume that I should do the same?

Example: A user does not want to use Slack so they leave these variables blank

Let me know if I should make these changes or if the PR is good as it stands. Thanks!

CC: @andryyy

GrantBirki: Did some further research and I am fairly confident that these values are not candidates for defaults with `:-`. Going to resolve this conversation and ping for review in the main thread. Thanks! 😸

GrantBirki:
@mkuron or @andryyy looking for a peer review. To my understanding the thread above is resolved.

docker-compose.yml
mkuron: They need default values. Otherwise you‘ll get error messages from docker-compose when you bring up Mailcow without these variables in mailcow.conf. So we need a default value for backwards compatibility.

GrantBirki: @mkuron I added some blank defaults in the latest commit. Let me know what you think.

GrantBirki:
@mkuron or @andryyy wanted to bump this. Any additional feedback?

GrantBirki:
Hey @andryyy @mkuron, did you have any additional feedback?

mkuron:
Looks okay, but since I don‘t use Slack I can‘t test it.

GrantBirki:
@mkuron I have tested it quite a bit in Slack and it is working for me. Not sure if we have anyone else who uses Slack and would like to test it out or weigh in.

mkuron:
In https://github.com/mailcow/mailcow-dockerized/pull/3672#issuecomment-670504416 @andryyy said 
> Can we leave this open a bit longer? I can try to use this as base for a more general HTTP notification config.

Should we still try to aim for that or is a Slack-only solution okay for now? @GrantBirki, you could easily make it a general HTTP notifier by renaming the variables from SLACK_* to WEBHOOK_* and making the payload template configurable.

GrantBirki:
That would be a good idea! I personally don't have the time ATM to work on said feature and I think the HTTP payload template should be a polished one. Shall I leave this PR open for now or close it out as the HTTP payload template seems like a good approach?

mkuron:
Just update this PR whenever you are ready.

PR #4201, opened 2021-07-28
[web] customized apps name in header also

feldsam:

### DIFF
--- a/data/web/admin.php
+++ b/data/web/admin.php
@@ -1214,7 +1214,7 @@ if (!isset($_SESSION['gal']) && $license_cache = $redis->Get('LICENSE_STATUS_CAC
             </div>
             <div class="form-group">
               <label for="uitests_apps_name"><?=$lang['admin']['apps_name'];?>:</label>
-              <input type="text" class="form-control" id="uitests_apps_name" name="apps_name" placeholder="mailcow Apps" value="<?=$ui_texts['apps_name'];?>">
+              <input type="text" class="form-control" id="uitests_apps_name" name="apps_name" placeholder="<?=$lang['header']['apps']?>" value="<?=$ui_texts['apps_name'];?>">
             </div>
             <div class="form-group">
               <label for="help_text"><?=$lang['admin']['help_text'];?>:</label>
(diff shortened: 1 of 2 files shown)


data/web/admin.php
```diff
@@ -1214,7 +1214,7 @@
             </div>
             <div class="form-group">
               <label for="uitests_apps_name"><?=$lang['admin']['apps_name'];?>:</label>
-              <input type="text" class="form-control" id="uitests_apps_name" name="apps_name" placeholder="mailcow Apps" value="<?=$ui_texts['apps_name'];?>">
+              <input type="text" class="form-control" id="uitests_apps_name" name="apps_name" placeholder="<?=$lang['header']['apps']?>" value="<?=$ui_texts['apps_name'];?>">
```
andryyy, line 1217: The PR title does not even mention this. :/ There are also other mailcow\* placeholders I want to keep.

I also tend to keep "Apps" in the header tbh.

andryyy: Hmm. Let's just remove that bit and keep the change below. What do you think?

feldsam: hi, in `$lang['header']['apps']` is "mailcow Apps", so if it is translated to "mailcow Aplikace" it is better from UI/UX perspective

data/web/inc/header.inc.php
```diff
@@ -105,7 +105,7 @@
           <li><a href data-toggle="modal" data-container="sogo-mailcow" data-target="#RestartContainer"><i class="bi bi-arrow-repeat"></i> <?= $lang['header']['restart_sogo']; ?></a></li>
           <?php } ?>
           <li class="dropdown">
-            <a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button" aria-expanded="false"><i class="bi bi-link-45deg"></i> <?= $lang['header']['apps']; ?> <span class="caret"></span></a>
+            <a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button" aria-expanded="false"><i class="bi bi-link-45deg"></i> <?=$UI_TEXTS['apps_name'];?> <span class="caret"></span></a>
```
feldsam, line 108: @andryyy I understand if you don't want to change this. I can remove it from the commit. Let me know

andryyy: Naah, let's keep it. 👍🏼 Thank you!

andryyy:
Thank you!

Issue #2026, opened 2018-11-19
Re, Execute script from sieve with vnd.dovecot.filter


RudySixOw:
Thanks you for your answer and sorry to rest but I can not reopen it and I did not get an answer.

sieve_plugins = sieve_extprograms and already in data / conf / 
dovecot / dovecot.conf, but I have the error that I mentioned.

Otherwise for the script, it's not a big deal because I just run a curl
that I tested in the bash of the container.

andryyy:
Can you please write in the closed issue? I don't know the reference here.

RudySixOw:
Hello, sorry for my english and if this question has already been asked.
I am looking to run a script (bash) with a condition.
But as soon as I put: require "vnd.dovecot.filter"or "vnd.dovecot.execute" or "vnd.dovecot.pipe", I have an error: quoted string "vnd.dovecot.filter" where require string expected.
Example:
require "variables";
require "subaddress";
require "envelope";
require "vnd.dovecot.filter";

if allof (address: contains "to" "mail@domain.com") {

 filter "script.sh";

}

Thanks in advance.

RudySixOw:
Sorry to insist but I would like to know if this function will be available because it is really necessary.
The answer that I was given as I explained is not functional in the state.
thanks in advance.

RudySixOw:
Hi, I think there is no module in /usr/lib/dovecot/sieve/ or /usr/lib/dovecot/modules/sieve/ like sieve-execute

PR #4204, opened 2021-07-31
[Web] Added vnd.dovecot.execute/filter/pipe sieve validator support

Ashus:
This fixes the problem mentioned here #2026 and enables editing the global sieve filter in GUI.
I use these extensions as well, namely to process payment notification emails from bank and for notifications via curl to localhost outside of docker container as well.
It is mostly correct according to RFC (https://github.com/dovecot/pigeonhole/blob/master/doc/rfc/spec-bosch-sieve-extprograms.txt), good enough for most use cases I tried.


andryyy:
Thank you! :)

PR #4205, opened 2021-07-31
pulling docker images by built-in functionality

macwinnie:
Since I run into unexpected problems with the `update.sh` script when changing one image name within the `docker-compose.yml` (invalid reference format – Error pulling image:, retrying...) where I was pretty sure, the reference format was correct, I inspected the `update.sh` and refactored it to the usage of builtin `docker-compose pull`.

Perhaps that is an improvement from which everyone can benefit.


andryyy:
No, we added this whole complexity for a reason. :( It is not that we were not aware of `docker-compose pull`. 😄

PR #4206, opened 2021-07-31
Update imapsync to 2.148

MAGICCC:
Also needs a rebuild of dovecot container

### **Changelog:**

- **2.140**
- **General:** New release numbering from 1.999 directly to 2.100; I want to avoid 1.1000 and sorting issues in directory listing.
- **Dependencies removed:** None
- **Dependencies added:** None
- **Enhancement:**  Check if SEARCH ALL works on both accounts and then add --noabletosearch if one can not. Use --nochecknoabletosearch to avoid that behavior.
- **Enhancement:**  Added an analyse and classification of the errors encountered. Most common error encountered is printed at the end. Exit value 112 (EXIT_WITH_ERRORS_MAX) is now replaced by the most common error encountered exit value:

         EX_OK          => 0  ; #/* successful termination */
         EX_USAGE       => 64 ; #/* command line usage error */
         EX_NOINPUT     => 66 ; #/* cannot open input */
         EX_UNAVAILABLE => 69 ; #/* service unavailable */
         EX_SOFTWARE    => 70 ; #/* internal software error */
         EXIT_CATCH_ALL              =>   1 ; # Any other error
         EXIT_BY_SIGNAL              =>   6 ; # Should be 128+n where n is the sig_num
         EXIT_BY_FILE                =>   7 ;
         EXIT_PID_FILE_ERROR         =>   8 ;
         EXIT_CONNECTION_FAILURE     =>  10 ;
         EXIT_TLS_FAILURE            =>  12 ;
         EXIT_AUTHENTICATION_FAILURE =>  16 ;
         EXIT_SUBFOLDER1_NO_EXISTS   =>  21 ;
         EXIT_WITH_ERRORS            => 111 ;
         EXIT_WITH_ERRORS_MAX        => 112 ;
         EXIT_OVERQUOTA              => 113 ;
         EXIT_ERR_APPEND             => 114 ;
         EXIT_ERR_FETCH              => 115 ;
         EXIT_ERR_CREATE             => 116 ;
         EXIT_ERR_SELECT             => 117 ;
         EXIT_TRANSFER_EXCEEDED      => 118 ;
         EXIT_ERR_APPEND_VIRUS       => 119 ;
         EXIT_TESTS_FAILED           => 254 ; # Like Test::More API
         EXIT_CONNECTION_FAILURE_HOST1     =>  101 ;
         EXIT_CONNECTION_FAILURE_HOST2     =>  102 ;
         EXIT_AUTHENTICATION_FAILURE_USER1 =>  161 ;
         EXIT_AUTHENTICATION_FAILURE_USER2 =>  162 ;

- **Enhancement:** Added option --dry1 it avoids to fetch the message on host1 in --dry mode. Option --dry1 is on when --dry is on. It speeds up --dry very much. To get the old --dry behavior, use --dry --nodry1
- **Enhancement:** Do both sides login and authentication before exiting on failure.
- **Enhancement:** --timeout can now be a float (was only integer). Same for --timeout1 --timeout2
- **Enhancement:**  Added option --syncduplicates to sync duplicates. On by default in CGI context. Off by default on command line.
- **Enhancement:**  Added option --trylogin. After a login failure of any kind, imapsync tries to login with the LOGIN command. Use --notrylogin to avoid that. --trylogin is on by default.
- **Enhancement:**  Added options --oauthaccesstoken1 and --oauthaccesstoken2 to support OAUTH2 authentication.
- **Enhancement:**  Added options --oauthdirect1 and --oauthdirect2 to support OAUTH2 authentication.
- **Enhancement:**  Added statistic line "CPU time and %cpu". Example: "CPU time and %cpu : 9.2 sec 39.8 %cpu 19.9 %allcpus".
- **Enhancement:**  Added option --filterbuggyflags to filter flags known to be buggy and generators of errors "BAD Invalid system flag" or "NO APPEND Invalid flag list".
   

-  **CGI context:** Back from Message-Id to Message-Id Received in cgisetcontext
-  **CGI context:** In CGI context, the log directory is no longer ./LOG_imapsync/ but simply ./ It should fasten the searches and grep
-  **CGI context:** Added --loglogfile option. Append ../list_all_logs_auto.txt with the logfile name. On by default in CGI context.
-  **Usability:** Changed time presentation on ETA and in final stats. Example: Wednesday 16 June 2021-06-16 00:32:05 +0200 CEST
- **Usability:** Do not check selectable folders for huge accounts with more than 152 folders, 98 percentile of /X; It could be very long and ended up with some users transformed in stone just by staring at the fixed output.
 - **Bug fix:** The number of detected cpus (cores) was accurate only on Linux. Added Freebsd and fixed Darwin and Windows parts.
 - **Refactoring:** Continuing to deglobalize the global variables.
 - **Refactoring:** 1734 unit tests.

### DIFF
--- a/data/Dockerfiles/dovecot/imapsync
+++ b/data/Dockerfiles/dovecot/imapsync
@@ -1,6 +1,6 @@
 #!/usr/bin/env perl
 
-# $Id: imapsync,v 2.140 2021/07/05 20:14:18 gilles Exp gilles $
+# $Id: imapsync,v 2.148 2021/07/22 14:21:09 gilles Exp gilles $
 # structure
 # pod documentation
 # use pragmas
@@ -25,7 +25,7 @@ and without duplicates.
 
 =head1 VERSION
 
-This documentation refers to Imapsync $Revision: 2.140 $
+This documentation refers to Imapsync $Revision: 2.148 $
 
 =head1 USAGE
 
@@ -243,7 +243,7 @@ Same thing for user2 password.
  --domain2      str  : Domain on host2 (NTLM authentication).
 
  --oauthaccesstoken1 str : The access token to authenticate with OAUTH2.
-                       It will be combined with the --user1 value to form the
+                       It will be combined with the --user1 value to form the 
                        string to pass with XOAUTH2 authentication.
                        The password given by --password1 or --passfile1
                        is ignored.
@@ -252,18 +252,18 @@ Same thing for user2 password.
                        If the value is a file, imapsync reads its first line
                        and take this line as the access token. The advantage
                        of the file is that if the access token changes then
-                       imapsync can read it again when it needs to reconnect
+                       imapsync can read it again when it needs to reconnect 
                        during a run.
 
 
- --oauthaccesstoken2 str : same thing as --oauthaccesstoken1
+ --oauthaccesstoken2 str : same thing as --oauthaccesstoken1 
 
  --oauthdirect1 str  : The direct string to pass with XOAUTH2 authentication.
-                       The password given by --password1 or --passfile1 and
+                       The password given by --password1 or --passfile1 and 
                        the user given by --user1 are ignored.
 
  --oauthdirect2 str  : same thing as oauthdirect1
-
+ 
 
 =head2 OPTIONS/folders
 
@@ -639,7 +639,7 @@ https://imapsync.lamiral.info/FAQ.d/FAQ.Gmail.txt
 
 =head2 OPTIONS/behavior
 
- --maxmessagespersecond flo : limits the average number of messages
+ --maxmessagespersecond flo : limits the average number of messages 
                               transferred per second.
 
  --maxbytespersecond int : limits the average transfer rate per second.
@@ -1263,7 +1263,7 @@ sub single_sync
 # or $acc1->{variable_name}
 # or $acc1->{variable_name}
 
-#
+# 
 $acc1 = {} ;
 $acc2 = {} ;
 $sync->{ acc1 } = $acc1 ;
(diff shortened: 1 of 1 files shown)


svengo:
I'm a little confused. Wouldn't it be better to download the `imapsync` script in the `dockerfile` directly from github? Such a download is already done [elsewhere](https://github.com/mailcow/mailcow-dockerized/blob/master/data/Dockerfiles/dovecot/Dockerfile#L88).

andryyy:
There seems to be an even newer version @ https://imapsync.lamiral.info/imapsync :)

I don't think we should auto-download, it may be necessary to add new dependencies.

MAGICCC:
> 
> 
> There seems to be an even newer version @ https://imapsync.lamiral.info/imapsync :)

Indeed, but on his [changelog](https://imapsync.lamiral.info/S/news.shtml) he only mentioned 2.140 which I guess is somewhat stable maybe? I can upgrade the PR to 2.148, but no idea what got changed there without checking the whole code.

feldsam:
> There seems to be an even newer version @ https://imapsync.lamiral.info/imapsync :)
> 
> I don't think we should auto-download, it may be necessary to add new dependencies.

This is an example Dockerfile https://imapsync.lamiral.info/INSTALL.d/INSTALL.Docker_build.txt, where is a script for checking dependencies.

andryyy:
But that script does not really help. How would it? The dep check fails and then the images does not build. I don't quite get it. :/

andryyy:
It would at least fail to build, hum. But I really prefer manual testing. Not sure yet. :)

feldsam:
It was just reply to required dependencies check. Manual testing is of course better from QA perspective, so better for all users. I have in mind only autodownload of specific version defined in build arg, like in SOGo Dockerfile, so no need to store whole program in our repo. Missing tags in the Imapsync github repo can be problem.

MAGICCC:
> 
> Missing tags in the Imapsync github repo can be problem.

Which indeed is a problem at the moment. I checked my imapsync and got an info that theres a new update available - like with my previous PR

MAGICCC:
As seen on Telegram this morning @andryyy I updated to 2.148

PR #4233, opened 2021-08-15
Update Nextcloud to 22

### LINKED ISSUE #4009: mailcow uses an outdated version of Nextcloud
## Summary

I know Nextcloud isn't the main focus of mailcow, but the mail application in the current Nextcloud releases feels better than SOGo, so I think helper-scripts/nextcloud.sh should use the latest version of the current major release and not just the latest version of Nextcloud 19.
Changing the links from "https://download.nextcloud.com/server/releases/latest-19.tar.bz2" to "https://download.nextcloud.com/server/releases/latest.tar.bz2" would use the latest version of Nextcloud (which is as of now Nextcloud version 21).

## Motivation

Users could benefit from the latest version of Nextcloud without having to update manually.

MAGICCC:
Updates the helper script for new installations to Nextcloud version 22 which is the current stable release as now.
Fixes https://github.com/mailcow/mailcow-dockerized/issues/4009 (Better to pin a version than using latest as seen in this issue)

### DIFF
--- a/helper-scripts/nextcloud.sh
+++ b/helper-scripts/nextcloud.sh
@@ -76,7 +76,7 @@ elif [[ ${NC_UPDATE} == "y" ]]; then
     echo "Cannot upgrade to new major version, please update manually."
     exit 1
   else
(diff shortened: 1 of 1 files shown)


andryyy:
Thanks. :)

PR #4234, opened 2021-08-16
Wording change in update.sh

PromoFaux:
Just something that keeps catching me out when I run `update.sh`, the `read` command waits for a press of the enter key before continuing.


PromoFaux:
🤷

PR #4236, opened 2021-08-17
[imapsync] - parse exit status value and show it human readable in new column

feldsam:


data/web/js/site/user.js
```diff
@@ -224,6 +225,11 @@ jQuery(function($){
             if (!item.last_run > 0) {
               item.last_run = lang.waiting;
             }
+            if (item.success == null) {
+              item.success = '-';
```
andryyy, line 229: Should this be "currently running" or similar?

andryyy: Hm, we have is_running, so that's ok, I guess.

feldsam: Hi, the "item.success" is set to null when a job enters running state or when a job is rescheduled.

feldsam:
@andryyy just updated with parsing exit status functionality. This is much better version for user experience

feldsam:
Screenshot 2021-08-24 at 11 07 51

andryyy:
Staaaaging.

PR #4238, opened 2021-08-19
[web] fido2 add support for M1 macs with Touch ID

feldsam:

According to [this comment](https://github.com/duo-labs/webauthn.io/issues/31#issuecomment-696090864) we need to use attestation 'none' for now.

Related to this [bug](https://bugs.webkit.org/show_bug.cgi?id=224042)


andryyy:
Can you please push to staging? :)

PR #4245, opened 2021-08-28
[Rspamd] Fix SOGo Contacts Dynmap

dragoangel:
Added new normalize_email function which applies to contacts inside wl_by_sogo setting.php dynmap to align with Rspamd:
1. Lowercase emails
2. Remove tag's
3. Remove dots from gmail.com and change googlemail.com to gmail.com with removed dots as per https://github.com/rspamd/rspamd/blob/master/lualib/lua_util.lua#L271-L274

### DIFF
--- a/data/conf/rspamd/dynmaps/settings.php
+++ b/data/conf/rspamd/dynmaps/settings.php
@@ -52,6 +52,7 @@ function parse_email($email) {
 }
 
 function normalize_email($email) {
+  $email = strtolower(str_replace('/', '\/', $email));
   $gm = "@gmail.com";
   if (substr_compare($email, $gm, -strlen($gm)) == 0) {
     $email = explode('@', $email);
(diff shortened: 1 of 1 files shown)


dragoangel:
I also wanted to remove duplicates in resulting from list, but don't know how to implement it properly. @andryyy can you please look in to this? Don't know from performance point what better - bigger output or additional check to minimize this output.

andryyy:
Hm, I don't think it is worth removing the duplicates. Rspamd can handle it itself very fast. It might be slower doing it in PHP.

Thank you. :)

dragoangel:
> Hm, I don't think it is worth removing the duplicates. Rspamd can handle it itself very fast. It might be slower doing it in PHP.
> 
> Thank you. :)

Yeah, this is what I thought as well, danke

PR #4252, opened 2021-08-31
Fixed ofelia container in SELinux environment

feldsam:

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -566,6 +566,8 @@ services:
         - dovecot-mailcow
       labels:
         ofelia.enabled: "true"
+      security_opt:
+        - label=disable
       volumes:
         - /var/run/docker.sock:/var/run/docker.sock:ro
       networks:


andryyy:
@shiz0 Can you confirm this works? :)

feldsam:
I tested it on Oracle Linux 7 with SELinux enabled, kernel UEK R6 and Docker version 19.03.11-ol. It was not working before these changes.

shiz0:
Hey, I am on holiday right now and cannot test it against my environment, but looking at it I doubt it will work like this.

I assume this PR aims to fix an issue with the ofelia-mailcow container being denied access to the docker.sock on SELinux enabled systems?
That would make sense, as SELinux is blocking access to docker.sock for containers not running either "privileged", or with labeling disabled. Few words on what the PR actually aims to achieve would be nice (in the first comment or so).
That's being addressed by setting it to run privileged, but I'd suggest adding
```
      security_opt:
        - label=disable
```
to it, like it has been done for the other containers accessing docker.sock. It would both work, but then the "method of making it accessible" will align across containers.

Please omit the labeling though:
Setting labels to the docker.sock is not something you'd want to do, and especially not "Z".
:Z is only to be used for mounts exclusively being accessed by one single container and will change the labels in the hosts FS accordingly. 
IIRC Docker also will not allow setting any labels to the Socket, but instead will throw an error (at least that's how I can remember it right now).

feldsam:
Hi @shiz0, thank you for your valuable comment. I will try to remove the Z flags and add `security_opt ` as you suggested.

andryyy:
Same here, please: staging, also add text and notes. Your PRs sometimes contain more than their title suggests, you need to change that, please.

PR #4253, opened 2021-09-01
[imapsync] parse, save and show last run status

feldsam:
This PR adds functionality to imapsync runner. It parses and saves result status. Then it is shown in a new column in the sync job table.


andryyy:
That's really cool. :)

PR #4254, opened 2021-09-01
[web] when you adding domain or aliasdomain generate dkim key too

feldsam:
Actual problems:
- two-step work - adding domain and then going to the admin section and generate DKIM key.
- domain admin can add an alias domain too, but it is unable to generate a DKIM key. So customer service needs to be contacted and the DKIM key has to be generated by support.

Both issues were solved in this PR by adding two DKIM related fields to "add" modals and updated permissions checks in `dkim.functions` to allow domain admins to generate DKIM keys.

Other proposals:
- configurable default values of DKIM selector and key size
note: if user-configured default DKIM selector is an empty value, it effectively disable this feature

### DIFF
--- a/data/web/inc/functions.dkim.inc.php
+++ b/data/web/inc/functions.dkim.inc.php
@@ -5,14 +5,6 @@ function dkim($_action, $_data = null, $privkey = false) {
   global $lang;
   switch ($_action) {
     case 'add':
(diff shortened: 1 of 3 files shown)


andryyy:
\o/

andryyy:
A bonus would be to add a toggle to make it optional. :)

PR #4264, opened 2021-09-07
[web] implemented twig templating system

feldsam:
This huge PR implements [Twig templating system](https://twig.symfony.com/doc/3.x/) to Mailcow web UI.

- templates are located in `/web/templates` directory
- templates compile cache is in `/web/templates/cache` directory
  - regarding cache, there are two lines in `docker-entrypoint.sh` for phpfpm container
  - make directory writable and deleting cache files on container restart - so basically on Mailcow updates
- `modals` directory moved to `templates/modals`
- twig is initiated in `inc/twig.inc.php` file along with few functions and filters extensions
- `nav` html element on the index (login) page is hidden by twig logic, so hidding by JS was removed
- original PHP files like `index.php, admin.php, user.php` don't contain HTML code anymore. They act more like "controllers" from MVC patterns
- new `DEV_MODE` env var, by default `n`, switches twig debug mode and exceptions handler
  - when twig debug is on, the cache is auto-recreated on template changes. When debug is false, the cache needs to be deleted after template changes (which is done in `docker-entrypoint.sh`)
- twig by default, auto-escapes all vars, so if we need not escaped var, then we have to use the `raw` filter
  - `raw` I mostly used on lang strings which contains HTML code

Test this branch and post here comment about any bugs. I fix them.


andryyy:
Thank you!!

PR #4266, opened 2021-09-08
Generalize FIDO2 description for biometric auth on Apple devices

accolon:
This is just a cosmetic change to avoid confusion. The recent FIDO2-related additions by @feldsam work with all kinds of biometric authentication on Apple devices, not just Touch ID on newer M1 devices.

I successfully tested with Touch ID on an Intel MacBook Pro, on iPad as well as Face ID on iPhone.

### DIFF
--- a/data/web/lang/lang.de.json
+++ b/data/web/lang/lang.de.json
@@ -646,7 +646,7 @@
         "register_status": "Registrierungsstatus",
         "rename": "Umbenennen",
         "set_fido2": "Registriere FIDO2-Gerät",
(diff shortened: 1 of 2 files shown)


feldsam:
Hi, I know it is working also with face id, but from my testing on iPhone 11, face id was working with original functionality without problems - with attestation. This M1 option should be only temporary, due to a bug in the Apple WebKit. On iPhone, you should use just the original button.

accolon:
Sorry, my mistake. I did not realize that this was a workaround needed JUST for M1 Macs.

Feel free to ignore the PR, then.

PR #4286, opened 2021-10-03
Clean up the xmpp remainder

heavygale:
- delete data/Dockerfiles/ejabberd/*
- remove creation of ZZZ-ejabberd.conf in data/Dockerfiles/phpfpm/docker-entrypoint.sh
- remove xmpp strings from language files


MAGICCC:
Strange that it never got removed..
Can you change the base branch to staging?

Issue #3869, opened 2020-11-23
2FA and application passwords for users


philipszalla:
## Summary

A clear and concise description of what the problem is.
For example: I'm always frustrated when [...]

I would like to secure my mailbox in mailcow with a second factor (e.g. TOTP, U2F, ...).

## Motivation

What are you about to solve or improve with this idea?
What would be the benefit for most users?

This would be a security improvement for users.
Admin already has 2FA

## Additional context

Add any other context or screenshots about the feature request.

Obviously SoGo already has a feature for TOTP.
But this affects only the SoGo login.

It would be nice, when I would be able to add application passwords for my clients (thunderbird, outlook, iOS Mail, Android Mail) and my normal password would only work for SoGo in combination with the TOTP.

I think this needs a custom configuration page for application passwords in SoGo, so the user is able to configure it on it own.

Nextcloud and Gitea have similar approaches.

andryyy:
Duplicate

philipszalla:
@andryyy could you explain this?
> duplicate

I looked into the issues #1348, #3698 and #3736 and none of them discussed a direct implementation of 2FA in mailcow without oauth integration. The issue #740 tried to solve it but is closed now.

andryyy:
Have not checked, enotime, but I will re-open it then. :)

MAGICCC:
Do I understand it right you want app passwords for your clients? If yes, you can create separate accounts when you login with your mailbox user @ the controlpanel and go to the 'App passwords' tab. (SOGo won't work with app passwords)

If you want to have 2FA for IMAP, nope that won't happen since there's no real approach yet to include this to postfix and such.

philipszalla:
@MAGICCC yes you’re right.

But can I still use the regular password for imap and smtp? This should be disabled (because I want to use 2FA or the very strong application passwords)
And can I enable2FA for the controlpanel, too?

andryyy:
We could add WebAuthn/TFA to users, disable IMAP and SMTP for those accounts and put a proxy in front of SOGo to authenticate via key. Only app passwords would work with IMAP and SMTP then. This would exclude access to calendars and address books from external access obviously. Only IMAP and SMTP would still work with app passwords.

andryyy:
@mkuron it's an older topic, but what's your opinion on this?

mkuron:
[...]

Issue #3710, opened 2020-08-18
iOS Profile does not work with App Passwords


pahy:
**Prior to placing the issue, please check following:** *(fill out each checkbox with an `X` once done)*
- [x] I understand, that not following or deleting the below instructions, will result in immediate closing and deletion of my issue.
- [x] I have understood that answers are voluntary and community-driven, and not commercial support.
- [x] I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

**Description of the bug**:

When using app-passwords and installing a profile including mail and cal/carddav on an ios-device, the app-password is only accepted when entered at mail. it is not accepted, when entering at cal/carddav. there it must be the original account-password.

**Docker container logs of affected containers**:

**Reproduction of said bug**:

**System information**:

| Question | Answer |
| --- | --- |
| My operating system | iOS 13.6.1 |
| Is Apparmor, SELinux or similar active? | I_DO_REPLY_HERE |
| Virtualization technlogy (KVM, VMware, Xen, etc - **LXC and OpenVZ are not supported** | I_DO_REPLY_HERE |
| Server/VM specifications (Memory, CPU Cores) | I_DO_REPLY_HERE |
| Docker Version (`docker version`) | I_DO_REPLY_HERE |
| Docker-Compose Version (`docker-compose version`) | I_DO_REPLY_HERE |
| Reverse proxy (custom solution) | I_DO_REPLY_HERE |

- Output of `git diff origin/master`, any other changes to the code? If so, **please post them**.
- All third-party firewalls and custom iptables rules are unsupported. *Please check the Docker docs about how to use Docker with your own ruleset*. Nevertheless, iptabels output can help us to help you: `iptables -L -vn`, `ip6tables -L -vn`, `iptables -L -vn -t nat` and `ip6tables -L -vn -t nat`.
- DNS problems? Please run `docker exec -it $(docker ps -qf name=acme-mailcow) dig +short stackoverflow.com @172.22.1.254` (set the IP accordingly, if you changed the internal mailcow network) and post the output.

andryyy:
![image](https://user-images.githubusercontent.com/2972950/90502001-821ea800-e14d-11ea-9bd0-7f1934938ab3.png)

pahy:
Better as feature request? #3711

mkuron:
[...]

PR #4296, opened 2021-10-24
Support for app-passwords in Cal-, CardDav & ActiveSync

jkellerer:
This PR adds support for app passwords to the missing 3 protocols, Cal-, CardDav & ActiveSync so that devices can be fully managed with app passwords and tools like `vdirsyncer` can be used for continuous calendar+contact backup/sync without a need to use the own user password for that.

There are multiple cases where such a feature was mentioned, e.g.: #3710. When checking the changes with CalDav, CardDav & ActiveSync on Android, iOS, MacOS and Win10, I couldn't find any technical issues, all just worked fine.

Atm there is no enforcement for app passwords as proposed in some cases especially when 2FA is also enabled (e.g. #3869) but it could be added as a configuration option to disallow login with normal passwords in `dovecot` and `sogo-auth.php`. I'm open to extend the PR but I tried to keep it simple at first. UI will also need an update to tell that app passwords are not limited to imap and smtp.

### DIFF
--- a/data/web/autodiscover.php
+++ b/data/web/autodiscover.php
@@ -68,8 +68,7 @@ if (empty($_SERVER['PHP_AUTH_USER']) || empty($_SERVER['PHP_AUTH_PW'])) {
   exit(0);
 }
 
-$allow_app_passwords = $ALLOW_APP_PASSWORDS_IN_EAS === true || $autodiscover_config['autodiscoverType'] == 'imap';
-$login_role = check_login($login_user, $login_pass, $allow_app_passwords);
+$login_role = check_login($login_user, $login_pass, true);
 
 if ($login_role === "user") {
   header("Content-Type: application/xml");
--- a/data/web/inc/vars.inc.php
+++ b/data/web/inc/vars.inc.php
@@ -188,9 +188,6 @@ $MAILBOX_DEFAULT_ATTRIBUTES['mailbox_format'] = 'maildir:';
 // Show last IMAP and POP3 logins
 $SHOW_LAST_LOGIN = true;
 
-// Allow app passwords in CardDav, CalDav and ActiveSync
-$ALLOW_APP_PASSWORDS_IN_EAS = true;
-
 // UV flag handling in FIDO2/WebAuthn - defaults to false to allow iOS logins
 // true = required
 // false = preferred
--- a/data/web/sogo-auth.php
+++ b/data/web/sogo-auth.php
@@ -15,7 +15,7 @@ if (isset($_SERVER['PHP_AUTH_USER'])) {
   $username = $_SERVER['PHP_AUTH_USER'];
   $password = $_SERVER['PHP_AUTH_PW'];
   $is_eas = preg_match('/^(\/SOGo|)\/(dav|Microsoft-Server-ActiveSync).*/', $_SERVER['HTTP_X_ORIGINAL_URI']);
-  $login_check = check_login($username, $password, $is_eas && $ALLOW_APP_PASSWORDS_IN_EAS);
+  $login_check = check_login($username, $password, $is_eas);
   if ($login_check === 'user') {
     header("X-User: $username");
     header("X-Auth: Basic ".base64_encode("$username:$password"));
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -164,7 +164,7 @@ services:
             - phpfpm
 
     sogo-mailcow:
-      image: mailcow/sogo:1.101
+      image: mailcow/sogo:1.102
       environment:
         - DBNAME=${DBNAME}
         - DBUSER=${DBUSER}


mkuron:
Thanks for looking into this. It's a feature I've been wanting myself for a while now. Looking at your code, I was wondering why does this even work? From the SOGo code, I think that what I wrote in https://github.com/mailcow/mailcow-dockerized/issues/3710#issuecomment-676077091 is still correct and it would need a change in the SOGo code. I always thought the sogo-auth.php mechanism was only used when you clicked the "Login to webmail" button in the Mailcow UI. But evidently, that is not true and SOGo makes requests to it every time it authenticates a user -- no matter if it's the webmail UI or EAS or DAV. So your patch is a nice and elegant solution.

Once we have 2FA support for user accounts (I think we only have it for admin accounts right now), we should do the following:

- Block login to IMAP, POP3, SMTP, EAS and DAV with regular passwords when 2FA is enabled. Allow login with app passwords without 2FA validation.
- Block login to SOGo webmail with app passwords.
- Block login to SOGo webmail when 2FA is enabled. Redirect to the main Mailcow page instead.
- Change "App passwords are alternative passwords for your IMAP and SMTP login. The username remains unchanged. SOGo (including ActiveSync) is not available through app passwords." to "App passwords are alternative passwords for your IMAP, SMTP, CalDAV, CardDAV and EAS login. The username remains unchanged. SOGo webmail is not available through app passwords. You cannot use your regular password with these services when 2FA is enabled." We also need translations of this message in a bunch of other languages.


Before this pull request can be merged, please do the following, @jkellerer:
- Block login to SOGo webmail with app passwords. I think you can detect that from the `X-Original-URI` header in sogo-auth.php. SOGo should only allow app passwords for EAS and DAV.
- Replace "App passwords are alternative passwords for your IMAP and SMTP login. The username remains unchanged. SOGo (including ActiveSync) is not available through app passwords." with "App passwords are alternative passwords for your IMAP, SMTP, CalDAV, CardDAV and EAS login. The username remains unchanged. SOGo webmail is not available through app passwords.". We also need translations of this message in a bunch of other languages.

jkellerer:
The reason that it works is that proxy authentication (`sogo-auth.php`) is used for php-session based SSO and basic HTTP auth (EAS, DAV). SOGo uses the "`Authorization`" header returned by `sogo-auth.php` to log into `dovecot` ([SoObjects/SOGo/SOGoProxyAuthenticator.m#L149](https://github.com/inverse-inc/sogo/blob/c07aeeffec48de28ccb69d222b38ea2d23f57b6a/SoObjects/SOGo/SOGoProxyAuthenticator.m#L149)) and `dovecot` accepts the original app-password. For the WebDav part, the password is only used for auth type "cas" (`SOGoAuthenticationType`) but this isn't configured so setting "`x-webobjects-remote-user`" is sufficient to have access.

andryyy:
- https://github.com/mailcow/mailcow-dockerized/blob/2a4764aa4164f92865e206e47f97f3f4baaffb8f/data/web/inc/functions.inc.php#L891 This should probably be a better join instead of a second query. Not a _must_. :)

- Last login via app password is not covered, _I think_.
  It is also not logged on SSO via mailcow UI, that should be included in this change.
  I can add this to this branch.

- The "app password for EAS" stuff should be removed imo and be the default. @mkuron?

- I will add TFA for users (this branch)

- With the last merge from staging today I allowed SSO to SOGo even with "SOGo access" disabled but mentioned this change in the UI. It is now a toggle to allow "direct access" instead of "access" in general. Is that something that we may need to rethink  in this case, @mkuron?

Thank you @jkellerer for this PR! :)

andryyy:
Can we also add how IMAP and SMTP protocol access restrictions work with SSO/proxy auth? I think they are covered. Iirc Dovecot sets the  real user to the proxied user via master.

mkuron:
> The "app password for EAS" stuff should be removed imo and be the default.

sure, that‘s fine too, as long as we keep the "SOGo webmail is not available through app passwords."

> With the last merge from staging today I allowed SSO to SOGo even with "SOGo access" disabled but mentioned this change in the UI. It is now a toggle to allow "direct access" instead of "access" in general. Is that something that we may need to rethink in this case

depends on why people would disable SOGo access. Is it because they want to split their users into ones that are worthy of the extra resources expended for SOGo and ones that are not? Is it for some vague security reason? To cover all potential cases, it would need to be a three-way toggle (disabled/indirect only/direct and indirect).

jkellerer:
Hi @andryyy, @mkuron,

* Access restriction on IMAP prevents ActiveSync. When access is not restricted, SASL log is populated with the internal SOGo address and due to the way it works, the app password is also logged for IMAP.
* `$ALLOW_APP_PASSWORDS_IN_EAS` was added only as I wasn't sure whether it was intention or technical limitation that this feature wasn't added. So I'm absolutely fine to remove it.
* Last login is not explicitly recorded but it appears in sasl log when SOGo performs a login into dovecot. Additional log might be a good idea but was unsure where to add that.
* SQL: Thought about this as well but considered it easier to read and review when done as separate call (code duplication is minimal).

Finally, I did find an issue that has to be resolved before the PR can be merged. Sending mail via ActiveSync doesn't work and when checking SOGo sources, I'm relatively sure it is caused by the fact that `_sendMail` in `ActiveSync/SOGoActiveSyncDispatcher.m` is the only location where the authenticator is not selected by method `authenticatorInContext`. Instead it is hard-coded and this means proxy auth doesn't work in this case. 

Originally I derived the changes from my experiments to enable EAS with LDAP, where this works fine when "plist_ldap" is also defined since that hides the issue with authenticator when the PW is also in LDAP. Looks like I missed this part in my tests for app-passwords.

The changes required in SOGo should be minimal (only a single line) but I'm not sure how quick this can be added. I'll try to verify this with a local build asap.

andryyy:
@mkuron

> > The "app password for EAS" stuff should be removed imo and be the default.
> 
> sure, that‘s fine too, as long as we keep the "SOGo webmail is not available through app passwords."
> 

Or we make it configurable per app password, that would be great! I missed the point that an app password might only be able to be used for DAV.

> > With the last merge from staging today I allowed SSO to SOGo even with "SOGo access" disabled but mentioned this change in the UI. It is now a toggle to allow "direct access" instead of "access" in general. Is that something that we may need to rethink in this case
> 
> depends on why people would disable SOGo access. Is it because they want to split their users into ones that are worthy of the extra resources expended for SOGo and ones that are not? Is it for some vague security reason? To cover all potential cases, it would need to be a three-way toggle (disabled/indirect only/direct and indirect).

Hmm. I see. No matter what we decide to change: Do you think it should default to "no direct login" when TFA is enabled for a user?

@jkellerer 

Thanks for your findings and testings! That's amazing. :) Very happy you put so much effort into it.

Perhaps we can trigger @inverse-inc, @cgx and @extrafu for a quick chat on that change in `ActiveSync/SOGoActiveSyncDispatcher.m`?

mkuron:
> Or we make it configurable per app password, that would be great! I missed the point that an app password might only be able to be used for DAV.

Great idea. For full flexibility, we could have individual options per app password for all services: IMAP/POP3/SMTP, EAS, DAV, SOGo UI. SOGo UI should be deselected by default though.

> Do you think it should default to "no direct login" when TFA is enabled for a user?

yes. When 2FA is enabled, there should be no path that lets you log into anything without 2FA or an app password. That shouldn’t even be configurable.

jkellerer:
Btw. small update, I've managed to fix SOGo and it's really only a single line of code change. Will create a PR later that day.

andryyy:
Nice, it was merged. Will wait for a new build and start to work on the other changes then. Can you merge this to the branch app-passwd-daveas?

andryyy:
I will need to add a POP3 toggle and perhaps sieve... forgot these. But so far it seems to be working fine.

jkellerer:
Hi @andryyy , just gave the staging branch a quick test and it really looks and works great. 

Just one question: Currently the EAS permission also requires SMTP and IMAP to be enabled to work, which may not be what a user expects or wants. EAS on the other hand produces requests from IP `${IPV4_NETWORK:-172.22.1}.248` (e.g. `29.10.2021, 12:33:21 IMAP from Web/EAS/Internal (172.22.1.248)`). Would it make sense to allow IMAP and SMTP from this IP when EAS is allowed?

andryyy:
Hi,
ah, damn. That's a good catch. Hum. SOGo has a static IPv4 .248, we could concat the env var IPV4_NETWORK with 248 and compare with real_ip in LUA.
SOGo always uses IPv4, so that'd work.
Another open task is modifying the language file when editing a mailbox. Toggling protocols should have a new info text like "disabling protocols for direct access without app password" or something like that. You know what I mean? I am in a hurry and doing 200 things at the same time atm, sorry.

andryyy:
Ah, there might also be a "master" variable available in LUA which could be used to determine the login. Or both? We can also compare the master user value. Or everything.

jkellerer:
Master user value would probably not work since for proxy auth, user and password is like entered in the client (it is passed through). Only static IP would work (using IPV4_NETWORK with 248).

jkellerer:
Good point also that mailbox restrictions do not apply to app passwords. Yes this needs an update in the L10N files.

jkellerer:
Just checked the lua code, real-ip is available via `con:escape(req.real_rip)` and since it is created in shell script `${IPV4_NETWORK:-172.22.1}.248` should evaluate too. Looks like this isn't really complicated. If you want I give it a try and create another PR later.

andryyy:
I can check it later, too, no worries. :)

andryyy:
Can you test the staging branch again? :)

jkellerer:
Found one extra thing, in `sogo-auth-php` and `functions.inc.php` the table `sasl_log` is now filled with DAV and EAS entries. When they are logged the remote IP is set to `127.0.0.1`:
```
30.10.2021, 11:31:25 EAS von Web/EAS/Internal (127.0.0.1)
```
The reason for this is that Nginx is the caller so `$_SERVER['REMOTE_ADDR']` is set to localhost. Can be fixed easily by using `($_SERVER['HTTP_X_REAL_IP'] ?? $_SERVER['REMOTE_ADDR'])` at: 
https://github.com/mailcow/mailcow-dockerized/blob/644b1f85d1b5ff43425c503c56554ac1c8d1d0a9/data/web/sogo-auth.php#L60
https://github.com/mailcow/mailcow-dockerized/blob/644b1f85d1b5ff43425c503c56554ac1c8d1d0a9/data/web/inc/functions.inc.php#L947

jkellerer:
There is a typo in the lua script to check app-passwords `req.real_ip` instead of `req.real_rip`:
https://github.com/mailcow/mailcow-dockerized/blob/644b1f85d1b5ff43425c503c56554ac1c8d1d0a9/data/Dockerfiles/dovecot/docker-entrypoint.sh#L180

Once fixed the changes work fine

jkellerer:
Btw. would it make sense to skip `sasl_log` for app passwords in lua when the `req.real_rip` is `__IPV4_SOGO__`? 

For example for EAS access I get 2-3 entries:
```
30.10.2021, 11:57:44 IMAP von Web/EAS/Internal (172.22.1.248)
30.10.2021, 11:57:44 SMTP von Web/EAS/Internal (172.22.1.248)
30.10.2021, 11:57:44 EAS von Web/EAS/Internal (192.168.22.11)
```

andryyy:
Thanks! Fixed all. :) I agree sasl_log should not trigger on IPv4_SOGo.

ciroiriarte:
This is  awesome. 

If I have to move to external authentication (SAML2/OIDC middleman in front of a legacy AD server), will the app password logic still apply?

Issue #4298, opened 2021-10-26
Mailcow Watchdog not sending notifications


Spartex:
Mailcow-Watchdog is stating that it is sending notifications in the logs but no notification is being received via email. No errors are shown in the server logs either. The mailcow.conf USE_WATCHDOG=y so not sure why this is happening?

MAGICCC:
Can you update to the lastest version of mailcow and set WATCHDOG_VERBOSE=y in your mailcow.conf?
Then you can restart the watchdog-mailcow container and check the logs.

Spartex:
Hello,

Thank you for your reply and time.
I have updated the image and now in the UI logs it is stating " Error while
sending a notification email to [email address]"

Now I can see with the verbose logs enabled in mailcow.conf on the server when running docker-compose logs --tail=200 watchdog-mailcow that the issue is caused by Microsoft blacklisting the server IP. Thank you, this was very helpful in narrowing down the cause :)

Issue #4197, opened 2021-07-24
watchdog isn't sending emails but says it is


GrantBirki:
**Prior to placing the issue, please check following:** *(fill out each checkbox with an `X` once done)*
- [x] I understand that not following or deleting the below instructions will result in immediate closure and/or deletion of my issue.
- [x] I have understood that this bug report is dedicated for bugs, and not for support-related inquiries.
- [x] I have understood that answers are voluntary and community-driven, and not commercial support.
- [x] I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

## Summary

Watchdog claims that it "sends" and email alert. However, the email is not actually being sent. I believe this is due to Azure blocking outgoing SMTP traffic over port 25.

## Logs

**mailcow.conf snippet:**

```
# Enable watchdog (watchdog-mailcow) to restart unhealthy containers

USE_WATCHDOG=y

# Send watchdog notifications by mail (sent from watchdog@MAILCOW_HOSTNAME)
# CAUTION:
# 1. You should use external recipients
# 2. Mails are sent unsigned (no DKIM)
# 3. If you use DMARC, create a separate DMARC policy ("v=DMARC1; p=none;" in _dmarc.MAILCOW_HOSTNAME)
# Multiple rcpts allowed, NO quotation marks, NO spaces

#WATCHDOG_NOTIFY_EMAIL=a@example.com,b@example.com,c@example.com
WATCHDOG_NOTIFY_EMAIL=<redacted>
```

```console
# docker-compose logs --tail=200 -f watchdog-mailcow 
Attaching to mailcowdockerized_watchdog-mailcow_1
watchdog-mailcow_1   | Waiting for containers to settle...
watchdog-mailcow_1   | Uptime: 25  Threads: 7  Questions: 356  Slow queries: 0  Opens: 51  Open tables: 42  Queries per second avg: 14.240
watchdog-mailcow_1   | Resolved MX for <redacted>: gmail-smtp-in.l.google.com
watchdog-mailcow_1   | Sat Jul 24 19:59:55 UTC 2021 Sent notification email to <redacted>
watchdog-mailcow_1   | Resolved MX for <redacted>: gmail-smtp-in.l.google.com
watchdog-mailcow_1   | Sat Jul 24 20:00:00 UTC 2021 Sent notification email to <redacted>
```

## Reproduction

[...]

PR #4299, opened 2021-10-26
Add Watchdog verbose logging

MAGICCC:
Adding the possibility to see verbose logs while sending mails via Watchdog
May fix https://github.com/mailcow/mailcow-dockerized/issues/4298 + https://github.com/mailcow/mailcow-dockerized/issues/4197

PS: Needs a new watchdog image then
PPS: Needs more testing - only tested it quickly. @andryyy maybe you have some time for this

### DIFF
--- a/data/Dockerfiles/watchdog/watchdog.sh
+++ b/data/Dockerfiles/watchdog/watchdog.sh
@@ -125,14 +125,13 @@ function mail_error() {
   IFS=',' read -r -a MAIL_RCPTS <<< "${WATCHDOG_NOTIFY_EMAIL}"
   for rcpt in "${MAIL_RCPTS[@]}"; do
     RCPT_DOMAIN=
-    #RCPT_MX=
+    RCPT_MX=
     RCPT_DOMAIN=$(echo ${rcpt} | awk -F @ {'print $NF'})
-    # Latest smtp-cli looks up mx via dns
-    #RCPT_MX=$(dig +short ${RCPT_DOMAIN} mx | sort -n | awk '{print $2; exit}')
-    #if [[ -z ${RCPT_MX} ]]; then
-    #  log_msg "Cannot determine MX for ${rcpt}, skipping email notification..."
-    #  return 1
-    #fi
+    CHECK_FOR_VALID_MX=$(dig +short ${RCPT_DOMAIN} mx)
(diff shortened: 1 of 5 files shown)


data/Dockerfiles/watchdog/watchdog.sh
```diff
@@ -133,7 +140,11 @@ function mail_error() {
       --hello-host=${MAILCOW_HOSTNAME} \
       --ipv4
       #--server="${RCPT_MX}"
-    log_msg "Sent notification email to ${rcpt}"
+    if [[ $? -eq 1 ]]; then # exit code 1 is fine
```
andryyy, line 146: Looks good in my test, but why? Did you find an issue regarding this? Is it normal?

andryyy:
Nice. :) Let me know when you can merge. I will build the image then first.

data/Dockerfiles/watchdog/watchdog.sh
MAGICCC: Looks like it. I found https://github.com/mludvig/smtp-cli/pull/17 and seems exitcode is 1 when everything is fine...

MAGICCC:
I just tested a bit and looks fine on my side, but I noticed when you set a mail with a bad domain (no mx record), then no mail will be sent.
Maybe we can reuse this: https://github.com/mailcow/mailcow-dockerized/blob/8b4e1ce70f4724243bf748a4ad231e18ef2732f2/data/Dockerfiles/watchdog/watchdog.sh#L129-L135

like this:
```
RCPT_DOMAIN=$(echo ${rcpt} | awk -F @ {'print $NF'})
CHECK_FOR_VALID_MX=$(dig +short ${RCPT_DOMAIN} mx | sort -n | awk '{print $2; exit}')
if [[ -z ${CHECK_FOR_VALID_MX} ]]; then
  log_msg "Cannot determine MX for ${rcpt}, skipping email notification..."
  return 1
fi
```

andryyy:
`CHECK_FOR_VALID_MX=$(dig +short ${RCPT_DOMAIN} mx)` would be enough. :)

Can you adjust the PR?

andryyy:
Thank youuu

andryyy:
We should probably add the variable information to the docs.

dragoangel:
Wow, nice :)

PR #4306, opened 2021-10-31
[Update] Add variable for watchdog verbose output to mailcow config

MaxXor:
Currently the new variable is not added to the mailcow.conf when updating a mailcow instance.

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -306,6 +306,7 @@ CONFIG_ARRAY=(
   "MAILCOW_PASS_SCHEME"
   "ADDITIONAL_SERVER_NAMES"
   "ACME_CONTACT"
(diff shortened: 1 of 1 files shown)


andryyy:
Not every variable is in the mailcow.conf

That's one of those that is documented and not meant to be there. :)

MaxXor:
I'm confused, [why is it then added for new mailcow instances](https://github.com/mailcow/mailcow-dockerized/blob/master/generate_config.sh#L290)?

andryyy:
Indeed. It’s a valid point then. Thanks.

MAGICCC:
Thanks, seems I forgot the line :)

PR #4309, opened 2021-11-01
nextcloud.conf: add_header Strict-Transport-Security

sujiba:
Hi,

i have seen that in nextcloud.conf hsts is missing. Maybe someone can add it.

https://ssl-config.mozilla.org/#server=nginx&version=1.17.7&config=intermediate&openssl=1.1.1k&guideline=5.6


andryyy:
HSTS is something people must decide on their own. Some headers MUST be considered on their own. Copy/paste some security headers without considering the impact can be fatal. :/ That was already discussed a while ago.

sujiba:
Oh. I'm sorry, i must have overlooked that. I close the request.

Issue #4281, opened 2021-09-30
API endpoint to return all app password objects


strifel:
## Summary

I would like to have an API endpoint that returns all app password objects for all users and not only for a specific user.
(Maybe even with a only created after filter option)

## Motivation

We want to automatically delete some app passwords after a specified time.
Having this endpoint would allow us to do this without having to do a lot of requests. (We might at some point have > 1000 mailboxes)

PR #4311, opened 2021-11-01
Add app-passwd list (#4281)

strifel:
This PR adds the option to list all app passwords.
The issue #4281  details the use case further.

I think the might be a better option for the check `$_data['username'] != null`, but I do not really know how to.

Thanks for reviewing the PR

### DIFF
--- a/data/web/api/openapi.yaml
+++ b/data/web/api/openapi.yaml
@@ -3520,6 +3520,39 @@ paths:
       description: You can list mailbox aliases existing in system.
       operationId: Get aliases
       summary: Get aliases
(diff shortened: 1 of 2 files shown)


andryyy:
Why an incomplete and broken PR instead of an issue?

strifel:
I am sorry to have caused any trouble. I opened #4281 a few weeks ago, but because we needed this implemented quite soon I implemented the changes you found here in our production installation without any problems.
Because I wanted to have others potentially benefit from this changes, too, I created this PR as a start and am/was open to get any feedback on what should be changed prior to merge.

PR #4314, opened 2021-11-05
[ClamAV] Update to 0.103.4

svengo:
ClamAV 0.103.4 is a critical patch release, see https://blog.clamav.net/2021/11/clamav-01034-and-01041-patch-releases.html for more information.


MAGICCC:
Would also need a new container to be built then:
https://github.com/mailcow/mailcow-dockerized/blob/master/docker-compose.yml#L61
```diff
-image: mailcow/clamd:1.41
+image: mailcow/clamd:1.42
```

PR #4316, opened 2021-11-09
Auto-generated app passwords for Apple configuration profiles

mkuron:
This pull request adds embedded app passwords to our Apple configuration profiles generator, available via a separate download link on the mailbox page. It builds upon #4296's app password support for CardDAV/CalDAV. The app password is randomly generated when the link is clicked. The app name is set to _iPhone_/_iPad_/_Mac_ depending on the HTTP user agent, and the protocols are restricted to what is needed. I have been testing it for a few days and it works fine. It is a relatively trivial change, but very convenient: you now only have to enter your password once at the mailcow web UI, and not three more times in Apple's settings app.

And while I was at it, I also added a warning to the change password modal that shows up when the account also has app passwords. When someone is changing their password for security reasons, they may also want to check for unexpected app passwords.


andryyy:
That's nice, thank you! :)

PR #4320, opened 2021-11-12
Add missing API endpoint to openapi.yaml

### LINKED ISSUE #3889: No documentation for time limited aliases API calls
## Summary

There is no documentation in the OpenAPI YAML about `time_limited_alias`. Because of that, I thought there wasn't an API at all but investigating the HTML did reveal there is one.

## Motivation

I was hoping to build a Firefox extension that can generate and inject a temporary email into a form field from the context menu. Making it super easy to use the feature.

MrKrisKrisu:
closes #3889

### DIFF
--- a/data/web/api/openapi.yaml
+++ b/data/web/api/openapi.yaml
@@ -131,6 +131,69 @@ paths:
                   type: boolean
               type: object
       summary: Create alias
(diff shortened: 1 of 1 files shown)


ntimo:
Thanks :)

PR #4326, opened 2021-11-16
[Assets] Beautify quarantine template

the-voidl:
I added explanations and examples to the quarantine notification using parts of bootstrap's CSS.
Please verify that the strings in my examples like "reject(ed?)" are correct since I don't currently have any mails in quarantine.

Also would I like to extend this template by translations. Is it possible to simply include strings e.g. by `{{ lang.rejected }}` in the `quarantine.tpl` file?


waja:
ping

PR #4356, opened 2021-12-03
Better Random Passwords

dioncodes:
Replaced unnecessarily "pronounceable" short password generation with more secure random passwords (default length of 14 characters containing at least 2 letters and 2 special characters)


andryyy:
Why not open an issue? I prefer the pronounceable passwords.

dioncodes:
> Why not open an issue? I prefer the pronounceable passwords. 

To be honest these are barely actually pronouncable, too short (unsafe) and by the many special characters really uncomfortable to type on the keyboard and especially mobile devices. If you tell someone the password you still need to spell it character by character and due to all these special characters it doesn't make it much easier to remember them, so where's the advantage? 

But of course another option would be to offer both password types when selecting to generate a password.

gnasch:
Please do not do this, this is actually really bad for security! Those punished with unpronouncable passwords will use any crutch to get around them. If you think, the pronouncable passwords are too short - just make them longer. Does mailcow have to become user-unfriendly?

waja:
Since there seems no common sense about it, how about make this configurable?

ghost:
My comment won't help but personally I like the Apple format very much: `doptaD-jyjqez-5jyncy`.

Almost pronounceable, sufficiently long and easy to type.

dioncodes:
@DerLinkman Is a similar enhancement already implemented in the main repo by now, or why did you close this?

DerLinkman:
Sorry, it was´nt my intention :/

DerLinkman:
We could think about a changeable mailcow.conf variable which includes several types like pronnouncable (default), randompasswords and maybe apple like.

It´s at least an idea. If it will implemented or not is not clear.

milkmaker:

PR #4359, opened 2021-12-03
Typo lang.de.json

roswitina:


MAGICCC:
Thanks, you can also suggest translations here, https://translate.mailcow.email/engage/mailcow-dockerized/

PR #4361, opened 2021-12-05
Add a note about creating a backup before starting over

TomK32:
I was stupid/tired enough to copy and run that `docker-compose down -v`. Hopefully this little change will prevent others from loosing their data as well.
Btw, I did get that page after my disk space was running out.


DerLinkman (commented):
Looks good :)
Will merged in the staging branch and soon™ to the master branch.

PR #4363, opened 2021-12-08
[WEB] fixed wrong value entry in Pushover.twig

DerLinkman:
pushover_data.token was used twice in Pushover.twig (instead of pushover_data.senders)


feldsam:
Thanks!

PR #4381, opened 2021-12-13
[ofelia] - ofelia-mailcow does not have the correct time zone

christianbur:
ofelia-mailcow does not have the correct time zone.
Test: docker exec -it mailcowdockerized_ofelia-mailcow_1 date


MAGICCC:
indeed thanks

PR #4385, opened 2021-12-15
Fix Apple Config: added CalDAVPassword and CardDAVPassword which was missing

holzhannes:
If the user is choosing the config with app password, it is now also used for CalDav and CardDav.


mkuron:
Thanks, I totally forgot about DAV when I did #4316.

saibotk:
@mkuron Seems like this was merged to master but not staging. So it currently is missing in staging, maybe we should merge master back into staging?

PR #4393, opened 2021-12-20
Nextcloud helper

Fly7113:
These two commits are aimed at:

1. Bumping up Nextcloud version to 23, since it's now the stable channel;
2. Fixing a bug in the purge mode of the script, caused by #4043 not customising anymore the tables' prefix (more details in the [commit](https://github.com/mailcow/mailcow-dockerized/commit/932469a4ed32342b653baea936643deb9e8112f2) message).

### DIFF
--- a/helper-scripts/nextcloud.sh
+++ b/helper-scripts/nextcloud.sh
@@ -76,7 +76,7 @@ elif [[ ${NC_UPDATE} == "y" ]]; then
     echo "Cannot upgrade to new major version, please update manually."
     exit 1
   else
-    curl -L# -o nextcloud.tar.bz2 "https://download.nextcloud.com/server/releases/latest-22.tar.bz2" || { echo "Failed to download Nextcloud archive."; exit 1; } \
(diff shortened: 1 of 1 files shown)


Lulalaby (approved):

MAGICCC:
I am not into MySQL commands but maybe we should also include something like `TABLE_NAME LIKE 'oc_%'  OR TABLE_NAME LIKE 'nc_%'` then people using nc_ prefixes can still purge their Nextcloud

Fly7113:
> I am not into MySQL commands but maybe we should also include something like `TABLE_NAME LIKE 'oc_%' OR TABLE_NAME LIKE 'nc_%'` then people using nc_ prefixes can still purge their Nextcloud

It's either this or dropping completely the support for old Nextcloud instances, though adding a few lines in the documentation mentioning this discrepancy between old (<20.0.0) and new (>=20.0.0) versions of Nextcloud, with a snippet to run to either modify the script for nc_ prefixes or correctly purge the database (IMHO the second option is the best: this way the script will keep working for other installs/purges).

milkmaker:

PR #4403, opened 2021-12-26
Fix "The operation is insecure." when trying to register fido2 device.

El-Virus:
navigator.credentials.create(); Doesn't accept a port in the "id" parameter (generated by https://domain.tld:port/api/v1/get/fido2-registration/user). So, when trying to register a fido2 device via WebAuthn throws: "The operation is insecure." on firefox and "The relying party ID is not a registrable domain suffix of, nor equal to the current domain." on Chrome or Edge.
This commit replaces `$_SERVER['HTTP_HOST']` with `$_SERVER['SERVER_NAME']` when initializing `$WebAuthn` which excludes the port to formulate correct requests.
Now Mailcow allows the registration of fido2 devices when running in a non-standard port(eg. 443).


DerLinkman:
Was not intentional (with the staging branch delete)

DerLinkman:
@FreddleSpl0it opinion on that? Still needed after the WebAuthn Update?

FreddleSpl0it:
Still needed.  $_SERVER['HTTP_HOST'] should be changed to $_SERVER['SERVER_NAME'] when initializing lbuchs\WebAuthn library

El-Virus:
@DerLinkman, was it closed accidentally again?

DerLinkman:
Yes... thanks to GitHub´s auto delete branch option. Deactivated it...

Sorry

Issue #4421, opened 2022-01-14
Deprecation of U2F API - Implementation of WebAuthn as 2FA-Provider


itshappyli:
## Summary

WebAuthn has been introduced as a new standard for 2FA, replacing the legacy U2F standard. Additionally, the WebAuthn standard is backward compatible with U2F devices. Unfortunately, the Google Chrome browser only supports U2F until February 2022. The same applies for the Microsoft Edge browser respectively.

## Motivation

Since many organizations rely on hardware security keys in their IT infrastructure, support for authentication with these keys must be maintained and guaranteed.

## Additional context

Also explained in this blog entry from the Chrome Developers:
[https://developer.chrome.com/blog/deps-rems-95/](https://developer.chrome.com/blog/deps-rems-95/)

As far as I know, there is already an implementation of WebAuthn in the mailcow UI. The current implementation allows hardware based login to the UI (hardware must also fully comply with the FIDO2 standard). My point of view is using WebAuthn as 2FA method as described above.

FreddleSpl0it:
I've already worked on that and have successfully migrated from u2f to webauthn https://github.com/FreddleSpl0it/mailcow-dockerized. I will try to do a PR next week.

itshappyli:
Thanks for your contribution! It's really nice to hear that you managed to successfully implement the WebAuthn API.

I can't wait until it's available!

DerLinkman:
Yes and we will merge it into the Janmooary Update :)

ghost:
Has anyone tested this new implementation under linux? Can't register a yubikey 5 NFC, i successfully used with the old U2F implementation. Tested under Ubuntu with Firefox 96 & Chromium 97.

(This yubikey demo works: https://demo.yubico.com/webauthn-technical/registration.)

DerLinkman:
Can you give us a error code?

FreddleSpl0it:
does your mailcow have a valid ssl certificate or does it have a self-signed/untrusted one?

ghost:
> Can you give us a error code?

![Bildschirmfoto-20220122211747-2190x214](https://user-images.githubusercontent.com/9143480/150654171-bf7da00b-c96d-4ec1-954f-58a46fb9c97c.png)

There is just a timeout. Normally the yubikey would blink, during the firefox popup, and accept a touch. But in this case it doesn't.

> does your mailcow have a valid ssl certificate or does it have a self-signed/untrusted one?

Yes, a valid ssl certificate.

[...]

PR #4428, opened 2022-01-18
Migrating from U2F to WebAuthn for 2FA

FreddleSpl0it:
### Description
The Google Chrome browser only supports the legacy U2F standard until February 2022 and therefore the current 2FA implementation will no longer work. Migrating to the new WebAuthn standard will solve this issue.

### This PR closes the following Issues
Closes [#4421](https://github.com/mailcow/mailcow-dockerized/issues/4421)
fixes some issues on android with the current fido2 login


DerLinkman:
Looks good so far. Before we merge I would like to get a review from @mkuron as well as another Live Test (a mailcow paid support customer was asking about it).

mkuron:
I have looked at it a few days ago and it looks reasonable. However, I am not an expert on Webauthn at all, so we should definitely get another review from someone who is actually familiar with it.

DerLinkman:
As i mentioned earlier tomorrow (hopefully) i will test this patch on a running machine.
Android and Apple WebAuthn works as expected (or even better) with the new Version.

DerLinkman:
Test failed! It did´nt worked. This key was used:

https://www.token2.swiss/shop/product/token2-t2f2-alu-fido2-u2f-and-totp-security-key

FreddleSpl0it:
I have disabled rootCA validation by default because of this comment [lbuchs/WebAuthn/issues/31.](https://github.com/lbuchs/WebAuthn/issues/31#issuecomment-763597036) This function can still be enabled via mailcow.conf.

generate_config.sh
```diff
@@ -344,22 +344,10 @@ DOVECOT_MASTER_PASS=
...
-#   googleHardware certified
-#   microsoftTpmCollection certified
-#   huawei certified
-#   trustkey certified
-#   bsi certified
-WEBAUTHN_DISABLE_ROOTCA=n
+# Enable webauthn device manufacturer verification
+# After setting WEBAUTHN_RESPECT_ROOTCA=y only devices from trusted manufacturers are allowed
+# root certificates can be placed for validation under mailcow-dockerized/data/web/inc/lib/WebAuthn/rootCertificates
+WEBAUTHN_RESPECT_ROOTCA=n
```
andryyy, line 350: Automatically enabled?

generate_config.sh
```diff
@@ -344,22 +344,10 @@ DOVECOT_MASTER_PASS=
...
-#   globalSign certified
-#   googleHardware certified
-#   microsoftTpmCollection certified
-#   huawei certified
-#   trustkey certified
-#   bsi certified
-WEBAUTHN_DISABLE_ROOTCA=n
+# Enable webauthn device manufacturer verification
+# After setting WEBAUTHN_RESPECT_ROOTCA=y only devices from trusted manufacturers are allowed
+# root certificates can be placed for validation under mailcow-dockerized/data/web/inc/lib/WebAuthn/rootCertificates
```
andryyy, line 349: Automatically enabled?!

update.sh
```diff
@@ -307,7 +307,7 @@ CONFIG_ARRAY=(
   "ADDITIONAL_SERVER_NAMES"
   "ACME_CONTACT"
   "WATCHDOG_VERBOSE"
-  "WEBAUTHN_DISABLE_ROOTCA"
+  "WEBAUTHN_RESPECT_ROOTCA"
```
andryyy: Are there more inconsistencies of respect and disable?

andryyy:
Questions might be already answered. Pretty low on time. Only on mobile, too. Sorry.

andryyy:
We allow for key only logins. Without passwords. If we do this with untrusted root CAs that’s pretty crazy.

andryyy:
Add a link to the docs about how to allow a new CA, but please don’t trust it. I think that’s more than crazy when using 🔑-only logins. :/

mkuron:
Actually, I think the correct thing to do may be to ignore the root CA. It is only needed to determine if a device is from a specific vendor. I don‘t think it‘s part of the public-key authentication. See https://developers.yubico.com/U2F/Attestation_and_Metadata/ and https://developers.yubico.com/WebAuthn/Concepts/Securing_WebAuthn_with_Attestation.html for what Yubico says about this.

FreddleSpl0it:
I can undo the changes. It was an idea that came to me while reading this thread https://github.com/lbuchs/WebAuthn/issues/31#issuecomment-780401348, since the author of the WebAuthn library himself says it is not necessary

DerLinkman:
As i mentioned earlier i tested this patch on a customer machine which had a Third Party FIDO Key (https://www.token2.swiss/shop/product/token2-t2f2-alu-fido2-u2f-and-totp-security-key) registered before as U2F. But as soon he wanted to register the same Key with WebAuthn he wasn´t able to because "Error invalid root Certificate - (reload browser if the error persists)" (and yes on a different browser, the same problem), which is the exact problem i guess.

andryyy:
My concern was yet to limit WebAuthn to trusted devices. Not really about the usefulness of the patch itself.

andryyy:
Thanks for all the input, guys. 🙏

DerLinkman:
That's understandable 😊
I only want to say that the original planned version (so respecting the Root CAs) probably won't work for users which don't have a Key from a manufacturer included into the CAs.
________________________________

DerLinkman:
Ok the new tests claims that it is working successfully (same customer and same key as yesterday)

The new Variable in the mailcow.conf is checking if the Webauthn should ignore the certificates or not.

I already started working on the documentation: 
https://github.com/DerLinkman/mailcow-dockerized-docs/blob/WebAuthn-TFA-Documentation/docs/u_e-mailcow_ui-tfa.md

So @andryyy and or @mkuron what do you think about it?

andryyy:
Hey @FreddleSpl0it, sorry for being a jerk, just figured something out internally.

Thanks for your contribution and help. I got something completely wrong.

Issue #4401, opened 2021-12-25
Olefy 1.8 broken 


cyrinux:
## Sumarry
Hi, since the olefy update to image 1.8, olefy don't start anymore with the error below.
After a downgrade to 1.7 image this is OK.

It's maybe related to this ? https://github.com/HeinleinSupport/olefy/commit/fee073b1ee44f014ace848ed3d31a972a7c61baf
https://github.com/HeinleinSupport/olefy/pull/14
## Logs

```
olefy-mailcow_1      | olefy ERROR default_exception_handler Fatal error: protocol.eof_received() call failed.
olefy-mailcow_1      | protocol: <__main__.AIO object at 0x7f2295529910>
olefy-mailcow_1      | transport: <_SelectorSocketTransport fd=7 read=polling write=<idle, bufsize=0>>
olefy-mailcow_1      | Traceback (most recent call last):
olefy-mailcow_1      |   File "/usr/lib/python3.9/asyncio/selector_events.py", line 882, in _read_ready__on_eof
olefy-mailcow_1      |     keep_open = self._protocol.eof_received()
olefy-mailcow_1      |   File "/app/olefy.py", line 171, in eof_received
olefy-mailcow_1      |     rspamd_id = olefy_headers['Rspamd-ID'][:6] or ''
olefy-mailcow_1      | KeyError: 'Rspamd-ID'
```

copart:
Any known side effects of downgrading to mailcow/olefy:1.7 and leaving everything else the same in docker-compose?

DerLinkman:
Will fixed in the January update

PR #4429, opened 2022-01-18
[Olefy] Update to Alpine 3.15 and include ping fix

DerLinkman:
This PR is updating the olefy-mailcow container to Alpine 3.15.

It is also including the ping fix from @a16bitsysop which is solving the issue: https://github.com/mailcow/mailcow-dockerized/issues/4401

The PR includes a temporarily local copy of the olefy.py file which includes the named fix. When this fix is officially merged into the olefy project we will update the olefy container again.

The new docker image tag is: **mailcow/olefy:1.8.1**


DerLinkman:
This PR has been tested on multiple instances.

PR #4437, opened 2022-01-26
 [Web] add version footer

### LINKED ISSUE #4434: Release cycle/versioning
Basicly for @DerLinkman and @andryyy

A quick idea from me I shared with @andryyy.

Since we now have our [first GitHub release](https://github.com/mailcow/mailcow-dockerized/releases/tag/major-release) we can somewhat release every month(?) with a tag like Docker does (YY.MM).
Also we can use [Milestones](https://github.com/mailcow/mailcow-dockerized/milestones) and since some users already asked us to implent a small hint in the Web UI which version they are on we can do it too.
https://github.com/mailcow/mailcow-dockerized/issues/2902
https://github.com/mailcow/mailcow-dockerized/issues/971
https://github.com/mailcow/mailcow-dockerized/issues/340

FreddleSpl0it:
This PR includes the current git version tag and a link to the changelogs in the Mailcow Web UI.


MAGICCC:
Neat, but maybe it would be better to have the version tag in the right corner (I am using a full HD screen and its floating somewhere between center and right).
image

FreddleSpl0it:
should be done, but I couldn't test it :)

DerLinkman:
Works: 

image

DerLinkman:
@MAGICCC can you check it on your Monitor as well?

DerLinkman:
@MAGICCC Does the Link (Version Tag) work?

MAGICCC:
Yes works fine.
Not sure if removing `padding: 0 !important;` would be an idea so it's not right next to the scrollbar

image

image

DerLinkman:
@FreddleSpl0it what do you think about this?

DerLinkman:
> Yes works fine. Not sure if removing `padding: 0 !important;` would be an idea so it's not right next to the scrollbar
> 
> image
> 
> image

Interestingly i don´t have this "issue". On my WQHD and Full HD screen it looks fine to me:
image

FreddleSpl0it:
i found a missing `</div>` that was causing problems.
`width: 100% !important;
padding: 0 !important;`
was not necessary

MAGICCC:
Hm I have the same 'problem' like before again, that the tag is not really on the right

FreddleSpl0it:
Are you sure you used the right patch? If so, could you try to empty the browser cache and reload?
With this patch the footer container should no longer be a child element of the main container.

https://user-images.githubusercontent.com/75116288/151770766-1b1e754c-c3de-45b0-b459-ed2dcac8c35f.mp4

DerLinkman:
Please test it guys! I want to release the 2022-01a Update tomorrow!

MAGICCC:
> Are you sure you used the right patch? If so, could you try to empty the browser cache and reload? With this patch the footer container should no longer be a child element of the main container.
> mailcow.UI.mp4

ah it seems I misunderstood there something :p  
It's ok for me too.

@DerLinkman If you want you can merge it, I'll let you do it :)

PR #4440, opened 2022-01-28
[WebAuthn] fido2 passwordless auth - fix

FreddleSpl0it:
Undo some changes i made, that may lead to some problems when a lot of fido2 devices are registered.

Originally i made those changes, because i experienced some problems with fido2 only authentication on Android with internal authenticator.
Android devices didn't react when the server send getArgs parameters with an empty allowCredentials array.

~~I also set the UV flags for WebAuthn 2FA from false to 'discouraged'. 
This will fix this [issue](https://github.com/mailcow/mailcow-dockerized/issues/4421#issuecomment-1019378066)~~
test failed. I will leave it for now that the UV flags are set to false.


DerLinkman:
Is the fix done?

So that we can merge it?

PR #4446, opened 2022-02-02
[Clamd] Update to clamav-0.104.2

MAGICCC:
This PR updates ClamAV to 0.104.2.
The Image Tag will be **mailcow/clamd:1.44** (already pushed)

Due to changes of the build mechanics it will use cmake now. Also adding + updates tini from the [official ](https://github.com/krallin/tini) GitHub repository instead adding it manually to the image

Image is currently in test since 2 days on my cow, no issues so far


DerLinkman:
Looks good :)

PR #4456, opened 2022-02-07
[Dovecot/FTS] Replace Solr with Xapian

DerLinkman:
### DO NOT MERGE UNTIL ALL CHECKS ARE MADE

**This PR replaces Solr FTS with Xapian FTS**

It is usable but not nearly finished!

What is done? What needs to be added?:

---

### Xapian itself
- [X] Dovecot Update + Xapian compilation
- [X] Xapian variable replacement from mailcow.conf and docker-compose.yml
- [X] Xapian Replacement in generate-config.sh & update.sh
- [X] Addition of new extra config file to control the Xapian Plugin

---

### WebUI Stuff
- [ ] Reorganisation of the general UI (Indevelopment by @FreddleSpl0it)
image
**NOT Final | Revision 2**
- [ ] Xapian Replacement in WebUI
- [X] Add new Reindex Dovecot Indexes Button in WebUI (controlled with DockerAPI) (Xapian section) **NOT FINAL**

---

### Behind the scenes stuff (API, Translations etc.)
- [ ] Remove/Replace Solr Strings with Xapian (Translations)
- [ ]  Solr API removal (Xapian replacement?)
- [ ] Cleanups (Dovecot Image)
- [ ] Final Tests


MAGICCC:
Nice! We could convert the PR into a draft so we won't merge it accidentally :D 

Also we need to redo the translation parts (renaming Solr to Xapian)
What about old installations and their configs? Do you plan to try to remove the corresponding lines from the mailcow.conf?
E.g.
https://github.com/mailcow/mailcow-dockerized/blob/master/generate_config.sh#L250-L257

DerLinkman:
Why do we need to remove the Xapian translation? It is replacing the solr. I didn't tested it until now. New mailcow Installations works flawlessly

data/Dockerfiles/dovecot/Dockerfile
```diff
@@ -84,12 +84,23 @@ RUN groupadd -g 5000 vmail \
...
+  build-essential \
+  autoconf \
+  automake \
+  libtool \
+  make \
+  libxapian-dev \
+  libicu-dev \
+  zlib1g-dev \
+  pkg-config \
+  sudo \
```
mkuron: You shouldn’t need sudo in the Dockerfile, it all runs as root.

data/Dockerfiles/dovecot/Dockerfile
```diff
@@ -101,13 +112,32 @@ RUN groupadd -g 5000 vmail \
...
+     make \
+     libxapian-dev \
+     libicu-dev \
+     zlib1g-dev \
+     pkg-config \
+     build-essential \
   && apt-get autoclean \
   && rm -rf /var/lib/apt/lists/* \
-  && rm -rf /tmp/* /var/tmp/* /root/.cache/
+  && rm -rf /tmp/* /var/tmp/* /root/.cache/
```
mkuron: You need to merge everything into a single `RUN` command, otherwise this cleanup won‘t help shrink the image size.

data/Dockerfiles/dovecot/decode2text.sh
```diff
@@ -0,0 +1,105 @@
...
+# plugin {
+#   fts_decoder = decode2text
+# }
+# service decode2text {
+#   executable = script /usr/local/libexec/dovecot/decode2text.sh
+#   user = dovecot
+#   unix_listener decode2text {
+#     mode = 0666
+#   }
+# }
```
mkuron: I would suggest to drop the attachment extraction script. We didn‘t have attachment indexing enabled with Solr either and I don’t think we have the necessary packages installed in this image for PDF or doc/xls/ppt support. Attachments also tend to be much larger than the email message themselves and indexing would thus increase the xapian database size significatnly.

mkuron (changes requested):
Could you please provide some rough numbers how much memory and disk space Xapian and Solr take up for an "average" user‘s mailbox, e.g. 10000 messages? Xapian should be better than Solr, but we should know by how much.

DerLinkman:
> Could you please provide some rough numbers how much memory and disk space Xapian and Solr take up for an "average" user‘s mailbox, e.g. 10000 messages? Xapian should be better than Solr, but we should know by how much.

### Test Results (Default Xapian config)

**Cons:**
- Xapian is slower than solr (5 mins min)
- needs more space on disk (100% min)
- Dovecot Image is larger than without Xapian (needs to be compiled)

**Pros:**
- No Java!
- Can run with 512MB Ram
- Quick query results (650 Values in 14ms)
- Indexes are not single files (these are Databases so to say)
- Build in Dovecot, no extra container needed

mkuron:
Those numbers sound pretty good.

> needs more space on disk (100% min)

Disk space is cheap (a lot cheaper than RAM), so that's a tradeoff probably anyone would be willing to make.

> Xapian is slower than solr (5 mins min)

The time the initial indexing takes is pretty much irrelevant because it's a one-time cost. On-the-fly indexing of new messages will take negligible time.

> Quick query results (650 Values in 14ms)

14ms seems like an absolutely acceptable query time, though I don't know how fast SOLR is.

andryyy:
As long as this is just to play around, it’s fine. Don’t ever merge this to master, please, before several things are fixed and there is an official release.

vermium:
I think the issue title should be named Replace Solr with Xapian, as I currently understands the title as going from Xapian to Solr

DerLinkman:
> I think the issue title should be named Replace Solr with Xapian, as I currently understands the title as going from Xapian to Solr

Absolutely right. Don't know if i was drunk or something why that happend :)

Thanks anyways

jpbaril:
@DerLinkman Feature abandoned?

DerLinkman:
No just redone with the current mailcow development branches (see: https://github.com/mailcow/mailcow-dockerized/tree/feat/fts-flatcurve)

Issue #4315, opened 2021-11-08
Docker Compose v2


chriscroome:
[Docker Compose v2](https://github.com/docker/compose#docker-compose-v2) is available and [compose-switch](https://github.com/docker/compose-switch) can be used via alternatives on Debian based distros to convert `docker-compose` commands to `docker compose` commands, I realise that the Mailcow `update.sh` script checks the version to use via [this URL](https://www.servercow.de/docker-compose/latest.php) and I don't think this is an urgent or critical issue but I thought it probably deserved a feature request opening as some work might be needed in the future to update the way Mailcow installs and updates Docker Compose.

MAGICCC:
Indeed its released and there's still an issue on compose side with IPv6: https://github.com/docker/compose/issues/8721
Also I will link some closed issues here too: https://github.com/mailcow/mailcow-dockerized/issues/4285 https://github.com/mailcow/mailcow-dockerized/issues/4280

nightah:
https://github.com/docker/compose/issues/8721 can be worked around but surrounding ipv6 literals with square brackets and double quotes, which is common for Go applications.

The key issue which prevented IPv6 routing from working was https://github.com/docker/compose/pull/8851 which has been resolved and merged in a newer version.

HorayNarea:
Now that there are no (known :sweat_smile:) bugs preventing the migration to v2… is there a plan to do this, maybe even a rough estimate when this will happen?

StarkZarn:
It doesn't look like there are any issues preventing this migration. I'll echo what the above poster asked, is there a plan for this? It would increase support for arm-based setups.

andryyy:
Am I missing something? Supporting v2 does help support ARM? ☺️

StarkZarn:
> Am I missing something? Supporting v2 does help support ARM? :relaxed:

Since compose v2 is in GO, there's native arm support. I know there was arm support from the pip package version of docker-compose v1, but it wasn't officially supported by docker, hence not having an arm release in the official packages.

It's just all-around better in my experience so far.

spantaleev:
#4285 is still a problem, for which I've [found some workaround](https://github.com/mailcow/mailcow-dockerized/issues/4285#issuecomment-1031553935).

I don't use IPv6, so I'm not hitting any other issues besides #4285 with Compose v2 + `compose-switch` so far.

----

[...]

PR #4457, opened 2022-02-07
Fall back to empty string if WATCHDOG_NOTIFY_EMAIL undefined

spantaleev:
By default, `.env` (`mailcow.conf`) does not define `WATCHDOG_NOTIFY_EMAIL`.

Using it in `docker-compose.yml` without having it defined leads to docker-compose v2 ([Compose v2](https://docs.docker.com/compose/cli-command/)) displaying this warning on startup:

> WARNING: The WATCHDOG_NOTIFY_EMAIL variable is not set. Defaulting to a blank string.

Related to https://github.com/mailcow/mailcow-dockerized/issues/4315


MAGICCC:
Thanks, will also suppress the warning in v1 docker-compose

PR #4462, opened 2022-02-10
Make quarantining-disabled log message better

spantaleev:
With this patch, something like this gets logged:

> WARNING: [pool system-worker] child 42 said into stderr: "NOTICE: PHP message: QUARANTINE: Not quarantining 564 bytes message, because Quarantine is not enabled (maximum quarantine size is 0 MiB)"

instead of:

> WARNING: [pool system-worker] child 42 said into stderr: "NOTICE: PHP message: QUARANTINE: Message too large: 564 b exceeds 0 b"

The new message seems more helpful in telling the server admin that
quarantining is disabled, so they should go and enable it.

The old message sent me digging through the code to figure out why
`Q_MAX_SIZE` is `0` and whether it got reset for me somehow, where it's coming from
and how I could configure it.
In the end, I did figure out where to configure the Quarantine feature from,
but I wasted some time in confusion chasing (non-)bugs.

-------

I suppose for quarantining to actually do anything, `Q_RETENTION_SIZE` must also be `>0`. I only see that value (`$retention_size`) used way down in `data/conf/rspamd/meta_exporter/pipe.php` though and only for restricting the total number of quarantined messages.


milkmaker:

PR #4464, opened 2022-02-12
oletools: disable template injection detection

mkuron:
This check seems to be causing a lot of false positives lately. It misrecognizes Office documents that were created from a template as embedding remote files -- when really they are only referring to the file path of their template. So let's disable it for now.

This check was actually disabled in upstream oletools in May 2021 (https://github.com/decalage2/oletools/commit/3cae86e62d99a29858721ca6f99c8c2cf15be8e4) and in Heinlein's fork in December 2021 (https://github.com/HeinleinSupport/oletools/commit/3d7f26c696604dbf434fb354a579fa9d73688e96). We are currently using Heinlein's master branch, which is severely outdated (November 2020). We should switch to their stable-patched branch (change master.zip to stable-patched.zip in line 11), but since that's potentially a bigger change that will require additional testing, for now I am only disabling this check. My server has been running with this patch for a few weeks now without problems and without false positives.

PR #4479, opened 2022-02-19
Update and fix oletools

mkuron:
As noticed by @MAGICCC (https://github.com/mailcow/mailcow-dockerized/pull/4464#issuecomment-1042347368), our olefy image does not work anymore if you rebuild it. This is because @HeinleinSupport recently updated their repository with the changes from @decalage2's repository, which renamed `olvba3` to `olevba`. Since @HeinleinSupport does not recommend using its own patched branch and is very slow in pulling in changes from upstream (@decalage2), let's switch to the latter. This also allowed me to revert #4464.

Finally, a minor patch to rspamd is necessary. While the [documentation](https://rspamd.com/doc/modules/external_services.html#oletools-extended-mode) says

> In the extended mode the oletools module will not trigger on specific categories, but will always set a threat string with all found flags when at least a macro was found.

This is not actually true -- it only sets it when _suspicious_ or _autoexec_ threats were detected. But it's a one-line patch to make rspamd behave as documented and we should submit that patch to @rspamd too. With this patch, I have confirmed that Mailcow will reject any incoming, non-whitelisted message containing attachments with macros.

### DIFF
--- a/data/Dockerfiles/olefy/Dockerfile
+++ b/data/Dockerfiles/olefy/Dockerfile
@@ -8,9 +8,9 @@ RUN apk add --virtual .build-deps gcc musl-dev python3-dev libffi-dev openssl-de
   && apk add --update --no-cache python3 py3-pip openssl tzdata libmagic \
   && pip3 install --upgrade pip \
   && pip3 install --upgrade asyncio python-magic \
-  && pip3 install --upgrade https://github.com/HeinleinSupport/oletools/archive/master.zip \
-  && apk del .build-deps \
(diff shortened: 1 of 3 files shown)


MAGICCC:
I just built **mailcow/rspamd:1.80** and **mailcow/olefy:1.9** and pushed them to dockerhub. I will test the changes in the next days. If anyone else want to it too, feel free to do so

DerLinkman:
Done with the tests? What are the results? @MAGICCC

MAGICCC:
Yes and had no issues so far

DerLinkman:
Used the containers now as well. By now no errors. But i´ll wait a bit for the merge, which will probably today anyway.

DerLinkman (approved):
Looks good! No issues so far.

Ready to merge!

PR #4481, opened 2022-02-21
[Web] Fix padding issue in UI admin panel

FreddleSpl0it:
This PR fixes a css padding issue in the Web UI Admin Panel.
panel-fix


MAGICCC:
thanks!

Issue #4308, opened 2021-11-01
Changing 2FA from Fido/U2F to TOTP 


bastischubert:
**Prior to placing the issue, please check following:** *(fill out each checkbox with an `X` once done)*
- [x ] I understand that not following or deleting the below instructions will result in immediate closure and/or deletion of my issue.
- [x ] I have understood that this bug report is dedicated for bugs, and not for support-related inquiries.
- [x ] I have understood that answers are voluntary and community-driven, and not commercial support.
- [x ] I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

## Summary
With enabled FIDO/U2F two factor authentication its quite difficult to switch to TOTP, while the other way around works "as expected".

## Reproduction
After enabling the TOTP, the FIDO/U2F is stil shown active and has to be "deleted" via the remove function. After that the TOTP Method is active.
The other way from TOTP to FIDO there's no second 2FA  in the background, becaus trying to delete a TOTP results in "last key cannot be deleted... disable 2fa .."

## System information

| Question | Answer |
| --- | --- |
| My operating system | Ubuntu 20 |
| Is Apparmor, SELinux or similar active? | No |
| Virtualization technlogy (KVM, VMware, Xen, etc - **LXC and OpenVZ are not supported** | KVM |
| Server/VM specifications (Memory, CPU Cores) | 16GB / 6 Cores |
| Docker Version (`docker version`) | 20.10.10 |
| Docker-Compose Version (`docker-compose version`) | 1.29.2 |
| Reverse proxy (custom solution) | Hell no.. nothing :-)))) |

- Output of `git diff origin/master`, any other changes to the code? If so, **please post them**.
```
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -575,36 +575,6 @@ services:
           aliases:
             - ofelia
 
-    ipv6nat-mailcow:
-      depends_on:
-        - unbound-mailcow
-        - mysql-mailcow
-        - redis-mailcow
-        - clamd-mailcow
-        - rspamd-mailcow
-        - php-fpm-mailcow
-        - sogo-mailcow
-        - dovecot-mailcow
-        - postfix-mailcow
-        - memcached-mailcow
-        - nginx-mailcow
-        - acme-mailcow
-        - netfilter-mailcow
-        - watchdog-mailcow
-        - dockerapi-mailcow
-        - solr-mailcow
-      environment:
[...]

PR #4484, opened 2022-02-22
[Web] change tfa flow

FreddleSpl0it:
### Changes
Here I have made a few more changes related to the tfa flow. I did this because a user on the Telegram channel showed that you can only use one authenticator for tfa. This problem is probably related to an older issue. ([#4308](https://github.com/mailcow/mailcow-dockerized/issues/4308))

After the changes, admins or domain admins, with an active tfa, are greeted with a modal where they can select a registered authenticator to log in.

### Update 07.03.2022
I have enabled 2 factor authentication for user as requested [(#4494)](https://github.com/mailcow/mailcow-dockerized/issues/4494)

Screenshot 2022-02-22 093709

**I have tested this so far, but would still be happy if others would also test it, because there are a lot of changes :).**

### Fixes
- probably fixes [#4308](https://github.com/mailcow/mailcow-dockerized/issues/4308)
- Domainadmins are currently unable to register webauthn due to a bug
- The password confirmation when registering or deactivating tfa is always true no matter what you enter.


MAGICCC:
Neat, I cant test it because I dont have a $yubikey

FreddleSpl0it:
if you want to test webauthn then you can emulate multiple authenticator in chrome 
[https://developer.chrome.com/docs/devtools/webauthn/](https://developer.chrome.com/docs/devtools/webauthn/)

The only thing I could not test is Yubico OTP

MAGICCC:
I received my Yubikey today and added it to my cow and applied your changes.
~~Seems I don't see the TFA confirmation dialog anymore. Tried to clear and using different browsers...~~
Selector is fine but after login I have only an empty page when using my Yubikey, TOTP is fine
Maybe also needs a rebase on master branch

Need to check why Webauth doesnt work, maybe because my reverseproxy?!

MAGICCC:
Works now as expected. @andryyy or @DerLinkman, would be great if anyone could also check.

DerLinkman:
Will test this tomorrow.

PR #4487, opened 2022-02-23
increase opcache.interned_strings_buffer to 16

BigMichi1:
since version 23.0.2 Nextcloud recommends having a value greater than 8 for `opcache.interned_strings_buffer`. As this memory will be only used when needed this should have no impact on installations that are not using nextcloud.

related discussion: https://help.nextcloud.com/t/nextcloud-23-02-opcache-interned-strings-buffer/134007/19
related nextcloud issue: https://github.com/nextcloud/server/issues/31223


MAGICCC:
Looks okish, but can you change the base branch you want to merge to to staging?

PR #4501, opened 2022-03-04
Detect OpenRC on multiple systems

MathyV:
In #4244 there was basic support added for the OpenRC system in Alpine Linux, however the implementation is not universal as it checks specifically for the /etc/alpine-release file. There are other distributions using this init system, in my case Gentoo, so the check is not complete. 

Checking if the distribution installed is of a certain flavor is also not portable as for example on Gentoo you can choose either OpenRC or systemd. I adapted the test to check for the existence of the rc-service command, which should be a better indication of an OpenRC system and would work for any distribution.

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -48,6 +48,9 @@ export LC_ALL=C
 DATE=$(date +%Y-%m-%d_%H_%M_%S)
 BRANCH=$(cd ${SCRIPT_DIR}; git rev-parse --abbrev-ref HEAD)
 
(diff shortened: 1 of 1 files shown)


MathyV:
@andryyy since you merged the previous OpenRC commit, is there anything I can do to speed up the process? Updating and applying my patch is kind of a convoluted process (unless there is some easy "developer" option to use my own branch I haven't found yet?) so I rather have this accepted sooner than later :laughing:

milkmaker:

PR #4502, opened 2022-03-04
Can't restore from older backup because script crashes?

roboalex2:
Hi, i had an older mailcow instance (don't know how old exactly... 2 years max). I created a backup(helper script) and deleted the old mailcow server. Then i made a fresh mailcow install. After it came up once i tried to restore my backup on it but after doing so the server was endless stuck on the still processing page. Sogo spammed the log with `Waiting for schema update...`. After looking through the logs a bit longer i found that the `data/web/inc/init_db.inc.php` script faild because a few views already existed in the DB (because they were already added through the restored backup), but whatever comes after the view creation would have been necessary to make the `Waiting for schema update...` problem go away and make the site work in general. 

Long story short after adding an `IF NOT EXISTS` check to the view creation part of the `data/web/inc/init_db.inc.php` script my backup restore problem was gone and since i don't see how that change could harm someone but maybe prevent other people  from running into the same problem as me i made it into a pull request.

### DIFF
--- a/data/web/inc/init_db.inc.php
+++ b/data/web/inc/init_db.inc.php
@@ -23,7 +23,7 @@ function init_db_schema() {
     }
 
     $views = array(
-    "grouped_mail_aliases" => "CREATE VIEW grouped_mail_aliases (username, aliases) AS
+    "grouped_mail_aliases" => "CREATE VIEW IF NOT EXISTS grouped_mail_aliases (username, aliases) AS
       SELECT goto, IFNULL(GROUP_CONCAT(address ORDER BY address SEPARATOR ' '), '') AS address FROM alias
       WHERE address!=goto
       AND active = '1'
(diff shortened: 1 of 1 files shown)


mkuron (changes requested):
This change does absolutely nothing. Down in line 1185 we run a `DROP VIEW IF EXISTS` for all these views before we do the `CREATE VIEW`. If it appears to have fixed anything for you, it must have been purely coincidental.

roboalex2:
> This change does absolutely nothing. Down in line 1185 we run a `DROP VIEW IF EXISTS` for all these views before we do the `CREATE VIEW`. If it appears to have fixed anything for you, it mist have been purely coincidental.

You are right the views are dropped... This is weird. I am 100% certain (tried about 6 times to restore the DB and allways got the same error) that it faild because the views still existed when it came to the creates. Could it be that the old backup (still used a sqldump) set some weird isolation levels or locks etc? Well i guess it dosn't matter at this point... (Already got rid of the backup and i didn't keep the logs). Sorry for wasting your time.

Issue #4485, opened 2022-02-22
SOGo: Password too long


Cheatha:
**Prior to placing the issue, please check following:** *(fill out each checkbox with an `X` once done)*
- [X] I understand that not following or deleting the below instructions will result in immediate closure and/or deletion of my issue.
- [X] I have understood that this bug report is dedicated for bugs, and not for support-related inquiries.
- [X] I have understood that answers are voluntary and community-driven, and not commercial support.
- [X] I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

## Summary
User can't log into SOGo if password length is more than 135 characters.

## Logs
```Feb 22 13:25:20 c41d2a948b9e sogod [93]: SOGoRootPage successful login from '2a01::XXX' for user 'test@example.com' - expire = -1  grace = -1
Feb 22 13:25:21 c41d2a948b9e sogod [93]: 2a01::XXX "POST /SOGo/connect HTTP/1.0" 200 41/222 0.274 - - 0 - 14
Feb 22 13:25:22 c41d2a948b9e sogod [93]: <0x0x55e1c46eb480[SOGoWebAuthenticator]> tried wrong password for user '<veryverylongstring>'!
Feb 22 13:25:22 c41d2a948b9e sogod [93]: 2a01::XXX "GET /SOGo/so/test%40example.com HTTP/1.0" 200 10343/0 0.018 37050 72% 0 - 14
```
No logs in browser console

## Reproduction
Create a new mailbox with a password with more than 135 characters. Try to login into SOGo. Login fails. Change user password to 135 or less characters. Login into SOGo will succeed.

## System information

| Question | Answer |
| --- | --- |
| My operating system | Debian 10.11|
| Is Apparmor, SELinux or similar active? |No|
| Virtualization technlogy (KVM, VMware, Xen, etc - **LXC and OpenVZ are not supported** | Proxmox KVM|
| Server/VM specifications (Memory, CPU Cores) |4 Cores, 8GB RAM|
| Docker Version (`docker version`) |Docker version 20.10.12, build e91ed57 |
| Docker-Compose Version (`docker-compose version`) | docker-compose version 1.29.2, build 5becea4c|
| Reverse proxy (custom solution) |No |

Used crypt lib is BLF-CRYPT. And: thank you for this great project! <3

MAGICCC:
Not sure if this is mailcow related or SoGo
[...]

PR #4506, opened 2022-03-11
[SOGo] Update SOGo to 5.5.1

DerLinkman:
This commit includes the SOGo Update to 5.5.2.

The new Image has been pushed to docker **BUT** don´t work yet.

Im not 100% sure why it don´t.

Error message i got:
```
cannot write record: <MySQL4Exception: 0x55f71168a0a0> NAME:ExecutionFailed REASON:Data too long for column 'c_value' at row 1
```

The new Image Tag will be **mailcow/sogo:1.107**

---

### Update 25.03.2022:
DB is now working.
PR delayed until SOGo 5.5.2

### Update 01.04.2022 (No April Fool):
Decided to bring SOGo 5.5.1 as the April Update. 5.5.2 will be part of 2022-04a then (if its releasing shortly after the April Update)


jmber:
https://github.com/mailcow/mailcow-dockerized/blob/2f9da5ae93d93bf62a8c2b7a5a6ae50a41170c48/data/web/inc/init_db.inc.php#L867

You can fix the problem, by changing the databasefield c_value to VARCHAR(4096) 

during the update process, we have to do:

`ALTER TABLE sogo_sessions_folder MODIFY COLUMN c_value VARCHAR(4096);`

because of this commit by sogo:
https://github.com/inverse-inc/sogo/commit/a965f276c8155f50d4aeb36b15fe35bcf6e48bd3

You build the image with the nightly debs of sogo, so it is not version 5.5.1, it is the "master"

DerLinkman:
We´ll always build from nightly. It is 5.5.1 in the Package Repository.

MAGICCC:
Also see https://github.com/mailcow/mailcow-dockerized/issues/4485#issuecomment-1064381197

FreddleSpl0it:
> https://github.com/mailcow/mailcow-dockerized/blob/2f9da5ae93d93bf62a8c2b7a5a6ae50a41170c48/data/web/inc/init_db.inc.php#L867
> 
> You can fix the problem, by changing the databasefield c_value to VARCHAR(4096)
> 
> during the update process, we have to do:
> 
> `ALTER TABLE sogo_sessions_folder MODIFY COLUMN c_value VARCHAR(4096);`
> 
> because of this commit by sogo: [inverse-inc/sogo@a965f27](https://github.com/inverse-inc/sogo/commit/a965f276c8155f50d4aeb36b15fe35bcf6e48bd3)
> 
> You build the image with the nightly debs of sogo, so it is not version 5.5.1, it is the "master"

https://github.com/mailcow/mailcow-dockerized/blob/2f9da5ae93d93bf62a8c2b7a5a6ae50a41170c48/data/web/inc/init_db.inc.php#L6

also $db_version needs to be updated to apply changes.

feldsam:
> `ALTER TABLE sogo_sessions_folder MODIFY COLUMN c_value VARCHAR(4096);`


hmm, interesting, from what I know, VARCHAR type in mysql/mariadb can be max 255 chars. We should use `TEXT(4096)`. Or I miss something?

mkuron (approved):

mkuron:
Actually, it's fine, you can have up to 64 KiB in a VARCHAR according to https://mariadb.com/kb/en/varchar/.

DerLinkman:
Yes but how do we add this to existing mailcows? The DB Scheme needs to be changed if i´m not 100% wrong right?

feldsam:
Db init runs during updates too. It handles alters.

DerLinkman:
> ALTER TABLE sogo_sessions_folder MODIFY COLUMN c_value VARCHAR(4096);

So with this change made it will also works for existing mailcows?

FreddleSpl0it:
> > https://github.com/mailcow/mailcow-dockerized/blob/2f9da5ae93d93bf62a8c2b7a5a6ae50a41170c48/data/web/inc/init_db.inc.php#L867
> > 
> > You can fix the problem, by changing the databasefield c_value to VARCHAR(4096)
> > during the update process, we have to do:
> > `ALTER TABLE sogo_sessions_folder MODIFY COLUMN c_value VARCHAR(4096);`
> > because of this commit by sogo: [inverse-inc/sogo@a965f27](https://github.com/inverse-inc/sogo/commit/a965f276c8155f50d4aeb36b15fe35bcf6e48bd3)
> > You build the image with the nightly debs of sogo, so it is not version 5.5.1, it is the "master"
> 
> https://github.com/mailcow/mailcow-dockerized/blob/2f9da5ae93d93bf62a8c2b7a5a6ae50a41170c48/data/web/inc/init_db.inc.php#L6
> 
> also $db_version needs to be updated to apply changes.

i think editing these two lines should do the trick
` $db_version = "DATE_TIME"; `
`"c_value" => "VARCHAR(4096) NOT NULL", `

the table should then be adjusted automatically after restart

feldsam (changes requested):
Update db_version

DerLinkman (commented):
DB Version has been updated.

data/web/inc/init_db.inc.php
```diff
@@ -864,7 +864,7 @@ function init_db_schema() {
       "sogo_sessions_folder" => array(
         "cols" => array(
           "c_id" => "VARCHAR(255) NOT NULL",
-          "c_value" => "VARCHAR(255) NOT NULL",
+          "c_value" => "VARCHAR(4096) NOT NULL",
```
feldsam: there is some white space at the end of this line added

feldsam (approved):

data/web/inc/init_db.inc.php
DerLinkman: Oh yes i saw that. Will fix it

feldsam (approved):

DerLinkman:
Can someone test this? I´ll test it tomorrow, today is my last day of my exam.

DerLinkman:
Ok i´ve tested it on my test machine. It don´t work either. I´ve implemented the fixes mentioned above. Can anyone else confirm that?

Still: Unhandled error response

Logs don´t show any curious errors

jmber:
DB Schema updated?

Login to the DB:

source mailcow.conf
docker-compose exec mysql-mailcow mysql -u${DBUSER} -p${DBPASS} ${DBNAME}

And then execute this SQL Command:

describe sogo_sessions_folder;

What does it shows?

DerLinkman:
image

Looks good to me...

DerLinkman:
Ok, due to some research i´ve found the solution for that.

It seems that the new SOGo Update requires bigger proxy_buffer_sizes in the nginx conf.

What i did:

Added:
    proxy_buffer_size          128k;
    proxy_buffers              64 512k;
    proxy_busy_buffers_size    512k;

Uncommented:
    proxy_buffers 64 512k;  at the active Sync section.

To nano data/conf/nginx/includes/site-defaults.conf before location settings.

Also see the commit.

mkuron (approved):
Looks good to me, although I would like to know what changed in SOGo so that it now requires that different proxy buffer size.

DerLinkman:
> Looks good to me, although I would like to know what changed in SOGo so that it now requires that different proxy buffer size.

Actually, i have no clue...

Maybe it is something with the increased password size.

feldsam:
I think too, it can be related to password size and increased cookie size.

jmber:
https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_buffer_size
The default  proxy_buffer_size of nginx is 4k. So it it to small for the new password/cookie size.

Perhaps we can disable proxy_buffering of nginx for the future? To avoid such problems?

[proxy_buffering](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_buffering) on – NGINX Plus receives a response from the proxied server as soon as possible, and buffers it (the default). If disabled, NGINX Plus passes the response to the client synchronously, as soon as it is received, which increases the load on NGINX Plus.

Disabling response buffering is necessary only for applications that need immediate access to the data stream.

mkuron:
Okay, that makes sense. Since buffering is the default on nginx, let's keep it on and just make the buffer big enough for the new password cookie.

DerLinkman:
So is that a approval from you? Did anyone else tested this?

MAGICCC:
Just tested the new version and works fine so far. 
The new password policy is not included in 5.5.1 when I check the commits, https://github.com/inverse-inc/sogo/compare/SOGo-5.5.0...SOGo-5.5.1

DerLinkman:
Yes it has been confirmed for 5.5.2. 

See: https://www.sogo.nu/bugs/changelog_page.php

DerLinkman:
So i would say we skip 5.5.1 and wait for 5.5.2.

soulsymphonies:
IMPORTANT:
**For all those who have a reverse proxy in front of mailcow**

like this here: https://mailcow.github.io/mailcow-dockerized-docs/post_installation/firststeps-rp/#nginx

You have to increase the proxy buffer size in your reverse proxy also, or you can't log into SoGo anymore

The Active Sync location directive has to be change:
You need to replace `proxy_buffers 64 256k;` with `proxy_buffers 64 512k;`

```
location /Microsoft-Server-ActiveSync {
    proxy_pass http://127.0.0.1:8080/Microsoft-Server-ActiveSync;
    proxy_set_header Host $http_host;
    proxy_set_header X-Real-IP $remote_addr;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    proxy_set_header X-Forwarded-Proto $scheme;
    proxy_connect_timeout 75;
    proxy_send_timeout 3650;
    proxy_read_timeout 3650;
    proxy_buffers 64 512k;
    client_body_buffer_size 512k;
    client_max_body_size 0;
  }
```

You also need to add the following to the root location directive
```
proxy_buffer_size 128k;
proxy_buffers 64 512k;
proxy_busy_buffers_size 512k;
```

It looks like this then:

```
  location / {
    proxy_pass http://127.0.0.1:8080/;
    proxy_set_header Host $http_host;
    proxy_set_header X-Real-IP $remote_addr;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    proxy_set_header X-Forwarded-Proto $scheme;
    client_max_body_size 0;
    proxy_buffer_size 128k;
    proxy_buffers 64 512k;
    proxy_busy_buffers_size 512k;
  }
```

took me a while to figure this out, hope it helps ;)

immanuelfodor:
I can't imagine how long would have I debugged this after the update :D thanks a bunch @soulsymphonies

LovesTha:
> took me a while to figure this out, hope it helps ;)

Took me a while to find this, and I lucked out on this being the first PR thread I dived into the details of. Could a pointer to this have percolated up to the release notes?

PR #4515, opened 2022-03-21
[Postfix] update postscreen access list

AikoAppeldorn:
Hi,
I saw that the postscreen access list was a bit outdated, so I updated the list.


DerLinkman:
Looks good.

PR #4541, opened 2022-04-08
[Clamd] fix whitelist

feldsam:
@MAGICCC 

### DIFF
--- a/data/Dockerfiles/clamd/clamd.sh
+++ b/data/Dockerfiles/clamd/clamd.sh
@@ -14,10 +14,10 @@ rm -rf /var/lib/clamav/clamav-*.tmp
 
 mkdir -p /run/clamav /var/lib/clamav
 
-#if [[ -s /etc/clamav/whitelist.ign2 ]]; then
-#  echo "Copying non-empty whitelist.ign2 to /var/lib/clamav/whitelist.ign2"
-#  cp /etc/clamav/whitelist.ign2 /var/lib/clamav/whitelist.ign2
-#fi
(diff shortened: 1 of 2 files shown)


dragoangel:
Maybe add `if [[ ! -f /etc/clamav/whitelist.ign2 ]]; then` condition before https://github.com/mailcow/mailcow-dockerized/blob/020cb8b35cd7a330764dbc1e894b524eb5f71af0/data/Dockerfiles/clamd/clamd.sh#L43

What you think? As we also before that copy `/etc/clamav/whitelist.ign2` to `/var/lib/clamav/whitelist.ign2` if it exist...

feldsam:
@dragoangel check line 22

feldsam:
this clamd.sh file doesn't need any changes, I just reverting back what was removed

dragoangel:
Yes, I checked line 22 :)
We have data/conf mounted to etc. If we have file there we copy it from etc to var where it actually apply whitelist. Then we always copy it back from var to etc and I don't get why we need do it if we already have it :P. This not big issue, but just additonal unesesary copying of file on each container restart

feldsam:
because it is there for years, so why change it? it doesnt cause overwriting. Probably we can remove that line, because it is useless

dragoangel:
Oh, missed that maybe. It not useless as it create default whitelist if it missing once

MAGICCC:
Seems I did some other things thats why I commented it. What about using clamd:1.50 then I can repush with same version and we dont need to merge into master for this change?

DerLinkman:
Yeah but please move that into staging anyways for the integrity you know what I mean guys 😉

PR #4545, opened 2022-04-13
[Netfilter] Exclude banning IPs when dovecot server not reacheble

dragoangel:


DerLinkman:
It needs a rebuild of the netfilter container right?

dragoangel:
Yes I think

dragoangel:
Oh, and one note - this will impact only new setups or people who will click on reset button* in fail2ban settings

DerLinkman:
Ok, that is maybe an idea for the docs? Clear the fail2ban Settings (if it is not there)

dragoangel:
I think better for update notes, yes. Sort of:
`If you not have custom rules - reset them once after update. If you have custom rules please copy them to notes, reset rules and reaply your custom rules back`

andryyy:
Did you verify this 100% works as intended?

dragoangel:
Yes, and I not asking for a merge without checking. If you stop dovecot and try auth over smtp you will reproduce same log line. I validate it over regex101.

DerLinkman:
So again: Safe to merge? You already said it but I want to be sure it is 😝

dragoangel:
Yep :). I use this regex for a month.

DerLinkman:
Alright then.

PR #4573, opened 2022-05-05
Fix for /api/v1/get/mailbox/{email}

jkellerer:
The tagging feature breaks the API endpoint "`/api/v1/get/mailbox/{username}`" (without tags parameter). Prior to the change it returned the mailbox details (directly not as list) when "{username}" matches directly.

At the moment (without this PR) it return a list of all mailboxes when the tags parameter is not set. When tags is set, untagged mailboxes cannot be retrieved and result is always a list.

While a list is ok for the case that tags are specified (or in general when multiple mailboxes can be expected), it is not ok when the details for a single mailbox are requested.

If this was by intention then a breaking change should be documented (and maybe a migration path how to get the details of a single mailbox that has no tags).

### DIFF
--- a/data/web/json_api.php
+++ b/data/web/json_api.php
@@ -989,11 +989,15 @@ if (isset($_GET['query'])) {
                 if (isset($_GET['tags']) && $_GET['tags'] != '') 
                   $tags = explode(',', $_GET['tags']);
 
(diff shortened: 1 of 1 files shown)


FreddleSpl0it:
This was not intentional. Thank you for pointing that out :). I have tested the patch and it works as expected.

FreddleSpl0it (approved):

Issue #4586, opened 2022-05-13
Possible HTML Injection in Swagger UI


andristeiner:
### Summary

We use Mailcow as the [TYPO3 association's](https://typo3.org/) mail server. We have a fairly big audience around the world, and our security team does run a [bug bounty program](https://typo3.org/community/teams/security/bug-bounty-program) to encourage security researchers to hand in security problems not only within TYPO3 CMS, but also our infrastructure.

This week, we received such a report about Swagger UI within the Mailcow API. One can load an URL directly through the UI, or by adding the `url` parameter:  https://demo.mailcow.email/api/?url=https://null.org. Through a extra crafted OpenAPI spec and by adding the corresponding CORS headers, it might be possible to load arbitrary content within any Swagger UI.

As far as I know, this feature is not really required within Mailcow, as only the default spec from `/api/openapi.yaml` is used. Therefore, I propose to define this explicitly through the `urls` parameter instead of using `url`. This does also remove the ability to load another definition through the `url` get parameter (see `url` within [Swagger's documentation](https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration/)).

### Motivation

* Mailcow's default configuration will be more secure
* Organisations targeted by security researchers will receive no reports about a potential problem within the Mailcow API anymore
* the API, and its Swagger UI documentation, will still be fully usable

PR #4587, opened 2022-05-13
[Web] Swagger UI: explicitly define used OpenAPI specifications

andristeiner:
Update Swagger UI configuration to explicitly define used OpenAPI specifications. This does prevent the potential loading of arbitrary content from any other, external specifications. See #4586 for context.


andryyy:
Thank you! That's a great catch.

andryyy:
I will merge to master, can you PR to staging next time? :)

PR #4589, opened 2022-05-16
[Clamd] Update to ClamAV 0.105

DerLinkman:
The new Dockertag is **1.52**

The full ClamAV Changelog can be found here: https://blog.clamav.net/2022/05/clamav-01050-01043-01036-released.html


DerLinkman (commented):
In testing already for almost 2 Weeks.

**Works flawlessly so far**

PR #4590, opened 2022-05-17
Update swagger docs - add/app-passwd

FreddleSpl0it:
Update swagger docs. Include "protocols" parameter for /api/v1/add/app-passwd.


DerLinkman (approved):
Looks good 👍

PR #4597, opened 2022-05-23
[Web] Add ratelimits

dorianim:
A friend of mine recently asked if my mail server ha a dos protection and tested it with a simple script. That lead to 100% CPU utilization and made the web interface unusable. Adding these rate limits resolved the lock-up.

Maybe this is not the ideal solution, but some kind of rate-limiting would be nice in my opinion :)

### DIFF
--- a/data/conf/nginx/includes/site-defaults.conf
+++ b/data/conf/nginx/includes/site-defaults.conf
@@ -40,6 +40,7 @@
   location ~ ^/(fonts|js|css|img)/ {
     expires max;
     add_header Cache-Control public;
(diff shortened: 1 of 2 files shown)


kozicpetar:
I think you can that implement without some special feature, because mailcow using nginx and nginx have own rate limit function.

https://www.nginx.com/blog/rate-limiting-nginx/

dorianim:
Yes, I changed the nginx config to use exactly that :)

andryyy:
Hey, a PR with fixed values without discussion on this matter to master is really not okay. :(

PR #4601, opened 2022-06-01
Get up to speed with 2022-05d

snevas:
Merge master to branch


DerLinkman:
Excuse me? What is the point of this PR? Am I not understanding something? Or was that an accident?

DerLinkman:
Oh now i understand the PR!

Yeah thanks for that but that is not needed :) Since these feature Branches are only temporarily for testing/development purposes.

Once their finished they will be moved into staging and then in master.

**These branches are not intended to be used in production.**

PR #4531, opened 2022-04-02
Change default HTTP_BIND, HTTPS_BIND

aminvakil:
https://github.com/mailcow/mailcow-dockerized/issues/4315#issuecomment-1083034329

Fixes a part of https://github.com/mailcow/mailcow-dockerized/issues/4315.

PR #4605, opened 2022-06-05
[Compose] Update to Docker Compose v2

DerLinkman:
This PR includes the official Docker Compose v2 Support for the mailcow Stack.

It has been tested but might have some deep dive issues.

The mailcow Stack is also still usable with Compose v1.

**Caution: As a compatibility reason, from 2022-06 until the 2022-12 update, the web interface will only be accessible via v4 (see Changelog from Docker: https://docs.docker.com/engine/release-notes/#20106 ,PR: https://github.com/mailcow/mailcow-dockerized/pull/4531 and Issue: https://github.com/mailcow/mailcow-dockerized/issues/4315) as a temporary workaround, the IPv6 bind must be set in an extra docker-compose.override.yml (see [docs](https://mailcow.github.io/mailcow-dockerized-docs/post_installation/firststeps-ip_bindings/), after 2022-06 release).**

**The other Containers are not affected.**

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -374,8 +374,8 @@ services:
         - ./data/conf/rspamd/meta_exporter:/meta_exporter:ro,z
         - sogo-web-vol-1:/usr/lib/GNUstep/SOGo/:z
       ports:
-        - "${HTTPS_BIND:-:}:${HTTPS_PORT:-443}:${HTTPS_PORT:-443}"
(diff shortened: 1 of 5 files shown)


DerLinkman:
@MAGICCC and or @andryyy Can you guys test it aswell and may check the changed codes?

Runs fine on my maschines.

PR #4607, opened 2022-06-05
Netfilter base image update

MAGICCC:
Updates Netfilter image to Alpine 3.16
**mailcow/netfilter:1.48** is available on Docker Hub


DerLinkman (changes requested):
Looks good. But the docker-compose.yml has to be changed aswell :)

DerLinkman (approved):
Looks good.

Container is working as expected.

PR #4612, opened 2022-06-05
acme base image update

MAGICCC:
Updates acme image to Alpine 3.16
**mailcow/acme:1.82** is available on Docker Hub


DerLinkman (approved):
Already tested :)

PR #4629, opened 2022-06-13
hotfix imapsync

FreddleSpl0it:
Filter custom params for imapsync by a whitelist to prevent argument injection that lead to RCE.


feldsam:
Hello @FreddleSpl0it, there is problem with this PR. When we need to add custom param `--sslargs1 SSL_version=TLSv1` them it show error "Bad character SPACE" and it is not saved.

How can we solve this? Need to know, why you want to filter spaces, to propose fix.

PR #4654, opened 2022-06-29
[Postfix] Fix DMARC aligment for bounces

dragoangel:


dragoangel:
Let's assume `MAILCOW_HOSTNAME=mail.example.com`, then this settings will change email of bounce from `MAILER-DAEMON@mail.example.com` to `MAILER-DAEMON@example.com` and in this case ARC signing will aling checks and DMARC will not fail.

I run mailcow with this setting more then a year and have strict DMARC without issues.
@navossoc asked for help and this change fixed his situation when gmail reject bounces due to strick not aligned DMARC.
@DerLinkman I sure it safe to apply, but can you please double check in your setup?

DerLinkman:
> Let's assume `MAILCOW_HOSTNAME=mail.example.com`, then this settings will change email of bounce from `MAILER-DAEMON@mail.example.com` to `MAILER-DAEMON@example.com` and in this case ARC signing will aling checks and DMARC will not fail.
> 
> I run mailcow with this setting more then a year and have strict DMARC without issues. @navossoc asked for help and this change fixed his situation when gmail reject bounces due to strick not aligned DMARC. @DerLinkman I sure it safe to apply, but can you please double check in your setup?

Is there a way how i can test this?

dragoangel:
Send email to user which over quotas in imap for example and check what bounce dmarc status will be without this PR and with it :)
Note: better send from gmail

DerLinkman:
Tested:

Before: 
image

After:
image

andryyy:
This will change the sender origin to a domain that might be a mailcow domain. I don't really want this. The origin should imo remain the mail host as it is mailer daemon sending the mail. The service accounts belong to the mailcow hostnames fqdn.

This does "fix" dmarc alignment with a strange workaround and might not be wanted.

andryyy:
Furthermore there are hints to setup dmarc in the config which this PR would not remove or edit.

andryyy:
Furthermore there are hints to setup dmarc in the config which this PR would not remove or edit.

milkmaker:

PR #4661, opened 2022-07-05
[Sync Jobs] Remove default selection for target mailbox

### LINKED ISSUE #4642: [Sync Jobs] Remove default selection for target mailbox
### Summary

Please remove the default selection for the Username-field when creating a new sync job.

### Motivation

It should never happen that a sync job accidentally syncs to a wrong mailbox because someone forgets to select the correct target mailbox.

### Additional context

image

The username should have no pre-selection. The user should be forced to select it manually.
Currently all fields are also remembered when the Modal is closed and opened again, also after creating a new sync job. I think the Username-field should always be cleared when the modal is opened.
Remembering the other fields is fine, this is helpful when creating Sync Jobs for multiple mailboxes from the same source system.

RafaelKr:
fixes #4642

cc @FreddleSpl0it

### DIFF
--- a/data/web/templates/modals/mailbox.twig
+++ b/data/web/templates/modals/mailbox.twig
@@ -435,11 +435,11 @@
       </div>
       <div class="modal-body">
         <p class="help-block">{{ lang.add.syncjob_hint }}</p>
-        <form class="form-horizontal" data-cached-form="true" role="form" data-id="add_syncjob">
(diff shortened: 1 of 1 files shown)


andryyy:
Can you rebase to staging? :)

RafaelKr:
Ah of course! Done. :)

andryyy:
Merci

PR #4666, opened 2022-07-07
Update imapsync whitelist

mlfreeman2:
Added `inet4` option and `ssl1` option to whitelist.
1. `inet4` takes no args so it should be safe?
2. `ssl2` was already there so `ssl1` should be there as its counterpart.
3. Sorted whitelist alphabetically, removed redundant entries for `folderfirst`, `folderlast`, and `folderrec`


andryyy:
Things that mailcow UI (as checkbox or whatever) should not be included here.

andryyy:
Furthermore this is a list I don't want modified externally and esp not with a push to master without issue. :/

mlfreeman2:
> Things that mailcow UI (as checkbox or whatever) should not be included here.

These two options are not checkboxes in the UI.

mlfreeman2:
I'll open an issue in a little while.
Basically, this new whitelist broke editing existing jobs if they used an "unapproved" option.

mlfreeman2:
Closed this and opened issue #4667 instead

PR #4671, opened 2022-07-08
Removed DroneCI & Travis CI

ntimo:
This removes both Drone CI and Travis. Since we should aim to switch the project over to Github Actions for CI/CD.


DerLinkman (approved):
Looks fine.

Drone is offline anyways now.

A native GitHub Action Integration is way better nowadays i guess.

PR #4672, opened 2022-07-09
add schema for tags in get domain and get mailbox

l-with:
this is needed for using the open api specification for generating the api with the openapi generator for go language


MAGICCC:
Seems a dupe of https://github.com/mailcow/mailcow-dockerized/pull/4670

ntimo:
Yes its a duplicate. Thank you for also finding and fixing this :) I will close the mr anyways :) Since I guess we can only merge one fix

PR #4676, opened 2022-07-12
Update sieve extension rules

pantherman594:
Updated the following extensions according to their RFC specs:

- [duplicate](https://www.rfc-editor.org/rfc/rfc7352.html)
    - `:handle <handle: string>`
    - `:header <header-name: string>` or `:uniqueid <value: string>`
    - `:seconds <timeout: number>`
    - `:last`
- [execute](https://github.com/dovecot/pigeonhole/blob/main/doc/rfc/spec-bosch-sieve-extprograms.txt)
    - `:input <input-data: string>`
    - `:output <varname: string>` if using the variables extension
- [pipe](https://github.com/dovecot/pigeonhole/blob/main/doc/rfc/spec-bosch-sieve-extprograms.txt)
    - `:copy` if using the copy extension

### DIFF
--- a/data/web/inc/lib/sieve/extensions/copy.xml
+++ b/data/web/inc/lib/sieve/extensions/copy.xml
@@ -2,7 +2,7 @@
 
 <extension name="copy">
 
-	<tagged-argument extends="(fileinto|redirect)">
+	<tagged-argument extends="(fileinto|redirect|pipe)">
 		<parameter type="tag" name="copy" regex="copy" occurrence="optional" />
 	</tagged-argument>
 
--- a/data/web/inc/lib/sieve/extensions/duplicate.xml
+++ b/data/web/inc/lib/sieve/extensions/duplicate.xml
@@ -3,7 +3,16 @@
 <extension name="duplicate">
 
 	<test name="duplicate">
-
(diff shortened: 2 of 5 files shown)


andryyy:
Thanks.

I'm not sure if it wouldn't be better to remove extprograms from sieve_extensions and only keep it available for global sieve scripts (it already is registered there).

The validation would need to check for global/non-global sieve scripts before applying/accepting the filter from within the UI though.

This closes a security risk imo. How do you think about that? Also pinging @mkuron

mkuron:
I don't think I'm the right person to review this because I'm not familiar with our sieve parser PHP library. However, command execution should definitely be restricted to the server admin (i.e. to global rules) and not be available to domain admins.

pantherman594:
I believe it is only allowed in global sieve scripts by default. I'm not sure how the validation would be able to check whether it is allowed in global or non-global sieve scripts, because that's configured in dovecot.

I think it's more a ux issue than a security issue, since the validation may let the script through but it should fail on the dovecot side when it tries to load an extension that isn't allowed.

andryyy:
It is allowed for users and should be removed from sieve_plugins while being kept as global sieve plugin.

milkmaker:

PR #4673, opened 2022-07-11
[CI] Added Mailcow tests & image builds

ntimo:
This MR adds the mailcow integration tests to run via GitHub CI/CD. It also adds a build for the Docker images on overy push using a matrix build which is quite performant (it takes 5 minutes for all images). The images are not yet pushed anywhere just build.

P.S. I also renamed the twitter job name because it triggered me that it uses dashes instead of spaces like the other ones.

PR #4688, opened 2022-07-15
Use yaml list style in workflow

MAGICCC:
As seen in https://github.com/mailcow/mailcow-dockerized/pull/4673#issuecomment-1184881421 we may use this style for listing the images in the workflow


larena1:
That's definitely cleaner. I think you don't even need the quotes around the strings but you can also keep them if you prefer

ntimo (approved):
Thank you :) Looks awesome.

PR #4693, opened 2022-07-20
return msg always as array for domain api

l-with:
return msg always as array matching api specification

otherwise a generated api fails if a case occurs when msg is only a string instead an array as specified


milkmaker:

PR #4703, opened 2022-08-03
Improve send-as behaviour

macwinnie:
Receiving mails for wildcard alias addresses is really easy – but sending mails from those any-aliases was not possible at all unless every sender address was added as an explicit alias to the database.

By this change in the database query for allowed sender addresses, the first finding `not NULL` (see [`SELECT COALESCE`](https://www.w3schools.com/sql/func_sqlserver_coalesce.asp) for how it works) – either an exact alias `mailbox@domain.tld` or the wildcard alias `@domain.tld` will be allowed to send mails as the given address ... without the need of explicit definition within the database.

I hope, you find this change useful and consider adding this feature to the official project =)

### DIFF
--- a/.github/workflows/close_old_issues_and_prs.yml
+++ b/.github/workflows/close_old_issues_and_prs.yml
@@ -14,7 +14,7 @@ jobs:
       pull-requests: write
     steps:
       - name: Mark/Close Stale Issues and Pull Requests 🗑️
-        uses: actions/stale@v5.1.0
+        uses: actions/stale@v5.1.1
         with:
           repo-token: ${{ secrets.STALE_ACTION_PAT }}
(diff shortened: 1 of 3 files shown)


andryyy (commented):
Looks good. :)

andryyy (approved):
Looks good!

andryyy:
Forgot to approve. 

@DerLinkman We should not forget to increase the image version.

DerLinkman:
@macwinnie can you increase the Image Tag inside of the docker-compose.yml for Postfix +1?

macwinnie:
@DerLinkman done the `docker-compose.yml` update.

DerLinkman (approved):

PR #4716, opened 2022-08-13
[web] Fix: dns_diagnostics.php SPF DNS lookup loop

### LINKED ISSUE #4715: dns_diagnostics.php SPF DNS lookup loop
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... I understand that not following the below instructions will result in immediate closure and/or deletion of my issue.
- [X] ... I have understood that this bug report is dedicated for bugs, and not for support-related inquiries.
- [X] ... I have understood that answers are voluntary and community-driven, and not commercial support.
- [X] ... I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

### The problem:
When using dns diagnostics to check SPF records, a possible loop can occur.
If the domain has an SPF record pointing to another domain that has an SPF record referencing the original via `include:`.
This causes an infinite stall.

### Example: 
- call dns_diagnostics.php?domain=example.com
- example.com has an SPF record with `include:mail.example.com`
- mail.example.com has an SPF record with `include:example.com`

Line 50 of `data/web/inc/spf.inc.php` only checks for the current domain `$check_domain != $domain` but the function below passes in the currently checking domain **without** context of what originally called it, so a loop occurs in this case.

I understand in this situation, _technically_ the SPF records may be at fault/invalid, but I found this bug while moving an existing mail system over to mailcow that had been working this way for years, and so may cause frustration to troubleshoot for others in a similar situation.

There are no logs generated (or needed) for this action as it simply keeps looping forever, stalling the current php-fpm session, at least on my system.

### The fix:
I appended `, $orig_domain = ""` to the `get_spf_allowed_hosts(..` function in `data/web/inc/spf.inc.php` and also set every call within `spf.inc.php` to pass `$check_domain` to this new argument. Then by adding an additional check on line 50 for this original domain ` && $orig_domain != $domain` it correctly stops after one loop as we're back on the original domain.

~~I may open a pull request for the fix, but it's a hacky workaround and am not sure if it handles edge cases or effects anything else.~~ Opened #4716 
Any better way to solve this is welcome.

Thank you!

### Logs

```bash
N/A
```

### Steps to reproduce

- example.com has an SPF record with `include:mail.example.com`
- mail.example.com has an SPF record with `include:example.com`
- call dns_diagnostics.php?domain=example.com
- It will hang indefinitely 

### System information

N/A

ualnwe:
### The problem:
When using dns diagnostics to check SPF records, a possible loop can occur.
If the domain has an SPF record pointing to another domain that has an SPF record referencing the original via `include:`.
This causes an infinite stall.

### Example: 
- call dns_diagnostics.php?domain=example.com
- example.com has an SPF record with `include:mail.example.com`
- mail.example.com has an SPF record with `include:example.com`

### The fix:
I appended `, $original_domain = ""` to the `get_spf_allowed_hosts(..` function in `data/web/inc/spf.inc.php` and also set every call within `spf.inc.php` to pass `$check_domain` to this new argument. Then by adding an additional check on line 50 for this original domain ` && $original_domain != $domain` it correctly stops after one loop as we're back on the original domain.

Closes #4715

### DIFF
--- a/data/web/inc/spf.inc.php
+++ b/data/web/inc/spf.inc.php
@@ -26,7 +26,7 @@ function get_spf_allowed_hosts($check_domain, $expand_ipv6 = false) {
 				$mod = explode('=', $mech);
 				if ($mod[0] == 'redirect') // handle a redirect
 				{
-					array_push($checked_domains, $check_domain);
+					array_push($checked_domains, $mod[1]);
 					$hosts = get_spf_allowed_hosts($mod[1],true);
 					return $hosts;
 				}
@@ -52,7 +52,7 @@ function get_spf_allowed_hosts($check_domain, $expand_ipv6 = false) {
 				$new_hosts = array();
         if ($mech == 'include' && $check_domain != $domain && !in_array($check_domain, $checked_domains)) // handle an inclusion
 				{
-					array_push($checked_domains, $check_domain);
+					array_push($checked_domains, $domain);
 					$new_hosts = get_spf_allowed_hosts($domain);
 				}
 				elseif ($mech == 'a') // handle a mechanism
--- a/data/web/inc/spf.inc.php
+++ b/data/web/inc/spf.inc.php
@@ -1,6 +1,6 @@
 <?php
 error_reporting(0);
-function get_spf_allowed_hosts($check_domain, $expand_ipv6 = false) {
+function get_spf_allowed_hosts($check_domain, $expand_ipv6 = false, $original_domain = "") {
 	$hosts = array();
 	
 	$records = dns_get_record($check_domain, DNS_TXT);
@@ -24,7 +24,7 @@ function get_spf_allowed_hosts($check_domain, $expand_ipv6 = false) {
 				$mod = explode('=', $mech);
 				if ($mod[0] == 'redirect') // handle a redirect
 				{
-					$hosts = get_spf_allowed_hosts($mod[1],true);
+					$hosts = get_spf_allowed_hosts($mod[1],true, $check_domain);
 					return $hosts;
 				}
 			}
@@ -47,9 +47,9 @@ function get_spf_allowed_hosts($check_domain, $expand_ipv6 = false) {
 				}
 				
 				$new_hosts = array();
-        if ($mech == 'include' && $check_domain != $domain) // handle an inclusion
+        if ($mech == 'include' && $check_domain != $domain && $original_domain != $domain) // handle an inclusion
 				{
-					$new_hosts = get_spf_allowed_hosts($domain);
+					$new_hosts = get_spf_allowed_hosts($domain, false, $check_domain);
 				}
 				elseif ($mech == 'a') // handle a mechanism
 				{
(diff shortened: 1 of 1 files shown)


mkuron (changes requested):
Your change only checks for a loop back to the original domain -- before it can be merged, it needs to be extended to detect _all_ loops, including those that loop back to intermediate domains. Also, your pull request contains some unrelated changes that need to be removed.

ualnwe:
This should now keep track of checked domains and detect a loop using a global array.
There's probably a better way to do this, but this works in my testing :sweat_smile: 
Apologies for the unrelated changes, and I may have selected the wrong branch.
Thank you DerLinkman. Hopefully these both are now resolved.

data/web/inc/spf.inc.php
```diff
@@ -1,6 +1,8 @@
 <?php
 error_reporting(0);
-function get_spf_allowed_hosts($check_domain, $expand_ipv6 = false, $original_domain = "") {
+$checked_domains = array();
+function get_spf_allowed_hosts($check_domain, $expand_ipv6 = false) {
+	global $checked_domains;
```
mkuron: Please don't use global variables. This fill break the loop detection when multiple SPF checks are performed on the same page.

mkuron (changes requested):

data/web/inc/spf.inc.php
ualnwe: Sorry, for some reason I didn't think about just passing the array back into the function.
I've changed it to pass in a new optional argument, the array containing all previously checked include or redirect SPF domain records.

mkuron:
Thanks, that's much better. Sorry for keeping requesting changes, but I think it might make sense to display a warning on the DNS diagnostics page if an SPF loop is detected.

I checked RFC 7208 and it actually says this:
> Some mechanisms and modifiers (collectively, "terms") cause DNS
   queries at the time of evaluation, and some do not.  The following
   terms cause DNS queries: the "include", "a", "mx", "ptr", and
   "exists" mechanisms, and the "redirect" modifier.  SPF
   implementations MUST limit the total number of those terms to 10
   during SPF evaluation, to avoid unreasonable load on the DNS.  If
   this limit is exceeded, the implementation MUST return "permerror".

So simply ignoring looping redirects is not RFC compliant. Instead, we should actually treat this as an error condition. Depending on the other server's SPF implementation, a permerror may either be considered an SPF pass or an SPF fail, so you won't be able to reliably deliver emails if your SPF has a loop.

DerLinkman:
Is still a thing?

mkuron:
Yes, I'd like to get that merged, but @ualnwe never addressed my latest comment. Sorry that it took multiple rounds of review (I hope you didn't lose interest because of that, @ualnwe), I should have looked at the RFC first.

ualnwe:
Apologies for the lack of updates, I've been rather busy with other things.
I'd still very much like to fix this, and to the correct standard as mentioned re: being RFC compliant

milkmaker:

Issue #4695, opened 2022-07-23
update.sh is broken since docker-compose v2 requirement


domrim:
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... I understand that not following the below instructions will result in immediate closure and/or deletion of my issue.
- [X] ... I have understood that this bug report is dedicated for bugs, and not for support-related inquiries.
- [X] ... I have understood that answers are voluntary and community-driven, and not commercial support.
- [X] ... I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

### Description

Since the requirement of docker-compose v2 (which came way earlier than December 2022, but that's another topic) the `update.sh` script is broken. It tries to call `docker-compose` but this executable doesn't exist anymore with docker-compose v2. This only works if you have also installed [`compose-switch`](https://github.com/docker/compose-switch).

I believe this was introduced with the commit 7166696aa2682617836c254e9243a3120a9243d0 by @DerLinkman  

### Steps to reproduce

```bash
root@host:/opt/mailcow-dockerized# ./update.sh
Cannot find docker-compose Standalone.
Please install it manually regarding to this doc site: https://mailcow.github.io/mailcow-dockerized-docs/i_u_m/i_u_m_install/
```

### System information

| Question | Answer |
| --- | --- |
| My operating system | Debian 11 |
| Is Apparmor, SELinux or similar active? | apparmor |
| Virtualization technology (KVM, VMware, Xen, etc - **LXC and OpenVZ are not supported** | KVM |
| Server/VM specifications (Memory, CPU Cores) | 2 Cores, 4GB Ram |
| Docker version (`docker version`) | 20.10.17 |
| docker-compose version (`docker-compose version`) | 2.6.0 |
| mailcow version (```git describe --tags `git rev-list --tags --max-count=1` ```) | 2022-07 |
| Reverse proxy (custom solution) | no |

domrim:
Small update:

I just read your [installation guide for docker-compose](https://mailcow.github.io/mailcow-dockerized-docs/i_u_m/i_u_m_install/) and understand now, that the script should work if one installed docker-compose like you want it.
Why do you diverge here from the [recommended method](https://github.com/docker/compose#linux) from docker?

[...]

PR #4723, opened 2022-08-22
[Update.sh] User Decision of which Compose Version they want to use (min v2)

DerLinkman:
This PR will fix the force of the standalone Version from Docker Compose which has been introduced with 2022-06 (https://github.com/mailcow/mailcow-dockerized/issues/4695).

It´ll read the value from a new Variable inside the mailcow.conf which defines the used Docker Compose Version.

On first run (generate-config) or first update with this change it will automatically detect the available Docker Compose Versions on the Maschine.

**Notice: It will always prefer the native Plugin from Docker itself, not the Standalone Version**

The user is able to change that variable manually to say what mailcow should use internally.

### DIFF
--- a/generate_config.sh
+++ b/generate_config.sh
@@ -135,8 +135,6 @@ else
   SKIP_SOLR=n
 fi
 
-if []
-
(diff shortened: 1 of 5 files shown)


DerLinkman:
Implemented into: https://github.com/mailcow/mailcow-dockerized/pull/4725

Issue #4743, opened 2022-09-04
Quarantine notifications no longer sent with 2022-08(a)


tmkis2:
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... I understand that not following the below instructions will result in immediate closure and/or deletion of my issue.
- [X] ... I have understood that this bug report is dedicated for bugs, and not for support-related inquiries.
- [X] ... I have understood that answers are voluntary and community-driven, and not commercial support.
- [X] ... I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

### Description

Quarantine Mails still works but no notification is sent to users mailbox regardless of setting it to 1 hour, 24 hours or something else

### Logs

```bash
don't know which logs to check for this
```

### Steps to reproduce

Enable Quarantine notifications to 1 hour - get some quarantined messages - wait

### System information

| Question | Answer |
| --- | --- |
| My operating system | Ubuntu 20.04 LTS |
| Is Apparmor, SELinux or similar active? | NO |
| Virtualization technology (KVM, VMware, Xen, etc - **LXC and OpenVZ are not supported** | KVM |
| Server/VM specifications (Memory, CPU Cores) | 8 GB RAM/4 Cores |
| Docker version (`docker version`) | 20.10.17 |
| docker-compose version (`docker-compose version`) | v2.10.2 |
| mailcow version (```git describe --tags `git rev-list --tags --max-count=1` ```) | 2022-08a |
| Reverse proxy (custom solution) | NO |

Output of `git diff origin/master`, any other changes to the code? If so, **please post them**:
```
diff --git a/create_cold_standby.sh b/create_cold_standby.sh
index 924339af..9d7bb0ed 100755
--- a/create_cold_standby.sh
+++ b/create_cold_standby.sh
@@ -2,6 +2,6 @@
 
 export REMOTE_SSH_KEY=/root/.ssh/id_rsa
 export REMOTE_SSH_PORT=22
-export REMOTE_SSH_HOST=my.remote.host
+export REMOTE_SSH_HOST=5.180.192.196
 
 /opt/mailcow-dockerized/helper-scripts/_cold-standby.sh
diff --git a/data/conf/postfix/main.cf b/data/conf/postfix/main.cf
[...]

PR #4758, opened 2022-09-19
Update quarantine_notify.py charset

MAGICCC:
Updates `quarantine_notify.py` to use utf8mb4 charset and utf8mb4_general_ci collation
Tested on mariadb:10.6 and works.
I sent a bogus mail from a server via telnet which classifies the mail as spam and thus rspamd is putting it in the quarantine. Triggering `quarantine_notify.py` by ofelia will deliver the quarantine mail to the inbox

Fixes https://github.com/mailcow/mailcow-dockerized/issues/4743

New image **mailcow/dovecot:1.20** on Dockerhub available


DerLinkman (approved):
Looks good

PR #4759, opened 2022-09-21
[Netfilter] Record IP address in F2B_ACTIVE_BANS instead of the network block

naudster:
Minor change that can improve the whois output of watchdog's IP ban email in certain cases.

Currently watchdog receives the network block that was banned in the F2B_ACTIVE_BANS message. However a whois lookup of the block can fail to return a result if a larger/smaller block is what's actually registered. Example:

IP that caused the ban: 20.199.179.218
Fail2ban parameter f2b_netban_ipv4: 24
Resulting banned network:  20.199.179.0/24
```
$ whois 20.199.179.0/24
No match found for r + = 20.199.179.0/24.
```

Changing F2B_ACTIVE_BANS to contain the actual IP means watchdog can perform a successful query and include the output in its email:
```
$ whois 20.199.179.218 
<snip>
NetRange:       20.192.0.0 - 20.255.255.255
CIDR:           20.192.0.0/10
NetName:        MSFT
NetHandle:      NET-20-192-0-0-1
Parent:         NET20 (NET-20-0-0-0-0)
NetType:        Direct Allocation
```

Note that the IP is included in the the email subject so in this example it changes from:
`Subject: Wed Sep 21 10:22:36 AEST 2022 - IP ban: 20.199.179.0/24`
to:
`Subject: Wed Sep 21 10:22:36 AEST 2022 - IP ban: 20.199.179.218`

Personally I prefer that because it identifies the actual source of the attempt.

The other changes in the commit simply strip some superfluous whitespace. That was the result of an vim plugin that auto stripped the space. It wasn't an intended change but it cleans it up a bit I guess. :)

### DIFF
--- a/data/Dockerfiles/netfilter/server.py
+++ b/data/Dockerfiles/netfilter/server.py
@@ -206,7 +206,7 @@ def ban(address):
         rule.target = target
         if rule not in chain.rules:
           chain.insert_rule(rule)
-    r.hset('F2B_ACTIVE_BANS', '%s' % net, cur_time + BAN_TIME)
+    r.hset('F2B_ACTIVE_BANS', '%s' % address, cur_time + BAN_TIME)
   else:
     logWarn('%d more attempts in the next %d seconds until %s is banned' % (MAX_ATTEMPTS - bans[net]['attempts'], RETRY_WINDOW, net))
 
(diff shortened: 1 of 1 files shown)


naudster:
Sorry, just realised this breaks the unban functionality from the Web UI. [data/web/inc/functions.fail2ban.inc.php](https://github.com/mailcow/mailcow-dockerized/blob/master/data/web/inc/functions.fail2ban.inc.php) references F2B_ACTIVE_BANS for display and action in the UI:
image
Function is broken when you banned a whole block but then attempt to unban just the IP.

Should probably leave F2B_ACTIVE_BANS alone and introduce a separate redis topic that contains the actual IP. Let me know if you have some advice.

PR #4763, opened 2022-09-26
[API] Update swagger version

ntimo:


ntimo:
This should also be backported to nightly

DerLinkman (approved):

PR #4806, opened 2022-10-19
[Backup] Swapped PIGZ instead of gzip (allow Threading)

### LINKED ISSUE #4805: Parallelize backup_and_restore.sh script (with pigz)
### Summary

Opening this request as Niklas suggested (thank you). This would introduce a way to utilize multiple cores during backup process. I can submit a PR but it's a few lines of change and it pretty much depends on how you guys would prefer this.

**This would not be default**, since pigz is not available on many systems by default. Pretty much all distributions I've ever seen ship one in their repos, but it's not a default install. pigz is a drop in gzip replacement. Even though it would not be the default, the script could "advertise it", inform the user that it exists.

**What would need to be added/changed:**
- `pigz` instead of `gzip` in the script in the IF branch (drop-in replacement)
- add command line / env var support for thread limiting (see below)
- the script could do a simple check IF `pigz` command exists just to have an additional safeguard against users. can never hurt.

**How it would be enabled:**
- It can be done via ENV VAR
- It can be done via command line argument
- 
Then simply an IF branch. I would honestly go with ENV VAR because the script already goes like (in my case):
`echo "$OUTPUT/" | /opt/mailcow-dockerized/helper-scripts/backup_and_restore.sh backup all`

So complicating that with more args would be messy I think. I think.

**One example implementation:**
- Add support for ENV VAR.
- "MCBT=" (yeah, sue me, MailCow Backup Threads)
- IF it exists AND above 1... ie.:
- MCBT=2: call `pigz` and `-p 2`
- MCBT=5: call `pigz` and `-p 5`

**PROS/CONS**
Pros:
- time needed for a backup would be cut down a lot. `pigz` scales very nicely.
- pigz supports thread limitation ( it's `-p threads`, see: https://linux.die.net/man/1/pigz )
- pigz is a small install and available on almost any distro.
Cons:
- It can eat up the whole CPU if given enough threads. But then again a user could remove their whole / filesystem as well. So not really a concern I think.
- Compression is not as great as single-threaded gzip. But FIXME on this one. I could not find any graphs, I just remember reading something about this months ago. The difference is not huge though.
- Few more lines in the script code.

### Motivation

This would speed up the backup process significantly.

### Additional context

_No response_

DerLinkman:
See Issue: https://github.com/mailcow/mailcow-dockerized/issues/4805

### DIFF
--- a/data/Dockerfiles/backup/Dockerfile
+++ b/data/Dockerfiles/backup/Dockerfile
@@ -1,0 +1,3 @@
+FROM debian:bullseye-slim
+
+RUN apt update && apt install pigz
--- a/helper-scripts/backup_and_restore.sh
+++ b/helper-scripts/backup_and_restore.sh
@@ -54,8 +54,6 @@ COMPOSE_FILE=${SCRIPT_DIR}/../docker-compose.yml
 ENV_FILE=${SCRIPT_DIR}/../.env
 THREADS=$(echo ${THREADS:-1})
 
-echo $THREADS
-
 if [ ! -f ${COMPOSE_FILE} ]; then
   echo "Compose file not found"
   exit 1
--- a/helper-scripts/backup_and_restore.sh
+++ b/helper-scripts/backup_and_restore.sh
@@ -48,6 +48,16 @@ else
   fi
 fi
 
+if [[ -z ${THREADS} ]]; then
+  read -ep "How many threads do you want to use? (Default $(echo "CPU threads: $(grep -c processor /proc/cpuinfo)" | grep -E -o "[0-9]+")):" THREADS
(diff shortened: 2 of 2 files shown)


DerLinkman:
Can someone of you guys check? Simply checkout the backup restore script and run it.

with the THREADS Variable you can pass the threads count to the script to skip the prompt.

DerLinkman:
I´m still not sure if i keep the Thread Prompt... it will mean that all current cronjobs have to modified. And i don´t know if i like that...

MAGICCC:
> I´m still not sure if i keep the Thread Prompt... it will mean that all current cronjobs have to modified. And i don´t know if i like that...

Maybe change that if not $THREADS is given it will defaults to 1 thread and just print a warning that you can set $THREADS

DerLinkman:
> 

That´s a great idea. Yeah i´ve already implemented the version without the prompt but without any notice it didn´t feld right. But with the message it sounds way better!

PR #4815, opened 2022-10-25
[redis] Bump Redis to version 7

ethrgeist:
Hi,

for all intends and purposes in this context, Redis 7 is a drop-in replacements, which could net a few performance gains.


DerLinkman:
Did you test it?

Won´t merge it for 2022-10 but for 2022-11 maybe-

Needs more tests (do it myself also).

ethrgeist:
I did test it on my instances, i also use Redis on a lot of other projects.

As Mailcow does not use any configuration but the default Redis config, this is also full backwards compatible with dumps from version 6, so this upgrade does not even required to drop the volume, so all data is kept.

DerLinkman (approved):
Looks good to me.
All components seems to be intact and usable as before.

Issue #4839, opened 2022-11-05
upgrade phpfpm to 8.1


BigMichi1:
### Summary

upgrade the phpfpm image from 8.0 to 8.1

### Motivation

according to https://www.php.net/supported-versions 8.0 reaches its EOL in about 3 weeks.

PR #4854, opened 2022-11-08
[PHP] Polish dockerfile

BigMichi1:
includes also #4839 because of `--with-avif` for `gd` configure command (is not available in 8.0)

contains the following adjustments:
    - upgrade APCu to 5.1.22
    - use PECL package for mailparse instead of git clone (3.1.4 is the latest one available and sice then no changes on master branch)
    - split PECL commands into separate ones (according to https://hub.docker.com/_/php this is the recommended way)
    - add missing configure options for gd extension to include webp, xpm and avif
    - specify composer version to be installed
    - cleanup more dev dependencies


milkmaker:
Thanks for contributing!

PR #4859, opened 2022-11-13
switch update.sh/check_online_status() from ping to curl to make it p…

mitch-geht-ab:
…roxy ready

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -3,9 +3,9 @@
 ############## Begin Function Section ##############
 
 check_online_status() {
-  CHECK_ONLINE_IPS=(1.1.1.1 9.9.9.9 8.8.8.8)
-  for ip in "${CHECK_ONLINE_IPS[@]}"; do
-    if timeout 3 ping -c 1 ${ip} > /dev/null; then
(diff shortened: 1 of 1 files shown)


update.sh
```diff
@@ -3,9 +3,9 @@
 ############## Begin Function Section ##############
 
 check_online_status() {
-  CHECK_ONLINE_IPS=(1.1.1.1 9.9.9.9 8.8.8.8)
-  for ip in "${CHECK_ONLINE_IPS[@]}"; do
-    if timeout 3 ping -c 1 ${ip} > /dev/null; then
+  CHECK_ONLINE_DOMAINS=('https://github.com')
```
DerLinkman: Maybe add one or two domains here as well.

google.com and hub.docker.com i would say.

The rest looks good

DerLinkman (changes requested):
See comment

update.sh
mitch-geht-ab: Had added only the used domains by intention.

FreddleSpl0it: i would at least add hub.docker.com as we pulling images from there.

DerLinkman (approved):
Looks fine now!

Issue #4862, opened 2022-11-15
Pushover variables


bluewalk:
### Summary

Currently the following variables exist for the Pushover integration: `{SUBJECT}, {SENDER}`.

More variables would make this feature even more powerful; eg `{SENDER_NAME}` (name of the sender in from tag), `{FIRST_PARAGRAPH}` for a preview etc.

### Motivation

As on a mobile phone many mail apps don't have real time notifications without having to share your login details; the Pushover integration works wonders.

But more information would come in very handy. {SENDER_NAME} for example as some mail from eg; `mailtoblabla+bounce@newsletter.somedomain.com` which isn't as clear as "Company X".

The first paragrtaph (just text) would make the notification even more detailed.

bluewalk:
I see that getting the body via rspam is only possible by passing the whole email (formatter = default) and then having to parse that email... this seems too much work.

However getting the name of the sender is easier (formatter = json) will post a json to pushover.php which contains more information than the headers do. Eg `header_from` which is an array containing the sender in the format `Name <email@domain.com>`.

bluewalk:
Just another improvement, the ability to choose a sound (which is a parameter in the request).

bluewalk:
I have created a PR with the improvements mentioned above 👍🏻

bluewalk:
I see my changed have been merged into staging, hence closing.

PR #4864, opened 2022-11-17
Pushover improvements

bluewalk:
Implementation of my suggested Pushover improvements, mentioned in #4862

This adds
* Two new variables SENDER_ADDRESS and SENDER_NAME
* The ability to choose the sound for the notification


DerLinkman:
There we go :)

Sorry for messing with your branch. Now it only has your changes and not ours in it :)

DerLinkman (changes requested):
In the init_db_inc.php file please increase the date at the beginning of the file as well. Otherwise no Database update is performed.

bluewalk:
No worries, happy to know it’s getting merged so others can enjoy it 

> In the init_db_inc.php file please increase the date at the beginning of the file as well. Otherwise no Database update is performed.

Not sure if the database change is required, seems to be more like a migration step of existing data. If it’s required for a new attribute I will change it otherwise I will remove it.

bluewalk:
Don't merge yet, found an issue in the parsing of the email address when it comes to subdomains. Will fix later today.

bluewalk:
There we go, removed the regex which was not needed as we already have the senders address from the rspamd headers.

bluewalk:
Just noticed the rspam from address sometimes differs from the actual address used in the header `Name <address>`. Eg Twitter has “notify@twitter.com” in its mail header while rspam comes back with something@bounce.twitter.com. Will do some improvements on that later 👍🏻

FreddleSpl0it:
SENDER_ADDRESS might be a duplicate with the aleady existing SENDER variable or am i wrong?

bluewalk:
> SENDER_ADDRESS might be a duplicate with the aleady existing SENDER variable or am i wrong?

Unfortunately not always the case. See earlier comment about Twitter. We can however remove the original SENDER as this could be a bounce email address. The from header of the email (part of the JSON body) is always correct and is what the email client shows.

FreddleSpl0it:
Yes, i read your last comment not properly. I have tested your changes and everything works great :) thanks for your contribution.

FreddleSpl0it (approved):

PR #4898, opened 2022-12-13
2022-11b

### LINKED ISSUE #4897: SoGO / CalDav: Personal Calendar issue on Mac OS 13
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... I understand that not following the below instructions will result in immediate closure and/or deletion of my issue.
- [X] ... I have understood that this bug report is dedicated for bugs, and not for support-related inquiries.
- [X] ... I have understood that answers are voluntary and community-driven, and not commercial support.
- [X] ... I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

### Description

There is currently a bug in mac OS 13 (?) which causes the Calendar.app to ignore calendars, that begin with lowercase letters (in their internal identifier).

This seems to be the case for the default (unremovable) calendar in SoGO.

More information can be found in their respective issue: https://bugs.sogo.nu/view.php?id=5639

It would be good to discuss, if we want to pull in 5.8.0 nightly, with this fix included.
There is only one other small commit that would be included and it seems reasonable to consider this stable, but im open to your suggestions.

DerLinkman:
Fixes https://github.com/mailcow/mailcow-dockerized/issues/4897.
Increased Timeout Value in update.sh + updated README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,5 @@
 # mailcow: dockerized - 🐮 + 🐋 = 💕
 
-## We stand with 🇺🇦
-
 [![Mailcow Integration Tests](https://github.com/mailcow/mailcow-dockerized/actions/workflows/integration_tests.yml/badge.svg?branch=master)](https://github.com/mailcow/mailcow-dockerized/actions/workflows/integration_tests.yml)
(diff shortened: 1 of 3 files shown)


milkmaker:
Thanks for contributing!

PR #4899, opened 2022-12-13
Update bulk_header.map

mhupfauer:
AWeber is a massive Mail as a Service provider which is used by many legitimate corporations and should not be handled negatively by default.


milkmaker:
Thanks for contributing!

PR #4900, opened 2022-12-16
fix incorrect render value

ethrgeist:
`text` is not valid for `render`, it is `plain text` according to the ENUM schema: <https://json.schemastore.org/github-issue-forms.json>


milkmaker:
Thanks for contributing!

PR #4903, opened 2022-12-21
Add Message-ID to pushover

Der-Jan:
Having Message-ID in the push message, allows to open mail in Mail.app via message:<MSG_ID>


milkmaker:
Thanks for contributing!

protree:
Thank you for this feature! It might be even more useful to include the message id in the supplementary URL as described here: https://pushover.net/api#urls

PR #4941, opened 2022-12-28
root@Mail:~/mailcow/mailcow-dockerized# docker-compose pull

florianw3ndel:
ERROR: The Compose file './docker-compose.yml' is invalid because: services.nginx-mailcow.ports is invalid: Invalid port ":443:443", should be [[remote_ip:]remote_port[-remote_port]:]port[/protocol] services.nginx-mailcow.ports is invalid: Invalid port ":80:80", should be [[remote_ip:]remote_port[-remote_port]:]port[/protocol]

Cant compose without BINDing IPs


milkmaker:
Thanks for contributing!

DerLinkman:
You use docker-compose v2 right?

This is not a global fix

PR #4967, opened 2023-01-06
[Web] Implemented SSO for domain admins

feldsam:
Hello, I am implemented token generation via API for domain admins SSO. This function is usefull, if you want/need redirect and login customer from external system directly to mailcow. Token expiration is set to just 30 seconds and there is garbage collection also.

### DIFF
--- a/data/web/inc/functions.domain_admin.inc.php
+++ b/data/web/inc/functions.domain_admin.inc.php
@@ -405,3 +405,64 @@ function domain_admin($_action, $_data = null) {
     break;
   }
 }
+function domain_admin_sso($_action, $_data) {
+  global $pdo;
+
+  switch ($_action) {
+    case 'check':
+      $token = $_data;
+
+      $stmt = $pdo->prepare("SELECT `t1`.`username` FROM `da_sso` AS `t1` JOIN `admin` AS `t2` ON `t1`.`username` = `t2`.`username` WHERE `t1`.`token` = :token AND `t1`.`created` > DATE_SUB(NOW(), INTERVAL '30' SECOND) AND `t2`.`active` = 1 AND `t2`.`superadmin` = 0;");
+      $stmt->execute(array(
+        ':token' => preg_replace('/[^a-zA-Z0-9-]/', '', $token)
+      ));
+      $return = $stmt->fetch(PDO::FETCH_ASSOC);
+      return empty($return['username']) ? false : $return['username'];
+    case 'issue':
+      if ($_SESSION['mailcow_cc_role'] != "admin") {
+        $_SESSION['return'][] = array(
+          'type' => 'danger',
+          'log' => array(__FUNCTION__, $_action, $_data),
+          'msg' => 'access_denied'
+        );
+        return false;
+      }
+
+      $username = $_data['username'];
+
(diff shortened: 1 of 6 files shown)


data/web/inc/functions.domain_admin.inc.php
```diff
@@ -1,407 +1,459 @@
...
+    case 'check':
+      $token = $_data;
+
+      $stmt = $pdo->prepare("SELECT `t1`.`username` FROM `da_sso` AS `t1` JOIN `admin` AS `t2` ON `t1`.`username` = `t2`.`username` WHERE `t1`.`token` = :token AND `t1`.`created` > DATE_SUB(NOW(), INTERVAL '30' SECOND) AND `t2`.`active` = 1 AND `t2`.`superadmin` = 0;");
+      $stmt->execute(array(
+        ':token' => preg_replace('/[^a-zA-Z0-9-]/', '', $token)
+      ));
+      $return = $stmt->fetch(PDO::FETCH_ASSOC);
+      return empty($return['username']) ? false : $return['username'];
+    case 'issue':
```
FreddleSpl0it, line 421: You need to limit this function to admins or access with API-Key. Currently everyone who knows or can successfull guess a domain-admins username can create a SSO Key and gain access to the UI

FreddleSpl0it (changes requested):

data/web/inc/functions.domain_admin.inc.php
feldsam: Hi, but issue method is called only via API here https://github.com/mailcow/mailcow-dockerized/blob/45d8c1720e04b93291d5e66ec73254ceb239d686/data/web/json_api.php#L294

FreddleSpl0it: yes and this is not restricted. i could issue a token with
```
curl https://mailcow-url.tld/api/v1/add/sso/domain-admin -H 'Content-Type: application/json' -d '{"username":"dadmin"}'
```

FreddleSpl0it: here you can see how adding domain_admins function handles this https://github.com/mailcow/mailcow-dockerized/blob/71cc607de65e8875f7e76c9a482521bfa6ad64fb/data/web/inc/functions.domain_admin.inc.php#L18

feldsam: I see, I supposted this permissions handling is already done in json_api.php file, because there is some check for RO/RW access. Looks like, there is check only for IP. Ok, so I add check to domain_admin_sso function. Thanks for pointing this out

feldsam: Fixed and tested:

```
# curl https://maildev.feldhost.cz/api/v1/add/sso/domain-admin -H 'Content-Type: application/json' -d '{"username":"feldsamm"}'

[{"type":"danger","log":["domain_admin_sso","issue",{"username":"feldsamm"}],"msg":"access_denied"}]
```

feldsam:
Hi, I also update branch from staging, so there is no merge conflict.

FreddleSpl0it (approved):

PR #5049, opened 2023-02-05
[Netfilter] Add support for nftables

amorfo77:
New linux distributions are increasingly adopting nftables as the default framework for packet filtering.

This PR tries to add support for nftables, in addition to the existing support for iptables, for the moment, the user decides which backend to use by using an environment variable that will run one or the other.
It can be enabled using `USE_NFTABLES` in the mailcow.conf file, the default value if not specified is 'n'.

I've tested it on the following distributions
Ubuntu 22.04LTS (currently in my production server for two months)
Debian 11
Fedora 37
during the next days I will test other distributions.


mkuron (changes requested):
Copying the entire server.py is not a good idea. Most of the code is the same in server-nft.py and server.py, so they would diverge over time as future changes might be made to one but not the other. Also, this shouldn't be configured by an environment variable, but automatically detected at container startup.

amorfo77:
Hi @mkuron I agree with you, I will try to implement the autodetect part first, I have only found this method to autodetect which backend is being used. 
[https://unix.stackexchange.com/questions/588998/check-whether-iptables-or-nftables-are-in-use](https://unix.stackexchange.com/questions/588998/check-whether-iptables-or-nftables-are-in-use)
if it's ok with you i will implement this method

Twilek-de:
Hmm this seems to be a non trivial problem. Tailscale seems to have failed at it until now. Maybe an environment variable IS the way to go in the end...
https://github.com/tailscale/tailscale/issues/5621

Twilek-de:
If I execute `iptables -V` I get `iptables v1.8.9 (nf_tables)  `. Does the "nf_tables" indicate that really nft is used?

Update:
Here is the answer https://developers.redhat.com/blog/2020/08/18/iptables-the-two-variants-and-their-relationship-with-nftables#

DerLinkman:
Hi there,

thanks for contributing.

Right know in this state at least it won't work on a default servercow mailcow setup which is our usual test environment.

It's hitting a error whenever it tries to initialise a SNAT Rule:

```
mailcowdockerized-netfilter-mailcow-1  | 2023-03-09T13:50:36.767406156Z Using Nftables backend
mailcowdockerized-netfilter-mailcow-1  | 2023-03-09T13:50:36.770094811Z Clearing all bans
mailcowdockerized-netfilter-mailcow-1  | 2023-03-09T13:50:36.774748812Z Initializing mailcow netfilter chain
mailcowdockerized-netfilter-mailcow-1  | 2023-03-09T13:50:36.778814893Z MAILCOW ip chain created successfully.
mailcowdockerized-netfilter-mailcow-1  | 2023-03-09T13:50:36.810081617Z MAILCOW ip6 chain created successfully.
mailcowdockerized-netfilter-mailcow-1  | 2023-03-09T13:50:36.859441258Z Watching Redis channel F2B_CHANNEL
mailcowdockerized-netfilter-mailcow-1  | 2023-03-09T13:50:46.864112107Z Error in ip6 forward chain order: MAILCOW on position 1, restarting container
mailcowdockerized-netfilter-mailcow-1  | 2023-03-09T13:50:46.871423300Z Remove rule for source network 172.22.1.0/24 to SNAT target 178.251.229.2 from POSTROUTING chain with handle 209
mailcowdockerized-netfilter-mailcow-1  | 2023-03-09T13:50:46.873027776Z Clearing all bans
mailcowdockerized-netfilter-mailcow-1  | 2023-03-09T13:50:46.876846330Z Clear completed: ip
mailcowdockerized-netfilter-mailcow-1  | 2023-03-09T13:50:46.880183203Z Clear completed: ip6
mailcowdockerized-netfilter-mailcow-1 exited with code 2
```

EDIT: 

> It did work, BUT it crashes on the first run whenever it deletes a SNAT rule.
> 
> That shouldn't be in the final version for production. If it needs to be restarted then the error message should be rewritten that no one is opening a ticket about that.
> 

This error message here appears on each first run without a SNAT Rule, afterwards it doesn't.

```
mailcowdockerized-netfilter-mailcow-1  | 2023-03-09T13:52:03.755653085Z Using Nftables backend
mailcowdockerized-netfilter-mailcow-1  | 2023-03-09T13:52:03.758667693Z Clearing all bans
mailcowdockerized-netfilter-mailcow-1  | 2023-03-09T13:52:03.764007041Z Initializing mailcow netfilter chain
mailcowdockerized-netfilter-mailcow-1  | 2023-03-09T13:52:03.768013990Z MAILCOW ip chain created successfully.
mailcowdockerized-netfilter-mailcow-1  | 2023-03-09T13:52:03.819056088Z MAILCOW ip6 chain created successfully.
mailcowdockerized-netfilter-mailcow-1  | 2023-03-09T13:52:03.856275307Z Watching Redis channel F2B_CHANNEL
mailcowdockerized-netfilter-mailcow-1  | 2023-03-09T13:52:13.868760303Z Error in ip6 forward chain order: MAILCOW on position 1, restarting container
mailcowdockerized-netfilter-mailcow-1  | 2023-03-09T13:52:13.873801091Z Clearing all bans
mailcowdockerized-netfilter-mailcow-1  | 2023-03-09T13:52:13.878135049Z Clear completed: ip
mailcowdockerized-netfilter-mailcow-1  | 2023-03-09T13:52:13.881450803Z Clear completed: ip6
mailcowdockerized-netfilter-mailcow-1 exited with code 2
```

Summary: If it needs to be restarted due to not a bug or something else the message should definately be replaced with a confident information about it.

amorfo77:
- I have changed the error messages
- I am going to try to solve the problem with the SNAT rule, it is always deleted when starting the container even if it is correct, it seems that in the check there is something that does not work correctly.

amorfo77:
After some tests I have managed to reproduce the error you mention.

It seems that when running the whole mailcow stack for the first time, the ipv6nat container or docker itself tries to put the DOCKER-USER rule first at the same time that the netfilter container tries to put the MAILCOW rule first.

running the stack for the first time without the ipv6nat container this error does not occur.

EDIT: this error only happens in the ipv6 table

With ipv6nat container
docker ipv6_on-ipv6nat_on

without ipv6nat container
docker ipv6_on-ipv6nat_off


Tested in Ubuntu 22.04

amorfo77:
About the SNAT rule that is deleted, and then inserted every time the stack is started, it is because the next rule is inserted at the beginning, which causes the MAILCOW rule to be moved further down the list.
```oifname != "br-mailcow" ip saddr 172.22.1.0/24 counter packets 172 bytes 12107 masquerade```

DerLinkman:
What is the status here?

Are your issues mentioned before fixable? Does it work with other OSes like Centos/Rocky Linux?

If not it needs to be done before it is going to be merged in master!

amorfo77:
Hi @DerLinkman , i've tested on:
Ubuntu 18.04, 20.04, 22.04
Centos 7
Centos stream 9
Debian 11

**[Known bug/limitation]** Debian 10 snat works, but can't list the rules on the host, maybe caused by the version of iptables-nft or nftables used.

EDIT: the issues that were occurring,i think they have been fixed, I have also changed the log messages.

DerLinkman:
Hi,

what is the current status here?

The Debian 10 issue is not a problem for much long as we drop Debian 10 support once Debian 12 is released. Speaking of: Has someone tested this on Debian 12 already?

What about Centos? Does it work there flawlessly?

Does the automatic detection works as expected? How does it behave once you've manually installed nftables? How does it behave once you've uninstalled nftables, does it switch back to iptables or is it broken then? (Looking especially in the migration direction for example one host has nftables installed the other one does not).

Until these questions are not answered/tested we won't merge it. Sorry!

But thanks for your work for now, really appreciate it!

milkmaker:

PR #5085, opened 2023-02-21
Fix cursor style when hovering 'Aliases' tab

kritzl:
The cursor style is wrong when hovering over the `Aliases` tab in the `Email > Configuration` view, caused by the missing `href` attribute.


DerLinkman:
Works! Thanks :)

PR #5093, opened 2023-02-27
Fix SNAT never being added because of exception

brunoleon:
Some firewall rule object (iptc) do not have a parameter attribute, which results in an exception being triggered, and the mailcow SNAT rule to never be created.

Firewall rules that trigger such exception are:
- -A POSTROUTING -s 192.168.122.0/24 -d 224.0.0.0/24 -j RETURN

This commit just verify attribute presence, and skip the rule properly instead of triggering an exception.


milkmaker:
Thanks for contributing!

FreddleSpl0it (approved):

PR #5097, opened 2023-02-28
fix URLHAUS_ABUSE_CH check

rekup:
The current `URLHAUS_ABUSE_CH` rule uses the `full` filter to compare urls to the `text_online` list of urls. However, the `full` filter "matches the complete URL (not the hostname)" (https://www.rspamd.com/doc/modules/multimap.html#url-filters). As a result, emails which contain urls in the list mentioned above will never match the `URLHAUS_ABUSE_CH` rule:

```bash
[root@mx01 ~]# cat /tmp/mail 
Date: Tue, 28 Feb 2023 14:15:57 +0100
To: Test <test@example.com>
From: Test <test@example.org>
Subject: test spam

test http://<REDACTED>:55682/i

[root@mx01 ~]# grep -A5 URLHAUS_ABUSE_CH /etc/rspamd/local.d/multimap.conf
URLHAUS_ABUSE_CH {
  type = "url";
  selector = "full";
  map = "https://urlhaus.abuse.ch/downloads/text_online/";
  score = 5.0;
}

[root@mx01 ~]# rspamc -h localhost:11334 < /tmp/mail
Results for file: stdin (0.13 seconds)
[Metric: default]
Action: no action
Spam: false
Score: 5.26 / 15.00
Symbol: ARC_NA (0.00)
Symbol: DMARC_NA (0.00)[example.org]
Symbol: FROM_HAS_DN (0.00)
Symbol: HFILTER_HOSTNAME_UNKNOWN (2.50)
Symbol: HFILTER_URL_ONLY (0.26)[0.11764705882353]
Symbol: MIME_GOOD (-0.10)[text/plain]
Symbol: MIME_TRACE (0.00)[0:+]
Symbol: MISSING_MID (2.50)
Symbol: NEURAL_SPAM (0.00)[1.000]
Symbol: RCPT_COUNT_ONE (0.00)[1]
Symbol: RCVD_COUNT_ZERO (0.00)[0]
Symbol: R_DKIM_NA (0.00)
Symbol: TO_DN_ALL (0.00)
Message-ID: undef
Urls: ["<REDACTED>"]
Emails: []
```

This can be fixed by using a selector map:

```bash
[root@mx01 ~]# grep -A5 URLHAUS_ABUSE_CH /etc/rspamd/local.d/multimap.conf
URLHAUS_ABUSE_CH {
  type = "selector";
  selector = "urls";
  map = "https://urlhaus.abuse.ch/downloads/text_online/";
  score = 5.0;
}

[root@mx01 ~]# rspamc -h localhost:11334 < /tmp/mail
Results for file: stdin (0.169 seconds)
[Metric: default]
Action: greylist
Spam: false
Score: 10.26 / 15.00
Symbol: ARC_NA (0.00)
Symbol: DMARC_NA (0.00)[example.org]
Symbol: FROM_HAS_DN (0.00)
Symbol: HFILTER_HOSTNAME_UNKNOWN (2.50)
Symbol: HFILTER_URL_ONLY (0.26)[0.11764705882353]
Symbol: MIME_GOOD (-0.10)[text/plain]
Symbol: MIME_TRACE (0.00)[0:+]
Symbol: MISSING_MID (2.50)
Symbol: NEURAL_SPAM (0.00)[1.000]
Symbol: RCPT_COUNT_ONE (0.00)[1]
Symbol: RCVD_COUNT_ZERO (0.00)[0]
Symbol: R_DKIM_NA (0.00)
Symbol: TO_DN_ALL (0.00)
Symbol: URLHAUS_ABUSE_CH (5.00)[http://<REDACTED>:55682/i]
Message-ID: undef
Urls: ["<REDACTED>"]
Emails: []

### CODE REVIEW

**Comment by milkmaker** [comment]
Thanks for contributing!

**Comment by DerLinkman** [comment]
Thank you!

**Comment by realizelol** [comment]
For anyone who is sceptical like me, this will also respect the part after the hostname.
So lines in https://urlhaus.abuse.ch/downloads/text_online/ like `https://github.com/adwxawjwj/avsfdvfd/releases/download/vsd/Software-PC-2023.rar` (slightly modified - don't want to post bad content) will not be triggered if any other url like e.g. `https://github.com/mailcow/mailcow-dockerized` will be in mail content.

What's the concrete difference I don't know. Maybe the protocol (http://|https://|ftp://|...)is being ignored?

But I think your variant is the better deal. Thank you for your contributing.

Proof:
```
Date: Tue, 28 Feb 2023 14:15:57 +0100
To: Test <test@example.com>
From: Test <test@example.org>
Subject: test spam

test http://115.59.11.123:55682/Mozilla.m
test http://115.59.11.123:55682/i
test https://github.com/adwxawjwj/avsfdvfd/releases/download/vsd/Software-PC-2023.rar
test https://github.com/PC
test https://pastebin.com/raw/Upi48Fex
test https://pastebin.com/raw/UasdKKce
test https://drive.google.com/uc?export=download&id=1HA7by--2GYop0o8j0_pCFdaHOuc7cAuX
test https://drive.google.com/uc?export=download
```
```
rspamc -h localhost:11334 < /tmp/testmail
Results for file: stdin (2.28 seconds)
[Metric: default]
Action: reject
Spam: true
Score: 44.21 / 15.00
Symbol: ARC_NA (0.00)
Symbol: DATE_IN_PAST (1.00)[72]
Symbol: DMARC_NA (0.00)[example.org]
Symbol: FROM_HAS_DN (0.00)
Symbol: HFILTER_HOSTNAME_UNKNOWN (2.50)
Symbol: HFILTER_URL_ONLY (0.87)[0.39546599496222]
Symbol: MIME_GOOD (-0.10)[text/plain]
Symbol: MIME_TRACE (0.00)[0:+]
Symbol: MISSING_MID (2.50)
Symbol: RCPT_COUNT_ONE (0.00)[1]
Symbol: RCVD_COUNT_ZERO (0.00)[0]
Symbol: R_DKIM_NA (0.00)
Symbol: TO_DN_ALL (0.00)
Symbol: URLHAUS_ABUSE_CH (40.00)[http://115.59.11.123:55682/Mozilla.m, https://pastebin.com/raw/Upi48Fex, https://drive.google.com/uc?export=download&id=1HA7by--2GYop0o8j0_pCFdaHOuc7cAuX, https://github.com/adwxawjwj/avsfdvfd/releases/download/vsd/Software-PC-2023.rar]
Message-ID: undef
Urls: ["115.59.11.231","pastebin.com","drive.google.com","github.com"]
Emails: []
```

best regards
realizelol

rekup:
Hi realizelol

Thanks for taking the time to verify my change.
I'm not sure if I understand you correctly, isn't it supposed to respect the full url and not just the host part of it?
As far as I can tell, the list contains only urls. So if I check an emali with a malicious github url and a legitimate github url I will get a match for the malicious one and not for the legitimate url.

realizelol:
Hi rekup,

yes, you're right. It matches the malicious url and not the legitimate url.

But in case of your statement:
`However, the full filter "matches the complete URL (not the hostname)"`
I was thinking it now checks just the hostnames part of the url of the URLHAUS list.
That would be a no-go as this list will contain some legit hostnames like github.com or drive.google.com.

So I have implemented your change and checked some stuff and recognized that it will not only use the hostname part
instead it will still use the part after the domain.tld/... or nodeip:port/...

So don't worry, everything is fine. I just don't know the exact difference between these to variants:
`type = "url"; + filter = "full";` and `type = "selector"; + selector = "urls";`
It's moreover the lack of transparency in the rspamd docs @ https://www.rspamd.com/doc/modules/multimap.html
between these two variants.

PS: I didn't test the previous variant, wasn't there no match at all?

best regards
realizelol

rekup:
Hi realizelol

Ah okay, thanks for clarifying :+1: 
No, unfortunately there was no match at all. But I also checked the other rule which checks the md5sum of attachments against the https://bazaar.abuse.ch/export/txt/md5/recent/ list and this one worked flawlessly.

Best regards
Reto

PR #5109, opened 2023-03-04
[SOGo] Disable password change option

dragoangel:
It doesn't work with ProxyAuth and in general not honor password policy set via mailcow UI. SOGo also do not provide own settings to provide any password policy. Due to this two issues I think that it's better have it disabled by default. People who need it can turn it back easily. We can update https://docs.mailcow.email/manual-guides/SOGo/u_e-sogo/#disable-password-changing to `enable-password-changin` and explanations of reasons why it is disabled.


DerLinkman:
I don't see a reason why not disabling this option so thanks!

Can you also rewrite the English documentation part so I can translate it?

dragoangel:
@DerLinkman please check https://github.com/mailcow/mailcow-dockerized-docs/pull/541

Issue #5118, opened 2023-03-08
Loading more logs (+100/+1000) loads always same logs


dalferth:
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... I understand that not following the below instructions will result in immediate closure and/or deletion of my issue.
- [X] ... I have understood that this bug report is dedicated for bugs, and not for support-related inquiries.
- [X] ... I have understood that answers are voluntary and community-driven, and not commercial support.
- [X] ... I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

### Description

```plain text
Under "System > Logs > (e.g.) Postfix", theres a button "Actions" to load more logs. When I inspect the network traffic with my browser, I see, that always the same request (for the same logs) is sent.
```

### Logs:

```plain text
There are no logs regarding this error
```

### Steps to reproduce:

```plain text
1. Go to "System > Logs > Postfix"
2. Open your browser developer tools
3. Click the button "Actions"
4. Click on "+ 1000"
5. Click on "+ 1000"
6. Inspect the last API-Requests ("/api/v1/get/logs/postfix/51-1050")
```

### Which branch are you using?

master

### Operating System:

Debian GNU/Linux 11 (bullseye)

### Server/VM specifications:

8 GB RAM, 2 vCPU

### Is Apparmor, SELinux or similar active?

no

### Virtualization technology:

KVM

### Docker version:

20.10.22

### docker-compose version or docker compose version:

v2.14.1

### mailcow version:

2023-03

### Reverse proxy:

nothing in addition to mailcow

### Logs of git diff:

```plain text
-
```

### Logs of iptables -L -vn:

```plain text
# Warning: iptables-legacy tables present, use iptables-legacy to see them
Chain INPUT (policy ACCEPT 0 packets, 0 bytes)
 pkts bytes target     prot opt in     out     source               destination         
 650K   68M f2b-sshd   tcp  --  *      *       0.0.0.0/0            0.0.0.0/0            multiport dports 22

Chain FORWARD (policy DROP 1 packets, 52 bytes)
 pkts bytes target     prot opt in     out     source               destination         
8661K 2919M DOCKER-USER  all  --  *      *       0.0.0.0/0            0.0.0.0/0           
[...]

PR #5119, opened 2023-03-08
Fixes Issue #5118 (Bug with load more logs buttons)

dalferth:
Fixes issue #5118


DerLinkman (approved):
Works as expected! Thanks

Issue #5086, opened 2023-02-22
[netfilter] Incremental banning similar to fail2ban


th-joerger:
### Summary

Introduce mechanism to incrementally increase ban times, similar to [fail2ban's bantime.increment](https://github.com/fail2ban/fail2ban/blob/bc075ea682a3e3678ac5449cb1baeb0420655e8d/fail2ban/server/jail.py).

### Motivation

I see some IPs failing to authenticate over weeks at a time. They currently are banned for the standard time of 30 minutes, but come back 30 minutes later. I increased this to 24h, but cannot extend it further, because it will exclude some legitimately failed authentication attempts. I also don't want to manually add them to a blacklist. This should be automated and configured as a standard setting.

### Additional context

- If the IP was banned previously, the next bantime should be doubled each time. This can be achieved on two ways.
    * Get the last ban from the logs (➕ works without database, ➖  bans get deleted after each container update/restart)
    * Get the last ban from a database (➕ works between restarts, ➖ additional database implementation required)
- Optional: Respect a maximum bantime (let's say 365 days).
- Optional: Let the user set the settings.

martinjgrunwald:
I am not sure this is working correctly. There are several IPs that are spamming me non-stop. I attached my Fail2Ban settings and the mails I receive from my mailcow. From my understanding it should block 24 hours at first but more than that afterwards. This IP address (among several others) however is never banned for more than 48 hours as you can see from the mails I get. 
I am on version 2024-04
<img width="1365" alt="Bildschirmfoto 2024-05-14 um 12 00 37" src="https://github.com/mailcow/mailcow-dockerized/assets/69245880/34a5f0bb-c14d-43e0-bfac-d2c1af54d38e">
<img width="294" alt="Bildschirmfoto 2024-05-14 um 12 01 50" src="https://github.com/mailcow/mailcow-dockerized/assets/69245880/da40c71e-17d0-4f19-bb84-20c97d8cc3ee">

PR #5127, opened 2023-03-17
[Netfilter] Implemented exponentially incrementing bantime

th-joerger:
Implemented exponentially incrementing bantime, removed active_window code that did nothing, cleanly initialized dictionary

This PR implements #5086 in the most basic form.

Container can be checked out and tested by replacing:

```
    netfilter-mailcow:
      image: mailcow/netfilter:1.51
```

with

```
    netfilter-mailcow:
      image: thorjoerger/netfilter:1.51
```

in docker-compose.yml


mkuron:
We should probably cap the exponential growth at some point (maybe at 24 hours). IP addresses do occasionally get reassigned, so we don't want to punish the new owner of an IP address with a two-year ban time even if the old owner deserved it for getting banned 20 times.

th-joerger:
> We should probably cap the exponential growth at some point (maybe at 24 hours). IP addresses do occasionally get reassigned, so we don't want to punish the new owner of an IP address with a two-year ban time even if the old owner deserved it for getting banned 20 times.

I agree and also mentioned this as optional in my original issue. But this option (as well as incremental ban times itself) should be exposed as options in the UI. This requires work in other containers. I can do this, but only if @DerLinkman agrees that this has any chance to be merged back.

DerLinkman:
Yeah sounds good to me!

No reason not to implement it in the ui as well.

th-joerger:
I implemented the UI elements and api, tested it on my instance and seems to work fine. I added all the translations I felt confident enough.

grafik

DerLinkman:
Thanks for your work, really appreciate it! Will test it next week and stage it if everything goes well :)

th-joerger:
Thanks a lot!

th-joerger:
> Thanks for your work, really appreciate it! Will test it next week and stage it if everything goes well :)

Did you get the chance to test it? Does everything work? @DerLinkman

FreddleSpl0it:
i have tested it and it works on a freshly installed mailcow.
However if i have the latest mailcow installed and upgrade the netfilter container i get the following error

```
Clearing all bans
Initializing mailcow netfilter chain
Watching Redis channel F2B_CHANNEL
Exception in thread Thread-2 (autopurge):
Traceback (most recent call last):
  File "/usr/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.10/threading.py", line 953, in run
    self._target(*self._args, **self._kwargs)
  File "/server.py", line 430, in autopurge
    MAX_BAN_TIME = int(f2boptions['max_ban_time'])
KeyError: 'max_ban_time'
```

th-joerger:
I updated my PR and reworked the option handling. Can you test if this fixes the problem for you? I updated the docker container from my first post here so you can use this for testing @FreddleSpl0it

FreddleSpl0it:
Looks good! it's working now as expected.
The only thing is, that there are no default values in the mailcow UI but netfilter uses it's default values `max_ban_time = 10000` and `ban_time_increment = True`. 
I think that the netfilter container needs to push the default values to redis if nothing is found.

th-joerger:
Do you have any idea where this is done initially? Because the only place where the defaults are mentioned (which I am aware of) are in the netfilter/server.py, so that should be the place where they are initially set.

th-joerger:
Ok, I think I found and fixed the problem. This should work now. Can you verify again?

FreddleSpl0it:
Found the issue. it's happening in the refreshF2boptions method

Edit: looks good i will test it

th-joerger:
I thought this whole option code felt a little off, so I looked a bit deeper into it and it seems, those redis suboptions where the result of a migration to netfilter from fail2ban itself in 38a819771bfdb49500f59048419c7ff170aa9737. They will never be set in a deployment, except you upgrade mailcow in a single step from a 2018 deployment to a 2023 deployment.

FreddleSpl0it (approved):

PR #5133, opened 2023-03-20
BS5 related fixes

feldsam:
Hi, I fixed some thing I see after new BS5 update. 

- mailboxes table - smtp/imap/pop login time badge - when no login, them badge is higher due to big "X" icon
- on aliases table, when row contains info icon, tooltip doesnt show because table is loaded by ajax now
- mailbox edit / user settings - btn groups doesn't differentiate well active vs non-active items, so I decided to go with light/dark. It is also better, if user select other BS theme.

### DIFF
--- a/data/web/css/site/mailbox.css
+++ b/data/web/css/site/mailbox.css
@@ -66,4 +66,6 @@ table tbody tr td input[type="checkbox"] {
   padding: .2em .4em .3em !important;
   background-color: #ececec!important;
 }
-
+.badge.bg-info .bi {
+  font-size: inherit;
+}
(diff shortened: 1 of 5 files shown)


FreddleSpl0it:
Using light and dark buttons to differentiate may not work in dark mode. We would need extra dark theme css.
Alternatively, we could consider using outlined and solid buttons to differentiate instead. 

image
image

instead of
image
image

feldsam:
Ok, I check dark theme and fix it

feldsam:
Hi @FreddleSpl0it , I just upated this PR - fixed buttons in dark mode

feldsam:
Screenshot 2023-03-31 at 13 23 29

feldsam:
Hi @FreddleSpl0it, would you review and merge? I have more things to push, but are blocked by this one

PR #5141, opened 2023-03-21
[MariaDB] Upgrade to 10.6 (LTS)

DerLinkman:
This PR is upgrading MariaDB (finally) to 10.6 LTS.

It also includes the mariadb_upgrade option which gets detected during the mysql upgrade phase.

This change is made since it seems that mysql_upgrade does not upgrade the MariaDB from 10.5 (or lower) to 10.6 properly if there is no --force option appended or if it get upgraded twice. But the upgrade process did not see this as a error.

The current tests applied that the upgrade process runs smoothly.

This issue mention the desired "bug": https://jira.mariadb.org/browse/MDEV-29648?attachmentOrder=desc

Updated Images:
+ mysql-mailcow (mariadb:10.5 -> mariadb:10.6)
+ phpfpm-mailcow (mailcow/phpfpm:1.82 -> mailcow/phpfpm:1.83)
+ dockerapi-mailcow (mailcow/dockerapi:2.01 -> mailcow/dockerapi:2.02)

### DIFF
--- a/data/Dockerfiles/dockerapi/dockerapi.py
+++ b/data/Dockerfiles/dockerapi/dockerapi.py
@@ -321,7 +321,7 @@ class DockerUtils:
   # api call: container_post - post_action: exec - cmd: system - task: mysql_upgrade
   def container_post__exec__system__mysql_upgrade(self, container_id, request_json):
     for container in self.docker_client.containers.list(filters={"id": container_id}):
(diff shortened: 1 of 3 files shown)


MAGICCC (approved):

DerLinkman:
Haven't seen this issue anymore.

Tested in on a hand picked amount of mailcows we host.

But i'm pretty unsure how to continue with that. Sure it is not needed as 10.5 is still supported for a few years but i would really like to switch to a official LTS Version from MariaDB.

PR #5152, opened 2023-03-28
[Mariadb] Update to 10.11 (LTS)

DerLinkman:
This PR includes the mariadb upgrade from the current mailcow used version (10.5) to the latest LTS Version from mariadb to 10.11.

It also includes a change in case of thread_stack size value which had to be increased from 128k to 256k due to the mysql_upgrade process.

In some very rare cases the mysql upgrade will fail from 10.5 to 10.11 and needs to be rerun manually from cli (at least it has confirmed for the versions < 10.11 >=10.6 (https://jira.mariadb.org/browse/MDEV-29648?attachmentOrder=desc)

### DIFF
--- a/data/conf/mysql/my.cnf
+++ b/data/conf/mysql/my.cnf
@@ -1,7 +1,7 @@
 [mysqld]
 character-set-client-handshake = FALSE
 character-set-server           = utf8mb4
-collation-server               = utf8mb4_unicode_ci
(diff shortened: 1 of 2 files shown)


milkmaker:

DerLinkman (commented):
In tests everything looked fine, after change of default collation.

PR #5160, opened 2023-04-02
Easier log_format adjustment of nginx

### LINKED ISSUE #5159: Make it easier to adjust the access log format in nginx
### Summary

In nginx.conf you can configure an log_format main which also exists by default. But that log_format doesn't get used because when access_log is specified with no log_format (access_log <logLocation> <logFormat>) the default log_format gets used. If you want to change that format you have to re specify that you want to use the main format at every acces_log variable. 

### Motivation

It makes it easier to adjust the logging format of nginx. Then you just have to override the access_format property in the nginx.conf with your desired value. 

### Additional context

_No response_

Toaster2-0:
Added the log_format in the configuration files of nginx. 

resolves #5159


milkmaker:
Thanks for contributing!

milkmaker:

Issue #4980, opened 2023-01-12
Performance: account creation gets considerably slower, the more users are in mailcow


missytake:
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... I understand that not following the below instructions will result in immediate closure and/or deletion of my issue.
- [X] ... I have understood that this bug report is dedicated for bugs, and not for support-related inquiries.
- [X] ... I have understood that answers are voluntary and community-driven, and not commercial support.
- [X] ... I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

### Description

We stress-tested mailcow with a script - it creates 4000 users, and every 100 users it reports on the time it took to create the users. It came out that every additional user in mailcow makes the request take longer significantly. This puts a limit to mailcow's performance with many users.

### Logs:

```plain text
Cleaning up first...
Request took 4.23 seconds.
Cleaned up 0 users.
Creating 4000 users - continue?
Created 0 accounts; the last 100 accounts took 0.00 seconds
Created 100 accounts; the last 100 accounts took 14.25 seconds
Created 200 accounts; the last 100 accounts took 16.05 seconds
Created 300 accounts; the last 100 accounts took 18.02 seconds
Created 400 accounts; the last 100 accounts took 20.19 seconds
Created 500 accounts; the last 100 accounts took 23.26 seconds
Created 600 accounts; the last 100 accounts took 26.67 seconds
Created 700 accounts; the last 100 accounts took 29.43 seconds
Created 800 accounts; the last 100 accounts took 34.02 seconds
Created 900 accounts; the last 100 accounts took 38.61 seconds
Created 1000 accounts; the last 100 accounts took 40.74 seconds
Created 1100 accounts; the last 100 accounts took 45.32 seconds
Created 1200 accounts; the last 100 accounts took 51.28 seconds
Created 1300 accounts; the last 100 accounts took 56.69 seconds
Created 1400 accounts; the last 100 accounts took 61.37 seconds
Created 1500 accounts; the last 100 accounts took 67.01 seconds
Created 1600 accounts; the last 100 accounts took 92.26 seconds
Created 1700 accounts; the last 100 accounts took 77.72 seconds
Created 1800 accounts; the last 100 accounts took 83.52 seconds
[...]

PR #5183, opened 2023-04-13
[Web] optimizing mailbox add/edit/delete performance

FreddleSpl0it:
As the number of mailboxes increases, the execution time of the `update_sogo_static_view` function will also increase. When a new mailbox is added, the information is saved in the mailbox table and is reflected in the `sogo_view`. The `update_sogo_static_view` function then updates the `_sogo_static_view` table accordingly.

The bottleneck is caused by this query:
```
REPLACE INTO _sogo_static_view (`c_uid`, `domain`, `c_name`, `c_password`, `c_cn`, `mail`, `aliases`, `ad_aliases`, `ext_acl`, `kind`, `multiple_bookings`)
      SELECT `c_uid`, `domain`, `c_name`, `c_password`, `c_cn`, `mail`, `aliases`, `ad_aliases`, `ext_acl`, `kind`, `multiple_bookings` from sogo_view
```

After some experimentation, I found a solution that doesn't suffer from performance issues (in my tests). When a mailbox is added, edited, or deleted, the `update_sogo_static_view` function will only try to update that specific mailbox in the `_sogo_static_view`. Furthermore, the mailbox table is used directly and the `sogo_view` view is skipped, as it is not possible otherwise.

@mkuron what do you think?
The issue was first mentioned here: https://github.com/mailcow/mailcow-dockerized/issues/4980. At that point, I only added a skip if SOGo was not being used.

Benchmark:
before
```
Created 0 accounts; the last 100 accounts took 0.00 seconds
Created 100 accounts; the last 100 accounts took 9.61 seconds
Created 200 accounts; the last 100 accounts took 9.86 seconds
Created 300 accounts; the last 100 accounts took 10.47 seconds
Created 400 accounts; the last 100 accounts took 11.82 seconds
```

after
```
Created 0 accounts; the last 100 accounts took 0.00 seconds
Created 100 accounts; the last 100 accounts took 9.37 seconds
Created 200 accounts; the last 100 accounts took 9.72 seconds
Created 300 accounts; the last 100 accounts took 9.57 seconds
Created 400 accounts; the last 100 accounts took 9.60 seconds
Created 500 accounts; the last 100 accounts took 9.78 seconds
Created 600 accounts; the last 100 accounts took 9.89 seconds
Created 700 accounts; the last 100 accounts took 10.08 seconds
Created 800 accounts; the last 100 accounts took 10.62 seconds
Created 900 accounts; the last 100 accounts took 9.91 seconds
Created 1000 accounts; the last 100 accounts took 10.36 seconds
Created 1100 accounts; the last 100 accounts took 10.82 seconds
```


mkuron:
Yes, that looks like a good idea.

Issue #5054, opened 2023-02-08
Unable to sort Mailboxes in the Mailcow UI by disabled state


pfetzing:
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... I understand that not following the below instructions will result in immediate closure and/or deletion of my issue.
- [X] ... I have understood that this bug report is dedicated for bugs, and not for support-related inquiries.
- [X] ... I have understood that answers are voluntary and community-driven, and not commercial support.
- [X] ... I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

### Description

```plain text
Hi,

after migrating from an old mail server setup to Mailcow, I mass disabled unused mailboxes. Now if I try to only see the enabled ones, first of all there is no possibility to search for something like disabled:false as it seems to only match the mailbox name, and second the ordering by disabled state when clicking on the table header for the matching mailboxes doesn’t work, it simply leaves the table unchanged.

I already attempted to report this via https://community.mailcow.email/d/2194-unable-to-sort-mailboxes-in-the-mailcow-ui-by-disabled-state, but I think an issue for this is better for tracking.

Thanks,
Stefan
```

### Logs:

```plain text
There is nothing in the logs related to this UI issue.
```

### Steps to reproduce:

```plain text
1. Go to the list of mailboxes in the Mailcow UI.
2. Disable at least one mailbox, or already have a disabled Mailbox
3. Click on the column title for the active state of the Mailboxes

Expected result:
Mailboxes are sorted by disabled/enabled state, showing either the disabled or the enabled mailboxes first.

Actual result:
The ordering of all mailboxes does not change.
```

### Which branch are you using?

master

### Operating System:

Debian 11

### Server/VM specifications:

4GiB RAM, 2 cores

### Is Apparmor, SELinux or similar active?

yes, apparmor

### Virtualization technology:

KVM

### Docker version:

23.0.0

### docker-compose version or docker compose version:

v2.15.1

### mailcow version:

2023-02

### Reverse proxy:

Nginx

### Logs of git diff:

```plain text
n/a, only minor changes like mynetworks for postfix.
```

### Logs of iptables -L -vn:

```plain text
Chain INPUT (policy ACCEPT 0 packets, 0 bytes)
[...]

PR #5186, opened 2023-04-13
[Web] Datatables sorting

goodygh:
I looked into the [Ticket](https://github.com/mailcow/mailcow-dockerized/issues/5054) and found that sorting on multiple columns where broken, because there where no values to sort on. So i fixed every column / table i found. Tested it with a local setup.

Is there something more what needs to be done?
Or in general i would appreciate any input.


milkmaker:
Thanks for contributing!

data/web/js/site/mailbox.js
```diff
@@ -1329,7 +1328,7 @@ jQuery(function($){
           defaultContent: '',
           responsivePriority: 4,
           render: function (data, type) {
-            return 1==data?'<i class="bi bi-check-lg"></i>':(0==data?'<i class="bi bi-x-lg"></i>':2==data&&'&#8212;');
+            return 1==data?'<i class="bi bi-check-lg"><span class="sorting-value">1</span></i>':(0==data?'<i class="bi bi-x-lg"><span class="sorting-value">0</span></i>':2==data&&'&#8212;');
```
FreddleSpl0it, line 1331: Thanks for the PR. Works so far, but could you check the mailbox table again?
There are 3 states (0, 1, 2). Please add the sorting value also for the last state.

goodygh: Your welcome and thanks for your awesome work. It's already sort able as it has the value [`&#8212;`](https://www.codetable.net/decimal/8212) which is shown as a dash. But i decided to not hide it as there was no Icon representing this state.

FreddleSpl0it (approved):

Issue #5136, opened 2023-03-20
Large number of blacklisted networks/hosts appears to break web interface session?


ThinIce:
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... I understand that not following the below instructions will result in immediate closure and/or deletion of my issue.
- [X] ... I have understood that this bug report is dedicated for bugs, and not for support-related inquiries.
- [X] ... I have understood that answers are voluntary and community-driven, and not commercial support.
- [X] ... I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

### Description

```plain text
My blacklisted networks/hosts list in the mailcow gui contains a number of existing bans, around 3500.

Upon adding a couple hundred more and clicking "save changes" the page posted back to a blank white screen, view source shows no content. The same is true if I try and hit the root URL rather than the /debug URL.

On opening up a new browser session, completing login, the page displays okay, and even contains the new IPs entered, however same problem can be reproduced by entering a new IP and repeating. I've not yet restarted the mailcow stack.

I imagine this could be a bug, unless I've somehow pasted in an odd character. Could you let me know what logs are likely to be relevant and I will post them, I don't immediately see anything wrong in the below.

Many thanks
```

### Logs:

```plain text
mailcowdockerized-php-fpm-mailcow-1    | 172.22.1.10 -  20/Mar/2023:21:02:08 +0000 "POST /json_api.php" 200
mailcowdockerized-nginx-mailcow-1      | snip - - [20/Mar/2023:21:02:08 +0000] "POST /api/v1/edit/fail2ban HTTP/2.0" 200 19128 "https://snip/" "Mozilla/5.0 (X11; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/110.0"
mailcowdockerized-php-fpm-mailcow-1    | 172.22.1.10 -  20/Mar/2023:21:02:09 +0000 "GET /settings.php" 200
mailcowdockerized-nginx-mailcow-1      | 172.22.1.9 - - [20/Mar/2023:21:02:09 +0000] "GET /settings.php HTTP/1.1" 200 1184 "-" "-"
[...]

PR #5215, opened 2023-04-29
[web] logger pdo exception handling workaround

goodygh:
Its more of a workaround then a clean solution, as i was not able to log the error in the frontend, or to to resolve the underlying problem for good.

But as it seems we are trying to log every request, but if the request is longer then about 65000 characters it wont fit the column and therefor results in a SQL Error.

```
SQLSTATE[22001]: String data, right truncated: 1406 Data too long for column 'call' at row 1
```

It should be fix on API level which would be quite bit of work. As the logging function is not working anyway, i suggest to implement this workaround, so at least the frontend is still working and a error log message in the stdout.

Ref: #5136

### DIFF
--- a/data/web/inc/functions.inc.php
+++ b/data/web/inc/functions.inc.php
@@ -526,8 +526,9 @@ function logger($_data = false) {
           ':remote' => get_remote_ip()
         ));
       }
-      catch (Exception $e) {
-        // Do nothing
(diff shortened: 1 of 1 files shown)


milkmaker:

FreddleSpl0it (approved):

Issue #5067, opened 2023-02-16
Using BCCMaps truncates mail in some circumstances


infinitytrans:
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... I understand that not following the below instructions will result in immediate closure and/or deletion of my issue.
- [X] ... I have understood that this bug report is dedicated for bugs, and not for support-related inquiries.
- [X] ... I have understood that answers are voluntary and community-driven, and not commercial support.
- [X] ... I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

### Description

```plain text
Receiving an E-Mail to an adress with activated bcc map truncates the mail, if the mail contains a dot "." followed by newline.

Guess whatever is doing the bcc stuff, parses dot and newline as EOF as expected in SMTP protocol.
```

### Logs:

```plain text
no error in the logs
```

### Steps to reproduce:

```plain text
1. Add mailbox1@ and mailbox2@
2. Add BCC-Maps, "cloning" mails from mailbox1@ to mailbox2@
3. Send a random mail to mailbox1@, you get the mail in mailbox1@ and mailbox2@, everything is fine
4. now send a mail, ending with "." followed by newline. Add an attachment. Sent it to mailbox1@
5. Result: Mail in mailbox1@ has the attachment, Mail in mailbox2@ is missing the attachment - and the "."
6. Confirm: Sent the same mail "with dot and newline and attachment" to mailbox1@ and mailbox2@
7. Result: You wil get 2 mails in mailbox2@, one with attachment and one without
```

### Which branch are you using?

master

### Operating System:

Linux mail 4.19.0-22-amd64 #1 SMP Debian 4.19.260-1 (2022-09-29) x86_64

### Server/VM specifications:

4GB, 2 Cores, KVM

### Is Apparmor, SELinux or similar active?

default config

### Virtualization technology:

KVM

### Docker version:

20.10.21

### docker-compose version or docker compose version:

docker-compose version 1.29.2, build 5becea4c

### mailcow version:

2022-10a

### Reverse proxy:

default config

### Logs of git diff:

```plain text
none
```

### Logs of iptables -L -vn:

```plain text
Chain INPUT (policy ACCEPT 0 packets, 0 bytes)
[...]

PR #5221, opened 2023-05-03
[Rspamd] add dot-stuffing to bcc forwarding

FreddleSpl0it:
If the `BCC` rule finds at least one `bcc_dest`, the rule will retrieve the e-mail's content using `task:get_content()` and forward it via SMTP to the destination. The rule will use the `sendmail` method, which does not perform dot-stuffing. https://datatracker.ietf.org/doc/html/rfc5321#section-4.5.2

If the e-mail contains a new line with a period, it won't be fully forwarded. 
This PR addresses the issue by adding dot-stuffing before forwarding the mail to the `bcc_dest`.
Resolves: https://github.com/mailcow/mailcow-dockerized/issues/5067

### DIFF
--- a/data/conf/rspamd/lua/rspamd.local.lua
+++ b/data/conf/rspamd/lua/rspamd.local.lua
@@ -340,6 +340,10 @@ rspamd_config:register_symbol({
       if not bcc_dest then
         return -- stop
       end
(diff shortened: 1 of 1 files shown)


milkmaker:

Issue #1485, opened 2018-06-17
Domain-wide disclaimer.


ayushin:
How difficult would it be to implement domain-wide disclaimers (signatures)?

What would be the right way to do it?

I suppose we need a "disclaimer" field per domain in the mysql database and then auto-generate config file for some additional milter (altermime?)

andryyy:
Hi, Rspamd will provide this feature probably in the near future. At least we can add/modify mime content. 
Altermime is not a milter and will break DKIM by changing the content after signing. 
We could also write our own milter for that.
Besides that we COULD implement a second pre-smtpd with Altermime. But I don’t like that idea. It would require multiple changes like another Rspamd listener for just rate limits and stuff like that.

ayushin:
@andryyy Thanks for the feedback. I am not intimately familiar with the milters, but I suppose its not a rocket science, so if you give me a pointer in the right direction I would be happy to code it in python.

So I just write a python milter, that would connect to the mailcow mysql database, look up the domain and the address (I guess we would need a flag to either add disclaimer to all the addresses or only those with a tick) and then add that disclaimer to the message.

Where in the pipeline should we call that milter? Is that smtpd content-milter?

deanroker123:
I have a requirement to be able to do this as well.
Any ideas if or when it would be possible?

iammattmartin:
@patschi I'd be interested in putting in for the future. Being able to add signatures on a per domain, or even per email address basis to ensure correct/uniform signatures would be great!

codiflow:
Hey together,

I've seen that **rspamd** has already implemented a function to do this **way back in 2019** with **version 1.9** of rspamd:
https://rspamd.com/announce/2019/03/12/rspamd-1.9.0.html

Mailcow actually ships with version 2.5 of rspamd so this tool is widely available. You can check this using the following command:
`docker-compose exec rspamd-mailcow rspamadm mime modify -h`

Since I don't know my way around well enough in hacking the config of rspamd or other mailcow components maybe someone of you does?

How can I process my outgoing mails as **files** to this rspamadm tool to get several disclaimers applied depending on their source domain?

Regards
Christian

andryyy:
I think there is no way to do it via API.

Schmandre:
Hello,

Iam also interested to use a domain wide disclaimer as signature. Any new informations/tutorials etc. here? ;)

Regards
Schmandre

LeonDelacroixe:
[...]

PR #5227, opened 2023-05-08
[Rspamd] add domain wide footer

FreddleSpl0it:
This PR adds a new rule to Rspamd. When an authenticated user sends an email, the rule will check if a domain-wide footer has been set for that specific domain. If such a footer exists, the rule will modify the MIME part of the email and add the footer to the very end of the email.

Resolves: https://github.com/mailcow/mailcow-dockerized/issues/1485


MAGICCC:
I just tested it and it works with the HTML footer, but I couldn't use the PLAIN one somehow. When I enter some text it won't add it to the mail

FreddleSpl0it:
@MAGICCC The email probably has both a text/plain and a text/html section. The mail client will then render the most suitable content type. Can you check the source code of the email? The footer should be there.

Maybe we should add some info. Admins need to provide both text/plain and a text/html footer to ensure that it will be displayed.

MAGICCC:
Oh @FreddleSpl0it, yes it's in the source of the mail. 
Indeed maybe a different wording should be used. Admins needs to know that they should fill both, one with HTML tags and the other without

MAGICCC:
@FreddleSpl0it is there some update on this btw?

alwey:
I tested this recently. This patch does indeed append text and HTML-footers to the mail body, respectively. However, it seems to break the DKIM-signatures. Sadly, therefore I had to roll back the patch.

FreddleSpl0it:
> I tested this recently. This patch does indeed append text and HTML-footers to the mail body, respectively. However, it seems to break the DKIM-signatures. Sadly, therefore I had to roll back the patch.

Can someone confirm this? In my tests DKIM was still working.

tilwegener:
Can you add variables, which will then be added to the Domain Wide Footer? For example %name% for the full name or %email% for the email address?

tilwegener:
@FreddleSpl0it Thanks for the great implementation of the domain wide footer. An improvement would be that you can exclude certain mailboxes from this, just no-reply emails or from one sends Laravel emails, it looks stupid if then also the footer is attached.

DerLinkman:
> @FreddleSpl0it Thanks for the great implementation of the domain wide footer. An improvement would be that you can exclude certain mailboxes from this, just no-reply emails or from one sends Laravel emails, it looks stupid if then also the footer is attached.

Could you please open a enhancement issue for that?

bnfone:
> > I tested this recently. This patch does indeed append text and HTML-footers to the mail body, respectively. However, it seems to break the DKIM-signatures. Sadly, therefore I had to roll back the patch.
> 
> Can someone confirm this? In my tests DKIM was still working.


Hi there,

I recently took the time to test the domain-wide footer feature in Mailcow version 2024-02 and encountered an issue regarding DKIM signatures. Despite various configurations and testing with different email clients such as Apple Mail on Mac, Thunderbird, and Spark, all emails sent to ping@tools.mxtoolbox.com for DKIM verification failed to maintain a valid DKIM signature. What's particularly intriguing is that this issue does not occur when emails are dispatched via SOGo, where the DKIM signature remains unaffected.


Has anyone else observed similar behavior, or could anyone provide further testing or insight into this issue? Any feedback or suggestions would be greatly appreciated .

Thanks in advance for your help.

Issue #5246, opened 2023-05-24
Notify error in sieve


baimard:
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... I understand that not following the below instructions will result in immediate closure and/or deletion of my issue.
- [X] ... I have understood that this bug report is dedicated for bugs, and not for support-related inquiries.
- [X] ... I have understood that answers are voluntary and community-driven, and not commercial support.
- [X] ... I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

### Description

```plain text
On mailbox filter configuration. I try to add à new filter with a notify command https://www.rfc-editor.org/rfc/rfc5435.html#page-3

But mailcow sail that the command not exist.
```

### Logs:

```plain text
mailcowdockerized-php-fpm-mailcow-1    | 172.22.1.13 -  24/May/2023:23:47:01 +0200 "GET /inc/ajax/sieve_validation.php" 200
mailcowdockerized-nginx-mailcow-1      | 192.168.4.10 - - [24/May/2023:23:47:01 +0200] "GET /inc/ajax/sieve_validation.php?script=if%20anyof%20(header%20%3Acontains%20%22to%22%20%22test%40test.fr%22%2C%20header%20%3Acontains%20%22cc%22%20%22test%40test.fr%22%2C%20header%20%3Acontains%20%22received%22%20%22for%20%3Ctest%40test.fr%3E%22)%0A%7B%0A%20%20%20%20%20%20%20%20fileinto%20%22Public.famille%22%3B%0A%7D HTTP/2.0" 200 76 "https://mail.cybercorp.fr/" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/112.0.0.0 Safari/537.36"
mailcowdockerized-watchdog-mailcow-1   | Wed May 24 23:47:03 CEST 2023 Olefy health level: 100% (5/5), health trend: 0
[...]

PR #5261, opened 2023-05-28
Update notify.xml

baimard:
#5246 correction for this bug report


milkmaker:
Thanks for contributing!

tomudding:
For the Mailcow maintainers; as I pointed out in GH-5246 I think that a better longterm solution would be the removal of the `notify` extension/module and replace it with `enotify` (the command will still be `notify`).

This not only ensures that you cannot use a heavily deprecated function, it also makes it easier for people who like to use the sieve as they do not have to scour the internet in search for an RFC _draft_ because the syntax they found in the Dovecot documentation does not match with the errors they receive from the sieve validation in Mailcow.

milkmaker:

milkmaker:

Issue #5252, opened 2023-05-27
Fail2Ban not working on Rspamd UI


bundyland:
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... I understand that not following the below instructions will result in immediate closure and/or deletion of my issue.
- [X] ... I have understood that this bug report is dedicated for bugs, and not for support-related inquiries.
- [X] ... I have understood that answers are voluntary and community-driven, and not commercial support.
- [X] ... I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

### Description

```plain text
Fail2ban doesn't seem to work on the Rspamd UI. Unsuccessful login attempts are not logged in the netfilter protocol and the attacking IP address is not blocked. I tried this on two different instances.
A reset of the netfilter entries in the UI hasn't fixed the problem.

I don't know if it's related, but in the logs I see:

 redis.exceptions.ConnectionError: Error 111 connecting to 172.22.1.249:6379. Connection refused.

Can the problem be related to this?
```

### Logs:

```plain text
mailcowdockerized-netfilter-mailcow-1  |     response = self._execute(conn, try_read)
mailcowdockerized-netfilter-mailcow-1  |   File "/usr/lib/python3.10/site-packages/redis/client.py", line 1507, in _execute
mailcowdockerized-netfilter-mailcow-1  |     return conn.retry.call_with_retry(
mailcowdockerized-netfilter-mailcow-1  |   File "/usr/lib/python3.10/site-packages/redis/retry.py", line 49, in call_with_retry
mailcowdockerized-netfilter-mailcow-1  |     fail(error)
mailcowdockerized-netfilter-mailcow-1  |   File "/usr/lib/python3.10/site-packages/redis/client.py", line 1509, in <lambda>
mailcowdockerized-netfilter-mailcow-1  |     lambda error: self._disconnect_raise_connect(conn, error),
mailcowdockerized-netfilter-mailcow-1  |   File "/usr/lib/python3.10/site-packages/redis/client.py", line 1496, in _disconnect_raise_connect
mailcowdockerized-netfilter-mailcow-1  |     raise error
[...]

PR #5262, opened 2023-05-28
Rspamd returns 401 on unsuccesful logins

MAGICCC:
This will fix https://github.com/mailcow/mailcow-dockerized/issues/5252

Seems Rspamd changed its code with https://github.com/rspamd/rspamd/commit/5e4de6c3fee4ed261225684cf2f2f4830ac3ccaf to close their https://github.com/rspamd/rspamd/issues/4218 which won't trigger the [netfilter regex](https://github.com/mailcow/mailcow-dockerized/blob/74bcec45f13b267be4f75919b9bd08c44c02ea6a/data/Dockerfiles/netfilter/server.py#L108) anymore

### DIFF
--- a/data/conf/nginx/includes/site-defaults.conf
+++ b/data/conf/nginx/includes/site-defaults.conf
@@ -114,7 +114,7 @@
       proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
       proxy_set_header X-Real-IP $remote_addr;
       proxy_redirect off;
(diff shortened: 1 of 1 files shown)


DerLinkman:
Is this still needed?

MAGICCC:
Yes the regex won't work because https://github.com/mailcow/mailcow-dockerized/blob/master/data/web/_rspamderror.php#L14-L17 dont get triggered at the moment

PR #5275, opened 2023-06-06
Fix typo for docker-compose log command example

ppicazo:
command is currently `docker compose` instead of `docker-compose`


milkmaker:
Thanks for contributing!

MAGICCC:
No it's fine. We are using the docker compose plugin as default https://docs.docker.com/compose/install/linux/

PR #5283, opened 2023-06-13
Update nextcloud heper script to disable SMTP TLS host verification

superpuffin:
Nextcloud started using another mailer library, which automatically uses starttls when available. Since nextcloud connects inside the docker network, the TLS certificate won't verify and no email could be sent. This PR disables TLS host verification in nextcloud for SMTP connections.


MAGICCC:
Is that because of v27 @superpuffin?

superpuffin:
No the library was changed in v26, see: https://docs.nextcloud.com/server/27/admin_manual/configuration_server/email_configuration.html#configuring-an-smtp-server

DerLinkman (approved):

PR #5295, opened 2023-06-23
Spamhaus DNSBL AS Detection

### LINKED ISSUE #5293: Spamhaus implement DQS
### Summary

Hello,

Since Spamhaus blocks OVHCloud IPs querying their DNSBL lists [now](https://twitter.com/spamhaus/status/1671141604705333248), we have to implent DQS.
This also includes maybe a check in the install script to check if the server is using an OVH IP and enables some options such an `echo hey you need to signup for a spamhaus key` and/or activating DQS Rspamd plugin https://github.com/spamhaus/rspamd-dqs

More info here with an install guide: https://www.spamhaus.com/resource-center/if-you-query-spamhaus-projects-dnsbls-via-ovhclouds-dns-move-to-the-free-data-query-service
Also postfix docs for postscreen setup: https://www.postfix.org/postconf.5.html#postscreen_dnsbl_reply_map

### Motivation

Allow users using OVHCloud IPs to query Spamhaus DNBLs

### Additional context

_No response_

DerLinkman:
This PR will solve issue: https://github.com/mailcow/mailcow-dockerized/issues/5293

It implements a check of the public IPv4 AS and reports if it is from AWS, Cloudflare or OVH as Spamhaus is blocking them to access the zen.spamhaus.org lists.

Therefore the DQS System of Spamhaus has been implemented within Postfix which uses a new variable called SPAMHAUS_DQS_KEY.

If the key is set it will use the DQS Lists for Postscreen if not it it will check if the IP is listed as a bad asn using curl to get a list from fuzzy.mailcow.email which is generated against bgpview.io API and deactivates (if necessary) the spamhaus.org list for postfix.


milkmaker:
Thanks for contributing!

MAGICCC:
I dont't have an OVH VPS, would need to get one for testing, but it looks okish

FingerlessGlov3s:
PBL is missing when using DQS. Is there a reason for this?

DerLinkman:
> PBL is missing when using DQS. Is there a reason for this?

Don't understand exactly what you mean by that? Could you please be a bit more prercise?

FingerlessGlov3s:
> > PBL is missing when using DQS. Is there a reason for this?
> 
> Don't understand exactly what you mean by that? Could you please be a bit more prercise?

You've added the following lists
 - https://www.spamhaus.org/zen/
 - https://www.spamhaus.org/dbl/
 - https://www.spamhaus.org/zrd/

I just had another read of the PBL https://www.spamhaus.org/pbl/ and on review, I'm not 100% sure if we should add this.

DerLinkman:
> I just had another read of the PBL https://www.spamhaus.org/pbl/ and on review, I'm not 100% sure if we should add this.

Just readded the PBL if DQS is enabled.

Can you test it? New Postfix image version has been uploaded to docker hub!

FingerlessGlov3s:
I haven't moved to the new way of doing the DQS yet, I manually changed my config file 😅.

I guess I'd have to move from stable to nightly too right to test? Which I don't really want to do.

my `main.cf` has the following enteries and it's working fine.
```
  redacted.zen.dq.spamhaus.net=127.0.0.[10;11]*8
  redacted.zen.dq.spamhaus.net=127.0.0.[4..7]*6
  redacted.zen.dq.spamhaus.net=127.0.0.3*4
  redacted.zen.dq.spamhaus.net=127.0.0.2*3
```

DerLinkman:
Yeah looks pretty much the same as i readded now :)

But good call on moving it to nightly!

milkmaker:
Thanks for contributing!

FingerlessGlov3s:
When this does hit stable, I'll have to have a look what changes I need to make to my `main.cf` so I can go back to supported config 😃

PR #5318, opened 2023-07-13
Limit quota notifications to once per-user, per-day.

SnailShea:
This commit uses datetime ordinal dates to create timestamps for comparison. If a quota notification email is sent, another will not be sent within the same day. I created this change due to some users receiving as many as 20 emails within one minute when their inbox begins to fill. 

It may be desirable to have more granular control over quota notification frequency. Ordinal dates were chosen because they make the time math much simpler, but are not useful checking any timestamp difference of less than 1 day. If desired, I can add additional commits to make the limitation configurable via the web UI. 

Testing was done by running a bash shell inside the dovecot container and manually invoking the script.

### DIFF
--- a/data/Dockerfiles/dovecot/quota_notify.py
+++ b/data/Dockerfiles/dovecot/quota_notify.py
@@ -2,6 +2,7 @@
 
 import smtplib
 import os
+from datetime import datetime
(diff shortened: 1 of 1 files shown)


milkmaker:
Thanks for contributing!

Issue #5353, opened 2023-07-31
Spamhaus DQS - Rejected email reveal DQS key


KiwiActinidia:
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... I understand that not following the below instructions will result in immediate closure and/or deletion of my issue.
- [X] ... I have understood that this bug report is dedicated for bugs, and not for support-related inquiries.
- [X] ... I have understood that answers are voluntary and community-driven, and not commercial support.
- [X] ... I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

### Description

```plain text
I installed the new update. All setup fixed if main.cf ok but after testing server configuration with SpamHaus tester the verify log reveal in the rejected message the service name:

550 5.7.1 Service unavailable; client [199.168.89.90] blocked using xxxxxxxxxxxxx.zen.dq.spamhaus.net 

In the SpamHaus Manual about postfix server configuration require more instruction for hiding the DQS key in the response message.

I suggest to disable DQS usage until new update.

Moooohooo
Regards
Alessandro
```

### Logs:

```plain text
N/A
```

### Steps to reproduce:

```plain text
N/A
```

### Which branch are you using?

master

### Operating System:

N/A

### Server/VM specifications:

N/A

### Is Apparmor, SELinux or similar active?

N/A

### Virtualization technology:

N/A

### Docker version:

N/A

### docker-compose version or docker compose version:

N/A

### mailcow version:

2023-07a

### Reverse proxy:

N/A

### Logs of git diff:

```plain text
N/A
```

### Logs of iptables -L -vn:

```plain text
N/A
```

### Logs of ip6tables -L -vn:

```plain text
N/A
```

### Logs of iptables -L -vn -t nat:

```plain text
N/A
```

### Logs of ip6tables -L -vn -t nat:

```plain text
N/A
```

### DNS check:

```plain text
N/A
```

sriccio:
Yes, this is a huge issue as your secret key can then be used by anyone.

That is because the postfix config is missing the postscreen reply map to hide the key.

### This should ideally be added to `data/conf/postfix/main.cf` (or `extra.cf` as a temp. workaround until it gets fixed)

```
[...]

PR #5357, opened 2023-08-01
Add postscreen_dnsbl_reply_map to avoid disclosure of DQS key

DocFraggle:
See docs at https://portal.spamhaus.com/dqs/#3.1.2

Mentioned also in #5353


FreddleSpl0it:
Thanks for the PR :)

I've made some changes to the postfix.sh script in staging. Can you double check if your PR has any conflicts?
Will then merge your PR.

DocFraggle:
OK @FreddleSpl0it, that should do it :)
I'm not quite sure if all of the reply domains are still needed after your change, but it won't hurt :D

Issue #3167, opened 2019-11-24
Postfix parameter "mail_name" allows a simple identification of a Mailcow system


christianbur:
With the [commit (16.10.2018)](https://github.com/mailcow/mailcow-dockerized/commit/d99b8aaf6952170218bb8c3b73f0025d0fdf32b6) the Postfix parameter "mail_name" was set to "Postcow" ([default "Postfix"](http://www.postfix.org/postconf.5.html)). 

Thus, the SMTP greeting banner of every Mailcow system now displays "Postcow". This allows a simple identification of all Mailcow systems.
If we now have a vulnerability in the system, the Mailcow systems can be more easily attacked because they do not get lost in the mass.  

[I currently find 15,251 Mailcow systems worldwide (8,676 are in Germany).](https://www.shodan.io/search?query=Postcow) 

If we would use the standard "Postfix", the Mailcow systems would be listed under the [3,049,901 Postfix systems. ](https://www.shodan.io/search?query=Postfix)

I know it's security by obscurity, but every little thing helps make a system more secure. 
How do you think about that?

andryyy:
It was implemented to only change our own headers when anonymizing to not break signatures. 

I understand your point but prefer to keep it.

mkuron:
You could just check the HTTPS server for each Postfix instance found by Shodan. If it exists and the page contains "mailcow UI", it's a Mailcow. Of course that's more work than actually searching for _Postcow_ on Shodan, but it's perfectly possible to do that for all 3 million Postfix servers.

christianbur:
What if we only change "smtpd_banner"?

current:
smtpd_banner = $myhostname ESMTP $mail_name

Future (example):
smtpd_banner = $myhostname ESMTP Postfix

Info: postscreen adopts the smtpd_banner setting through the already existing config line "postscreen_greet_banner = $smtpd_banner".

http://www.postfix.org/postconf.5.html

christianbur:
@mkuron 
Correct, but some (like me, for example) use a proxy for port 80/443 and the website doesn't listen to the MX record names.

mkuron:
> What if we only change "smtpd_banner"?

That would work.

patschi:
Wouldn't that still break the anonymization? 
https://github.com/mailcow/mailcow-dockerized/blob/master/data/conf/postfix/anonymize_headers.pcre

christianbur:
With mail_name this should be the case with smtpd_banner it should work without problems.

[...]

PR #5372, opened 2023-08-04
[Postfix] fix extra.cf updating

Habetdin:
I was trying to extend the postfix configuration by adding custom directives to extra.cf, and found out that some of them were disappearing on postfix-mailcow restarts. Specifically, this line:
`smtpd_banner = $myhostname ESMTP Postfix`
Same problem is described [here](https://github.com/mailcow/mailcow-dockerized/issues/3167#issuecomment-1639213527).

That happens due to untied `/myhostname/d` sed operation in `data/Dockerfiles/postfix/postfix.sh` - it deletes not only the `myhostname = example.org` line before updating it, but also lines like mentioned above.

This PR fixes the issue by ignoring lines that contain `$myhostname` (variable, not a directive) when cleaning up the `extra.cf`.

### DIFF
--- a/data/Dockerfiles/postfix/postfix.sh
+++ b/data/Dockerfiles/postfix/postfix.sh
@@ -486,7 +486,7 @@ fi
 # Append user overrides
 echo -e "\n# User Overrides" >> /opt/postfix/conf/main.cf
 touch /opt/postfix/conf/extra.cf
-sed -i '/myhostname/d' /opt/postfix/conf/extra.cf
+sed -i '/\$myhostname/! { /myhostname/d }' /opt/postfix/conf/extra.cf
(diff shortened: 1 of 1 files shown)


sriccio:
Shouldn't `extra.cf`  basically left untouched by any install/update scripts ? as it should be only user overrides...

Habetdin:
@sriccio yeah, I agree that the `myhostname` should be added to `main.cf` in different way.

Current method was introduced in [this commit](https://github.com/mailcow/mailcow-dockerized/commit/835c4eebb86a32ab6ff7a1cfe744654265b27028) back in the end of 2020, I don't know why it was done in that way.

I didn't change the current behaviour, just fixed the issue it has.

PatrickDeZordo:
same error here, please change it to work like mentioned in [https://docs.mailcow.email/de/manual-guides/Postfix/u_e-postfix-extra_cf/](url)

Thx! 🤞

DerLinkman (approved):
Looks very good! Thank you!

PR #5401, opened 2023-09-02
Improved the FQDN check and Ask before changing Git Repository URL in "update.sh"

AlexHuebi:
I've added a check for if the last letter in MAILCOW_HOSTNAME is a dot (because if there is one, Postfix doesn't start.

I've also added a bypass if the case in https://github.com/mailcow/mailcow-dockerized/issues/1572 (starting the updater while no subdomain is set eg. "example.org" instead of "mail.example.org") is met. 
The user has to confirm the Message (as seen in the commit) with a link to the issue. 

A better way would be to use a Regex for the check, but as seen here https://stackoverflow.com/a/26618995 , may require a much larger codechange.

### DIFF
--- a/generate_config.sh
+++ b/generate_config.sh
@@ -109,9 +109,24 @@ echo "Press enter to confirm the detected value '[value]' where applicable or en
 while [ -z "${MAILCOW_HOSTNAME}" ]; do
   read -p "Mail server hostname (FQDN) - this is not your mail domain, but your mail servers hostname: " -e MAILCOW_HOSTNAME
   DOTS=${MAILCOW_HOSTNAME//[^.]};
-  if [ ${#DOTS} -lt 2 ] && [ ! -z ${MAILCOW_HOSTNAME} ]; then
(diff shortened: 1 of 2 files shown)


dkpost3:
image

AlexHuebi:
I've created a VM where a Mailcow instance was set up, but I am not able to recreate that error. Did you change anything that is not in the .gitignore? 
And how did you get to this point? Because I had to modify the update.sh in my repository (as seen in the 2nd commit that should be probably removed before merge) so it wont change itself back to the mailcow/mailcow-dockerized repo.

AlexHuebi:
Note: I've now also modified the part where it says `# Silently fixing remote url from andryyy to mailcow` and now its asking before changing it back to the default repository

DerLinkman (commented):
Looks fine to me but need to test it. But don't have time at the moment.

DerLinkman (approved):
Thank you! Looks good. Just tweaked it a bit to be more colorful and vibrant to notice the users even better.

I also added the DOT Check to the generate_config.sh as it was used their as well (the old version).

PR #5402, opened 2023-09-02
enable search in bodies from EAS

cero1988:
enable search in bodies from EAS

### DIFF
--- a/data/conf/sogo/sogo.conf
+++ b/data/conf/sogo/sogo.conf
@@ -83,6 +83,7 @@
   //SoDebugBaseURL = YES;
   //ImapDebugEnabled = YES;
   //SOGoEASDebugEnabled = YES;
+  SOGoEASSearchInBody = YES
   //LDAPDebugEnabled = YES;
   //PGDebugEnabled = YES;
   //MySQL4DebugEnabled = YES;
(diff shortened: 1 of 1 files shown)


data/conf/sogo/sogo.conf
```diff
@@ -83,6 +83,7 @@
   //SoDebugBaseURL = YES;
   //ImapDebugEnabled = YES;
   //SOGoEASDebugEnabled = YES;
+  SOGoEASSearchInBody = YES
```
MAGICCC:
```suggestion
  SOGoEASSearchInBody = YES;
```

MAGICCC (changes requested):

DerLinkman:
Hi?

Could you please write at least your motivation or something else for this pr instead of : topic = discription?

Did you have tests for this or else?

cero1988:
Hello,

I am testing this in my home mailcow server and It works.
I had configured my email via EAS protocol, and I needed to allow search even on message body.
Without this parameter, I could not search via EAS in message body, the only way for searching in messages bodies was to use webmail.
I think that this behaviour could be useful enabled by default.

DerLinkman (approved):
Looks fine!

PR #5453, opened 2023-10-12
Allow suppressing watchdog start notification

smarsching:
The default behavior is still the old one (send a notifcation when the watchdog is started), but this notification can now be suppressed by setting `WATCHDOG_NOTIFY_START=n`. This is analogous to `WATCHDOG_NOTIFY_BAN`, the only difference being that `WATCHDOG_NOTIFY_BAN` uses a default value of `n` in `generate_config.sh`.

For `WATCHDOG_NOTIFY_START`, I think that a default value of `y` makes more sense, because for a new setup, one wants to see that watchdog mails can actually be delivered. Once this has been ascertained, the adminstrator can set this to `n` in order to not receive a notification every time the Docker stack is restarted (e.g. due to the host system being rebooted).


MAGICCC:
You also need to add an env variable in the docker-compose.yml file
https://github.com/mailcow/mailcow-dockerized/blob/master/docker-compose.yml#L478

smarsching:
@MAGICCC You are right, thank you for point this out. I just fixed it.

milkmaker:

FreddleSpl0it (approved):

PR #5459, opened 2023-10-12
Fix for git < 2.7

SecT0uch:
This change should be compatible with all git version. (get-url available from git 2.7)


milkmaker:
Thanks for contributing!

DerLinkman:
Does it work on other systems which are not using an old Git Version?

This needs to be 100% bulletproof otherwise it won't merged!

SecT0uch:
@DerLinkman yes it is:

```
$ git --version
git version 2.42.0
$ git config --get remote.origin.url
git@gitlab.com:fdroid/fdroidserver.git
$
```

SecT0uch:
@DerLinkman Any chance we can get that merged ?

Centos 7 users are having issues when running the update script. 
The version I committed is way more bulletproof than `get-url` which has been added later on

DerLinkman:
Ah yes. Sorry I missed the merge. Yes that works! Thanks

Issue #5457, opened 2023-10-12
Translation into Brazilian Portuguese (pt-br)


ghost:
### Summary

Hello, I would like to translate all of mailcow into pt-br (Brazilian Portuguese). However, there is no option for pt-br in Weblate. Could you please create the language so that I can begin the translation process?

### Motivation

Language translation included for Brazilian users.

MAGICCC:
I created a new translation for it: https://translate.mailcow.email/projects/mailcow-dockerized/translation/pt-br/
Do note that you can't validate any string just adding suggestions! There's an issue with weblate that validated strings will go in the live project even when nobody checked them. I will trust everyone that they enter valid things, but that's to avoid unnessary pushs on the main repo.

While you are on it, maybe you can also take a look into the normal portuguese translation? I assume it's almost the same?
https://translate.mailcow.email/projects/mailcow-dockerized/translation/pt-pt/

ghost:
Thank you. I will handle the translation carefully to ensure accuracy and quality.

I understand the caution regarding validating strings and will be mindful of it. I'll be translating the content diligently and will not suggest any validations without proper review.

Additionally, I'll also take a look at the Portuguese (Portugal) translation to identify any differences and make necessary adjustments if needed. Collaboration between different Portuguese variants is crucial for a comprehensive translation.

I'll get to work on this promptly and will keep you updated on my progress.

Best regards,
Pedro Lucca S.C

PR #5468, opened 2023-10-13
ADD Brazilian Portuguese Translation

ghost:
Initial issue: #5457 

Let me know if any update is necessary.


MAGICCC:
Nice, but lang.pt-br.json should be the right file for brazilian portuguese, pt-pt is normal portuguese.

PR #5474, opened 2023-10-16
Add switch to skip fetching auto{config,discover} subdomains

schichtnudelauflauf:
Add switch to skip fetching auto{config,discover} subdomains.

Might be useful in certain reverse proxy situations.


milkmaker:
Thanks for contributing!

DerLinkman:
If you have a reverse proxy setup you have to share the certs from the reverse proxy with mailcow.

See: https://docs.mailcow.email/post_installation/firststeps-ssl/?h=cer#how-to-use-your-own-certificate

In the past curious things happend if you used a seperate certificate for a reverse proxy instead of mailcows cert.

We need to take a deeper look into it though.

schichtnudelauflauf:
The whole reason is, that the reverse proxy also deals with other services and is its own VM. I do not want to share a trust boundary there.
But for reasons they share only one IPv4.
Thats why port 80,443 are blocked by the reverse proxy.
I did a passthrough for the .well-known/acme-challenge paths relevant to mailcow on port 80.
But this reverse proxy can not passthrough tls according to sni sniffing. It can only terminate tls.
So I made it fetch certs for mailstuff by itself and webstuff gets terminated by the reverse proxy.
The proxy and mailcow do not have overlapping SAN+CN in their cert.

Edit( Clarification about trust boundary. The proxy can reach into mailcow by being proxy for the admin interface and being first in the path to be able to put acme tokens, but mailcow should not be able to reach into the proxy. )

docker-compose.yml
```diff
@@ -405,6 +405,7 @@ services:
         - LOG_LINES=${LOG_LINES:-9999}
         - ACME_CONTACT=${ACME_CONTACT:-}
         - ADDITIONAL_SAN=${ADDITIONAL_SAN}
+        - ACME_IGNORE_AUTOCONFS=${ACME_IGNORE_AUTOCONFS}
```
mkuron, line 408: Needs a default value (`${ACME_IGNORE_AUTOCONFS:-n}`) for people who don't run the update script.

mkuron (commented):
Sorry for the late review, I was out for a few days. The implementation looks mostly okay to me. I'm just not completely happy with the name `ACME_IGNORE_AUTOCONFS`, please reconsider that one.

jennypaxian:
> Sorry for the late review, I was out for a few days. The implementation looks mostly okay to me. I'm just not completely happy with the name `ACME_IGNORE_AUTOCONFS`, please reconsider that one.

Just a quick headsup that we are also using this patch for our infrastructure since we manage HTTP(s) centrally via our loadbalancers.

Could you elaborate on the change(s) needed to get this upstream, especially help us out with the naming conventions for config params?

Thanks a lot,

Jenny Paxian

schichtnudelauflauf:
Check this newer version of this PR.

This one can be closed.

https://github.com/mailcow/mailcow-dockerized/pull/5838

PR #5487, opened 2023-10-23
Add a helper script for generating CAA records

artemislena:
T.: So, the other day we figured out how ya could get the Let's Encrypt account ID for the CAA `accounturi` parameter (after asking for't n no one knowing how), now someone else asked for that too n then we were asked for making a helper script so everyone could use't. So, we made one, basically it's just forked from https://github.com/diafygi/acme-tiny, w major portions removed, n the output changed.


milkmaker:
Thanks for contributing!

PR #5503, opened 2023-10-30
[update.sh] fix getting current repo url on older git versions

feldsam:
get-url subcommand is not available on older git versions


feldsam:
duplicate of #5459

PR #5512, opened 2023-11-02
Fixed SQL query for retrieving SSO users.

GeertHauwaerts:
# Issue

When retrieving users via SSO, the as-is query always returns the first user. This results in the problem where only a single mailbox is created (first SSO user), and no mailboxes are created for additional users.

# Resolution

Modified the query so that the last `OR` query is part of the last `AND` statement, otherwise the last `OR` always returns true.


milkmaker:
Thanks for contributing!

FreddleSpl0it (approved):

PR #5521, opened 2023-11-09
Add `--dry` IMAPsync Parameter as Button to select for SyncJobs

raph-topo:
Without this, setting Edit sync job > Custom parameters > `--dry` fails with:
image
(Even though `--dry` is a placeholder suggestion)


MAGICCC:
I guess we can also allow `--releasecheck` @DerLinkman

raph-topo:
I also added the very useful `--f1f2` in my instance, what do you think about that?

DerLinkman:
> I guess we can also allow `--releasecheck` @DerLinkman

Nah that is not necessarily i think.

DerLinkman:
`--dry` is not a problem. That dry option might make more sense as a clickable selection button in my eyes though.

DerLinkman:
Tested it and it works but don't trust my php skills here. @FreddleSpl0it can you take a look at this whole thing?

DerLinkman:
Looks like this now:

image

(Enabled for the Screenshot, disabled per default)

DerLinkman:
Btw, don't see a problem why `--f1f2` should not be allowed.

raph-topo:
Thank you for having accepted and improved this!

Issue #5529, opened 2023-11-13
Domain Wide Disclaimer breaks attachments visualization on Gmail and Outlook


thomisus:
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... I understand that not following the below instructions will result in immediate closure and/or deletion of my issue.
- [X] ... I have understood that this bug report is dedicated for bugs, and not for support-related inquiries.
- [X] ... I have understood that answers are voluntary and community-driven, and not commercial support.
- [X] ... I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

### Description

```plain text
Enabling Domain Wide Disclaimer breaks visualization of email and attachment in Gmail, sent from SOGO web ui.
Disabling does not fix the error.
```

### Logs:

```plain text
`
MIME-Version: 1.0
Message-ID: <46-65524400-f-37436000@186720726>
Subject: test email
User-Agent: SOGoMail 5.9.0
X-Last-TLS-Session-Version: None

------=_=-_OpenGroupware_org_NGMime-70-1699890218.060232-8------
Content-Type: multipart/alternative; boundary="----=_=-_OpenGroupware_org_NGMime-70-1699890218.059984-7------"

------=_=-_OpenGroupware_org_NGMime-70-1699890218.059984-7------
Content-Type: text/plain; charset=utf-8
Content-Length: 16


test attachment

------=_=-_OpenGroupware_org_NGMime-70-1699890218.059984-7------
Content-Type: text/html; charset=utf-8
Content-Length: 28

<html>test attachment</html>

------=_=-_OpenGroupware_org_NGMime-70-1699890218.059984-7--------
------=_=-_OpenGroupware_org_NGMime-70-1699890218.060232-8------
Content-Type: application/pdf
Content-Transfer-Encoding: base64
Content-Disposition: attachment; filename="invoice.pdf"
Content-Length: 110140


------=_=-_OpenGroupware_org_NGMime-70-1699890218.060232-8--------
`


`
MIME-Version: 1.0
Message-ID: <46-65524480-17-37436000@186722145>
Subject: test email
User-Agent: SOGoMail 5.9.0
X-Last-TLS-Session-Version: None

[...]

PR #5546, opened 2023-11-20
Fix: Domain Wide Disclaimer breaks attachments visualization on Gmail and Outlook #5529

FreddleSpl0it:
When Rspamd function lua_mime.add_text_footer(task, html_footer, text_footer) (https://rspamd.com/doc/lua/lua_mime.html#fa3d82) adds the Domain Wide Disclaimer, it will return new content types for rewriting. These new content types are somehow prefixed with "--\x0D\x0A" which breaks attachments visualization on Gmail and Outlook. 
This PR will remove the prefix before rewriting the content type as a workaround.

### DIFF
--- a/data/conf/rspamd/lua/rspamd.local.lua
+++ b/data/conf/rspamd/lua/rspamd.local.lua
@@ -631,15 +631,19 @@ rspamd_config:register_symbol({
             end
             local out_parts = {}
             for _,o in ipairs(out) do
-               if type(o) ~= 'table' then
-                 out_parts[#out_parts + 1] = o
-                 out_parts[#out_parts + 1] = newline_s
-               else
-                 out_parts[#out_parts + 1] = o[1]
-                 if o[2] then
-                   out_parts[#out_parts + 1] = newline_s
-                 end
-               end
+              if type(o) ~= 'table' then
+                out_parts[#out_parts + 1] = o
+                out_parts[#out_parts + 1] = newline_s
+              else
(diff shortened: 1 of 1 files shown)


thomisus (commented):
Unfortunately, this fix introduce another bug.
Domain Wide Disclaimer now is "inline" with message instead of on a new line.
At least it should be after a new blank line .

thomisus:
Actually,  the domain wide disclaimer is added outside `<html></html> ` tags, so it result displayed inline instead of on a new line.

FreddleSpl0it:
> Domain Wide Disclaimer now is "inline" with message instead of on a new line. 

This was also the case before the fix. I'm not sure if this is a big problem. You can add a `newline` for plain text or a `<br>` for html at the beginning.

> Actually, the domain-wide disclaimer is added outside the <html></html> tags, so it ends up being displayed inline instead of on a new line. 

This is how the Rspamd function behaves (https://rspamd.com/doc/lua/lua_mime.html#fa3d82), and unfortunately, I can't fix this.

thomisus:
It's not a big problem I think, but to be 100% correct footer should be inside `<html></html>` tags.
Maybe this is a bug in raspamd and not in mailcow.

Best regards



```
------=_=-_OpenGroupware_org_NGMime-66-1700652219.440647-1------
Content-Type: text/html; charset=utf-8
Content-Transfer-Encoding: 7bit

<html>test</html>

<br>
<br/>
FOOTER

```

PR #5554, opened 2023-11-22
Support automatic mariadb-upgrade execution on mariadb upgrades

MaxXor:
To prepare the upgrade to the next mariadb lts version 10.11 I think it makes sense to add the newly available environment variable `MARIADB_AUTO_UPGRADE` which will execute `mariadb-upgrade` in the container, to perform necessary upgrades in the MariaDB system tables if necessary.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -32,6 +32,7 @@ services:
         - MYSQL_USER=${DBUSER}
         - MYSQL_PASSWORD=${DBPASS}
         - MYSQL_INITDB_SKIP_TZINFO=1
+        - MARIADB_AUTO_UPGRADE=1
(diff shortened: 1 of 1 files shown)


DerLinkman:
Hi, our php-fpm container is doing a mysql upgrade once mysql is started.

So i think that is not needed.

See: https://github.com/mailcow/mailcow-dockerized/blob/979f5475c3c07f5a61fa2fa55c512532b64cc25f/data/Dockerfiles/phpfpm/docker-entrypoint.sh#L24C1-L24C1

MaxXor:
Thanks for the feedback. However I still think that looks rather complicated when it can be simplified?

PR #5577, opened 2023-12-07
[Rspamd] Fixed Ratelimit forced by global ratelimits

DerLinkman:
This PR will fix the "bug" which was introduced with 2023-11 which applied the global ratelimits which were meant to be not applied.

You can however setup your own global ratelimits inside the newly added ratelimit.conf in the local.d path of rspamd.


milkmaker:
Thanks for contributing!

PR #5313, opened 2023-07-10
[Web] add f2b_banlist endpoint

FreddleSpl0it:
Add an api endpoint to retrieve a plain text list of blacklisted and active banned IPs.

Also add an option to manage F2B externally. If enabled, Netfilter will not ban or permBan any sources. Unbanning still works in case of leftover iptable rules.

PR #5585, opened 2023-12-11
[Netfilter] add nftables support

FreddleSpl0it:
This PR is based on @amorfo77 work https://github.com/mailcow/mailcow-dockerized/pull/5049.
IPTable and NFTable functions are now separated in classes.


data/Dockerfiles/netfilter/docker-entrypoint.sh
RafaelKr, line 1: I think some comments could be useful. Also I think lines 15-17 aren't required, when `iptables` is set on line 3 anyways?

MAGICCC:
Merged in
https://github.com/mailcow/mailcow-dockerized/pull/5313

PR #5219, opened 2023-05-02
armcow64: dockerized 

DerLinkman:
## Description:
This PR adds the ARM64 compatibility for our lovely mailcow: dockerized.

ARM64 does not need a manual configuration change in the mailcow Configs, all changes to be ARM64 compatible (not much) are **included into this PR**.

## Phase Status:
1. Port mailcow to ARM64:  
2. Test all components of mailcow on a ARM64 Server: 
3. Publish Images for public tests: 
4. Update Documentation Pages for ARM64: 
5. Merges and final Tests: 

## Latest Events:
+ August 7st 2023: We stumbled across a big issue with decrypting migrated e-mails with the new Alpine based dovecot images. Older E-Mails are not decryptable, new once are. It's not 100% clear why this happens but right know it looks like a non completed implementation of OpenSSL within Dovecot which is causing this strange behaviour as the new OS Versions (Alpine 3.17+ and Debian 12+ are now shipped with OpenSSL 3.X instead of 1.1.1x)
+ August 2nd 2023: Within the latest Dev Build from Rspamd (3.6) the issue with Hyperscan/Vectorscan has finally been solved. Now we can concentrate on optimizing the rest.
+ May 5th 2023: We've published the first ARM64 images (for testing to Docker Hub) which means that you can now test the ARM64 Port of mailcow. Simply clone mailcow as you normally would on a ARM64 Device and switch the branch with `git checkout feat/arm64-cow`. To generate/update your mailcow please use the `--dev` Parameter alongside the desired script to keep the current branch checked out instead of the master/nightly branch.

## Estimated Release:
TBA

PR #5587, opened 2023-12-11
mailcow Multiarch (x86 and ARM64) support

### LINKED ISSUE #879: ARM64 Support
I have extensive experience with docker on arm, so before I get to work on this, has anyone got mailcow-dockerized ported or running on arm ?

Did a quick search through the repo and I do not see any references to arm. 

Does the repo owner have any issues against arm support being added ?

Thanks

DerLinkman:
## Description:
This PR adds the ARM64 compatibility for our lovely mailcow: dockerized.

ARM64 does not need a manual configuration change in the mailcow Configs, all changes to be ARM64 compatible (not much) are **included into this PR**.

## Phase Status:
1. Port mailcow to ARM64:  
2. Test all components of mailcow on a ARM64 Server: 
3. Publish Images for public tests: 
4. Update Documentation Pages for ARM64: 
5. Merges and final Tests: 

## Latest Events:
+ December 11th: Cleanups to the code has been done, that's the reason for this new PR with the same content.
+ August 7st 2023: We stumbled across a big issue with decrypting migrated e-mails with the new Alpine based dovecot images. Older E-Mails are not decryptable, new once are. It's not 100% clear why this happens but right know it looks like a non completed implementation of OpenSSL within Dovecot which is causing this strange behaviour as the new OS Versions (Alpine 3.17+ and Debian 12+ are now shipped with OpenSSL 3.X instead of 1.1.1x)
+ August 2nd 2023: Within the latest Dev Build from Rspamd (3.6) the issue with Hyperscan/Vectorscan has finally been solved. Now we can concentrate on optimizing the rest.
+ May 5th 2023: We've published the first ARM64 images (for testing to Docker Hub) which means that you can now test the ARM64 Port of mailcow. Simply clone mailcow as you normally would on a ARM64 Device and switch the branch with `git checkout feat/arm64-cow`. To generate/update your mailcow please use the `--dev` Parameter alongside the desired script to keep the current branch checked out instead of the master/nightly branch.

## Estimated Release:
January 17th 2024

### DIFF
--- a/data/Dockerfiles/acme/Dockerfile
+++ b/data/Dockerfiles/acme/Dockerfile
@@ -1,6 +1,6 @@
 FROM alpine:3.19
 
-LABEL maintainer "The Infrastructure Company GmbH <info@servercow.de>"
+LABEL maintainer "The Infrastructure Company GmbH GmbH <info@servercow.de>"
 
 ARG PIP_BREAK_SYSTEM_PACKAGES=1
 RUN apk upgrade --no-cache \
--- a/data/Dockerfiles/clamd/Dockerfile
+++ b/data/Dockerfiles/clamd/Dockerfile
@@ -1,12 +1,14 @@
-FROM clamav/clamav:1.0.3_base
+FROM alpine:3.19
 
-LABEL maintainer "The Infrastructure Company GmbH <info@servercow.de>"
+LABEL maintainer "The Infrastructure Company GmbH GmbH <info@servercow.de>"
 
 RUN apk upgrade --no-cache \
   && apk add --update --no-cache \
   rsync \
+  clamav \
   bind-tools \
-  bash 
+  bash \
+  tini
 
 # init
 COPY clamd.sh /clamd.sh
@@ -14,7 +16,9 @@ RUN chmod +x /sbin/tini
 
 # healthcheck
 COPY healthcheck.sh /healthcheck.sh
+COPY clamdcheck.sh /usr/local/bin
 RUN chmod +x /healthcheck.sh
+RUN chmod +x /usr/local/bin/clamdcheck.sh
 HEALTHCHECK --start-period=6m CMD "/healthcheck.sh"
 
 ENTRYPOINT []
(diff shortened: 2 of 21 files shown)


DerLinkman:
Originated from: https://github.com/mailcow/mailcow-dockerized/pull/5219

aSauerwein:
hi!
thanks for your work!
currently I'm unable to test armcow64, because the image tags in the current docker-compose.yaml have no arm64 architecture. 
is there any workaround, except rolling back to the previous docker-compose.yaml ?

nbx3:
Why is solr planned for removal?

DerLinkman:
> Why is solr planned for removal?

It's to old and deprecated within Dovecot + Too Resource intensive nowadays.

grusch-it:
Since the official SOGo package repository was swapped out for this PR in favor of the (for me unkown) package repository `http://www.axis.cz/linux/debian`, are there any more information about the maintainer of this repo? How about trustworthiness?

Visiting www.axis.cz shows an quite old fashioned website (what is not something bad at all) were I can't find any information about their debian package repository (neither motivation or anything else). Maybe I just can't find anything because the website is only available in czech language.

Since I guess this change to the repository was just made to support ARM: is it really worth the risk to use (unkown) third-party package repositories for all, even if someone is not using the ARM docker images at all?

Edit: Maybe ARM support could be achieved using the new official packaging of SOGo: https://bugs.sogo.nu/view.php?id=6020?

DerLinkman:
Yes it was originally planned to used sogos packages and we'll swap back to them. At the time we were building arm support there was this offer so we kept this.

Arm64 is pretty common in usage of mailcow since it's release so that was definately worth it.

We can try switching back to sogos official nightly packages in one of the next releases of mailcow.

grusch-it:
@DerLinkman Nice to hear that.

To use the latest SOGo nightly build, one would have to just swap out the Dockerfile and set it back to the old version which is using the SOGo package repository as long as ARM64 support is not needed, right?

DerLinkman:
Yes, that should work.

PR #5525, opened 2023-11-12
Update Dockerfiles to Alpine 3.18

MAGICCC:
Updates all relevant Dockerfiles to Alpine 3.18

PR #5592, opened 2023-12-13
Update Dockerfiles to Alpine 3.19

MAGICCC:
Updates all relevant Dockerfiles to Alpine 3.19
For images with python I added `PIP_BREAK_SYSTEM_PACKAGES` variable to 'fix' https://peps.python.org/pep-0668/

### DIFF
--- a/data/Dockerfiles/acme/Dockerfile
+++ b/data/Dockerfiles/acme/Dockerfile
@@ -1,7 +1,8 @@
-FROM alpine:3.17
+FROM alpine:3.19
 
 LABEL maintainer "The Infrastructure Company GmbH <info@servercow.de>"
 
+ARG PIP_BREAK_SYSTEM_PACKAGES=1
 RUN apk upgrade --no-cache \
   && apk add --update --no-cache \
   bash \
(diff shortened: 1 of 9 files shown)


MAGICCC:
Superseeds https://github.com/mailcow/mailcow-dockerized/pull/5525

MAGICCC:
Works fine on my server

DerLinkman:
Not merging this for now.

Currently adding the missing pip packages to the Alpine Repo themselves to not use the PIP_BREAK_SYSTEM_PACKAGES method as i don't like this :/

DerLinkman:
Will modify your commit though when these packages are included into Alpine.

DerLinkman:
However we'll use the PIP_BREAK_SYSTEM_PACKAGES flag for olefy as it is using the olefy tools which they install manually via pip.

MAGICCC:
Sounds good!

DerLinkman:
Will merge this after all as we can swap out the native alpine packages in the long run.

DerLinkman (approved):
Looks fine now!

PR #5523, opened 2023-11-12
Implemented Server Side processing for domains and mailboxes datatables

feldsam:
When there is a lot of domains and/or mailboxes, tables are loading very slowly, for ex. on one server with 516 domains, it takes almost five seconds. So I decided to leverage datatables SSP functionality, to speedup things. Only drawback is, that there is no sorting capabilities on all columns, because lot of data need to be loaded from different tables. Search is also working server side.

PR #5597, opened 2023-12-15
[Web] Fix sorting mailboxes by in use percent

pave1h:
There seems to have been a previous attempt to fix this, but for me the mailboxes are still being sorted alphabetically (in descending order e.g. 90%, 9%, 80%, 8%, etc.). Changing the type of the data column to "num" fixed it for me.

There is now also another issue: mailboxes with infinite quota are sorted arbitrarily because their In use % is not a number but "- ". I propose fixing this by changing the API to return 0 as used percent for such mailboxes, but I have not included the change here because I don't know if it could break something else.

### DIFF
--- a/data/web/js/site/mailbox.js
+++ b/data/web/js/site/mailbox.js
@@ -1114,7 +1114,8 @@ jQuery(function($){
           title: "",
           data: 'in_use.sortBy',
           defaultContent: '',
-          className: "d-none"
+          className: "d-none",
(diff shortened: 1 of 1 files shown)


FreddleSpl0it:
Thank you for the pull request, but I think this will be obsolete once https://github.com/mailcow/mailcow-dockerized/pull/5523 is merged.

PR #5621, opened 2023-12-29
[Postfix] Remove pipeling from ehlo keywords as we block it in data

dragoangel:
Related to [CVE-2023-51764](https://security-tracker.debian.org/tracker/CVE-2023-51764) and to the fact we already have set `smtpd_data_restrictions = reject_unauth_pipelining` to not propose clients what we not allow.


milkmaker:
Thanks for contributing!

PR #5624, opened 2023-12-30
Add new SOGoMailHideInlineAttachments option to sogo.conf

ghost:
Hello,

This pull request fixes a [long lasting SOGo usability issue](https://bugs.sogo.nu/view.php?id=5490). `SOGoMailHideInlineAttachments = YES;` allows to hide inline (body and footer) images being shown as attachments.

It makes things much more easier for end users, while before that they had all the images from their HTML email signatures showing up in the file attachments lists. Making things very confusing, duplicating while receiving every responses and growing as the email thread discussion continue.

Users would loose a lot of time searching through this list to find the proper PDF document attachment they are looking for. I think enabling this new option by default would improve the usability of Mailcow and will be beneficiary to all users.

Thanks in advance.

### DIFF
--- a/data/conf/sogo/sogo.conf
+++ b/data/conf/sogo/sogo.conf
@@ -12,6 +12,7 @@
     SOGoJunkFolderName= "Junk";
     SOGoMailDomain = "sogo.local";
     SOGoEnableEMailAlarms = YES;
+    SOGoMailHideInlineAttachments = YES;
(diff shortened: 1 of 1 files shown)


ghost:
Thanks a lot @DerLinkman!

boospy:
Very thanks. That helped a lot! :+1:

PR #5639, opened 2024-01-15
unbound: rewrote of healthcheck

DerLinkman:
This PR includes a rewrote of the unbound healthchecks.

Instead of simply checking the DNS Resolution of unbound the check is also checking the ability to ping the hosts which are resolved to test as well as doing a simple 80 443 accessibility check within unbound to ensure a connection can be made.


milkmaker:
Thanks for contributing!

PR #5640, opened 2024-01-15
Fix/repetitive code

SnailShea:
Created a few functions to remove repeated use of the same lines of code throughout json_api.php. The UI hasn't broken after a full `docker compose down && docker compose up -d`, and I was able to create a user via API, so I believe this works. I have clicked around pretty much every page in the UI, but I have not tested all API endpoints to verify fully.


DerLinkman (commented):
Thanks for your PR!

Please test your own PR completely to be ready for merge.

Issue #5568, opened 2023-11-28
Attempt Base64-Decoding of mailcow_password when using Mailpassword


mcmdev:
### Summary

Attempt to Base64-Decode the value of mailcow_password before denying login when using the Mailpassword Flow in the SSO feature of the nightly branch.

### Motivation

This would enable easy integration when using an LDAP Backend. The userPassword could simply be mapped to mailcow_password in Keycloak.

flischi:
+1 from me, this would really help! AFAIK all DS are responding to a userPassword request with a base64 encoded value. I've searched for something like Attribute Transformation in Keycloak but no luck. I'm wondering how the LDAP feature is intended to work with the mailcow_password then...

m-lindemann:
+1 same for me. Stuck at base64 encoded value from openldap.

jbheuschen:
+1 from me, any news on this/has anyone found a viable solution in the meantime?

PR #5653, opened 2024-01-19
Try to Base64-decode mailcow hash if not already

atoav:
This PR tries to address the problems described in issue #5568: When using a LDAP backend the string for `mailcow_password` might be base64-encoded. This PR adds a block that tries to base64-decode the stored hash or do nothing if this fails.

The processing steps are the following:

1. Check if the the string that should contain the hash starts with a hash-algorithm specifier (e.g. "{SHA}") by checking if it starts with "{". If it does not, do nothing and skip the next steps.
2. Try to base64-decode the string. If decoding fails, do nothing and skip the next steps.
3. Set mailcow_password to the base64-decoded version of the string


milkmaker:
Thanks for contributing!

milkmaker:
Thanks for contributing!

atoav:
The part of code my PR should touch, does not exist in the staging branch, nor did grepping for parts of it reveal any other place where this could be found. I guess we need to re-evaluate the staging version. As of now the PR can be closed.

PR #5658, opened 2024-01-21
New Feature: Quarantine Outgoing Mail

SnailShea:
I have implemented outbound quarantine scanning for mailcow as an optional feature. The UI and language files have been updated to reflect this feature, and the ability to enable/disable it. 

image

The  `data/conf/rspamd/meta_exporter/pipe.php` script will now check if the sender is a mailcow user, and will quarantine a message if the feature is enabled and the action is "reject".

Mail is stored per-recipient, similar to how incoming quarantined mail is processed. If the message being released (not via quick release) from quarantine is from a mailcow sender and a DKIM key exists for that domain, the message  will be signed before being released. I have tested this with various email formats (plain text emails, emails with only html, multipart emails with text and html, emails with inline images and attachments) via a few mail clients (SOGo, BetterBird, Evolution), and all released mail appears to leave in its original state. The already included `PhpMimeMailParser` library was used to load raw message contents, select message body components (text/html/etc). New headers were constructed using the message metadata before signing.

I decided that leveraging the existing PHP tooling in this context made more sense than setting up an additional SMTP port to sign outgoing mail via rspamd being released from quarantine, as the existing quarantine port (590) does not use milters.

Bonus: [Headers signed via PHPMailer with some data redacted](https://pastebin.com/WSmRSxmR)

### DIFF
--- a/data/conf/rspamd/meta_exporter/pipe.php
+++ b/data/conf/rspamd/meta_exporter/pipe.php
@@ -90,6 +90,63 @@ catch (RedisException $e) {
   exit;
 }
 
+// If the sender is managed by mailcow, quarantine and skip the rest
+// TODO: check if outgoing quarantine is enabled along with domain map
+try {
+  $domain = parse_email($sender)['domain'];
+  if ($redis->hGet('DOMAIN_MAP', $domain)) {
+    foreach (json_decode($rcpts, true) as $rcpt) {
+      error_log("QUARANTINE: quarantine pipe: processing quarantine message for rcpt " . $rcpt . PHP_EOL);
+      try {
+        $stmt = $pdo->prepare("INSERT INTO `quarantine` (`qid`, `subject`, `score`, `sender`, `rcpt`, `symbols`, `user`, `ip`, `msg`, `action`, `fuzzy_hashes`)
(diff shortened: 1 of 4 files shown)


milkmaker:
Thanks for contributing!

SnailShea:
I forget to check for staging branch almost every time :upside_down_face:

dragoangel:
@SnailShea I would add one thing to description: user still would think email was rejected and before releasing email admin should verify that user doesn't send email again.

dragoangel:
Also I am not sure if is okay to have the same limits per email (meaning we have limits to accept email per mailbox) here they are for opposite and maybe need to be handled separately? What I mean they have strongly increase db size if not be handled specially

SnailShea:
These are very good ideas, thank you. I will work on them as soon as I get the chance.

Edit: I think the description about verifying user doesn't send mail again should go at the top of the quarantine page, maybe after "Blacklisted elements are excluded from the quarantine.". Does this seem good? I could put it in the settings as well, but I would either need to create a new text field over the top of settings, or add way too much text I think next to the added checkbox.

dragoangel:
@SnailShea If implement special handling for outgoing mail - it can be dedicated section in quarantine settings and this info can be placed there optimally I think

milkmaker:

milkmaker:

PR #5660, opened 2024-01-22
Test for openrc configuration file instead of alpine

lu-zero:
This way other distro using openrc can be supported.


milkmaker:
Thanks for contributing!

DerLinkman:
Thanks!

lu-zero:
Thank you, sorry for not knowing about staging :)

PR #5661, opened 2024-01-22
rspamd: upgrade to rspamd 3.9.1

DerLinkman:
This PR updates ~~the Debian within the container to Debian 12~~ (freeze, needs more testing) and Updates Rspamd to Version 3.8.

~~Patchnotes for Rspamd 3.8: https://github.com/rspamd/rspamd/releases/tag/3.8.0~~

~~Patchnotes for Rspamd 3.8.4: https://github.com/rspamd/rspamd/releases/tag/3.8.4~~

Patchnotes for Rspamd 3.9.1: https://github.com/rspamd/rspamd/releases/tag/3.9.1


dragoangel:
Better wait for 3.9.0 or 3.8.5

In 3.8.x there was issues with bayes that not yet in any release.

data/Dockerfiles/rspamd/Dockerfile
```diff
@@ -11,7 +11,8 @@ RUN apt-get update && apt-get install -y \
   gnupg2 \
   apt-transport-https \
   dnsutils \
-  netcat \
+  netcat-traditional \
+  lua-cjson \
```
dragoangel, line 20: What is this for?

data/Dockerfiles/rspamd/Dockerfile
```diff
@@ -11,7 +11,8 @@ RUN apt-get update && apt-get install -y \
...
   && apt-key adv --fetch-keys https://rspamd.com/apt-stable/gpg.key \
   && echo "deb https://rspamd.com/apt-stable/ $CODENAME main" > /etc/apt/sources.list.d/rspamd.list \
```
dragoangel: We need align with new flow from: https://github.com/mailcow/mailcow-dockerized/blob/9c814cc18235208f574043211d3d8424c309ff39/data/Dockerfiles/rspamd/Dockerfile#L20-L22

But change varsion to `RSPAMD_VER=rspamd_3.9.0-1~fb1b597d7`

DerLinkman (commented):
The PR is also adding Spamhaus DQS Blacklists to add spam scores if it is used.

dragoangel (approved):

data/Dockerfiles/rspamd/Dockerfile
DerLinkman: Used internally for some lua scripts of rspamd. (the cjson)

flo-m:
> The PR is also adding Spamhaus DQS Blacklists to add spam scores if it is used.

Would it also make sense to add ABUSIX like SpamHaus?
https://abusix.com/docs/rspamd/

dragoangel:
> > The PR is also adding Spamhaus DQS Blacklists to add spam scores if it is used.
> 
> Would it also make sense to add ABUSIX like SpamHaus?
> https://abusix.com/docs/rspamd/

@flo-m . Abusix isn't free at all (free 5k queries per day is too low), no, it not makes sense.
If you willing to pay to Abusix to get proper queries and use it in mailcow, nobody blocking you from modifying rspamd rbl configuration and add all needed pieces.

Even when you will leave only their IP, domain and url rbls - one email in avg will eat 10-20 dns requests, 5000/20=250 emails per day, this really not covers anything... 10,000rq/d is already 10$, and this will cover only ~500 emails.

JFYI - Their docs example provide many things that not all needed in rspamd because it's detected by Rspamd with other things already, so some of their lists better just be skipped. About scores - will just say - score need be dropped in 3-4 times compared to what they wrote, Rspamd should not instantly reject something just because of 1 symbol, this not how proper antispam should work, to exclude FP.

PR #5662, opened 2024-01-22
fix: rollback curl bug

DerLinkman:
This PR rollsback the effected Alpine Images which cause trouble for some users.

Dovecot however cannot be rolled back due to the Alpine 3.19 dependency.

But Dovecot only uses curl to Download things from the internet which still works even in the affected images.


milkmaker:
Thanks for contributing!

PR #5681, opened 2024-01-30
php: added bz2 + Alpine 3.19

DerLinkman:
Fixes: https://github.com/mailcow/mailcow-dockerized/issues/5678

PR #5691, opened 2024-02-01
Enable bz2 in php container

krono:
Newer Nextcloud version will complain (not too loud, but will) if the php bz2 extension is missing in the  php-fpm container.

It's fairly easy to install after the fact, but with caveats, and seems to go away after an update.

Should not harm normal mailcow behavior and benefits nextcloud users.


MAGICCC:
Dupe of https://github.com/mailcow/mailcow-dockerized/pull/5681

PR #5635, opened 2024-01-10
[Postfix] Upgrade to Deb12 + PF to 3.7.10 & Drop TLS 1.0/1.1 per default

DerLinkman:
This PR upgrades Postfix underlaying OS to Debian 12 and updates Postfix to the latest Debian Repo Version.

It also drops TLS1.0 and 1.1 per default now.

PR #5694, opened 2024-02-01
Create extra.cf

ajnadox:
disabling weak ciphers and TLS versions below 1.2


FreddleSpl0it:
We won't create the extra.cf inside the repo as a tracked file... Also: This PR is already planned some kind in: https://github.com/mailcow/mailcow-dockerized/pull/5635

PR #5707, opened 2024-02-04
Add mjs to javascript mimetypes and serve mjs files

mstinsky:
Fix javascript detect in nextcloud >=28.0.2


milkmaker:

PR #5736, opened 2024-02-11
Netfilter Fix os.getenv has no attribute lower

thethink3r:
Netfilter crash
```text
netfilter-mailcow-1  | Using IPTables backend
netfilter-mailcow-1  | Clearing all bans
netfilter-mailcow-1  | Initializing mailcow netfilter chain netfilter-mailcow-1  | Traceback (most recent call last):
netfilter-mailcow-1  |   File "/app/main.py", line 418, in <module>
netfilter-mailcow-1  |     if os.getenv("DISABLE_NETFILTER_ISOLATION_RULE").lower() in ("y", "yes"):
netfilter-mailcow-1  |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
netfilter-mailcow-1  | AttributeError: 'NoneType' object has no attribute 'lower'
netfilter-mailcow-1 exited with code 1
```
os.getenv has no attribute lower

Fix. Cast to String.

### DIFF
--- a/data/Dockerfiles/netfilter/main.py
+++ b/data/Dockerfiles/netfilter/main.py
@@ -415,7 +415,7 @@ if __name__ == '__main__':
   tables.initChainIPv4()
   tables.initChainIPv6()
 
-  if os.getenv("DISABLE_NETFILTER_ISOLATION_RULE").lower() in ("y", "yes"):
+  if str(os.getenv("DISABLE_NETFILTER_ISOLATION_RULE")).lower() in ("y", "yes"):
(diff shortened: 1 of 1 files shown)


milkmaker:
Thanks for contributing!

FreddleSpl0it:
os.getenv already returns a string type https://docs.python.org/3/library/os.html#os.getenv.
I think the issue is, that the env var is not set so it returns None and then the `.lower()` call throws `AttributeError`.
So better check if os.getenv return value is not None before calling `.lower()`

PR #5802, opened 2024-03-24
Update heinlein-support.de URI

realizelol:
http is now redirected to https and cert is only valid for *.heinlein-support.de, so also remove "www."


dragoangel:
It is actually valid for both www and non www hostnames

realizelol:
Yes, you're right. The alternative name with "www." must have been added on: Mon, 25 Mar 2024 07:17:51 GMT
But the https:// should be needed to download the file.

dragoangel:
> Yes, you're right. The alternative name with "www." must have been added on: Mon, 25 Mar 2024 07:17:51 GMT But the https:// should be needed to download the file.

Curl follows 301 redirect without issues to download file. Just to clarify I'm not against pushing this PR.

realizelol:
Did you check the downloaded content of the http version?

```bash
curl --connect-timeout 15 --retry 10 --max-time 30 http://www.spamassassin.heinlein-support.de/$(dig txt 1.4.3.spamassassin.heinlein-support.de +short | tr -d '"' | tr -dc '0-9').tar.gz --output /tmp/sa-rules-heinlein.tar.gz
if gzip -t /tmp/sa-rules-heinlein.tar.gz; then echo valid; fi
```
> gzip: /tmp/sa-rules-heinlein.tar.gz: not in gzip format
```bash
curl --connect-timeout 15 --retry 10 --max-time 30 https://www.spamassassin.heinlein-support.de/$(dig txt 1.4.3.spamassassin.heinlein-support.de +short | tr -d '"' | tr -dc '0-9').tar.gz --output /tmp/sa-rules-heinlein-https.tar.gz
if gzip -t /tmp/sa-rules-heinlein-https.tar.gz; then echo valid; fi
```
> valid

MAGICCC:
I guess we should add a failsafe with curl -k in case SSL cert is invalid (yes not secure but better than dovecot not starting)

realizelol:
Should I add the "-k" flag or do you wait for response of someone else?
If you ask me it should not be added, maybe only if the content files don't exist in /etc/rspamd/custom/sa-rules.
If the cert is not valid, the file may also not be valid. (MitM-Attack)

dragoangel:
I agree that `-k` is bad and I would not do this, but we can skip failure if file already exists by catching error and checking if file exist goith futher otherwise throw error.

realizelol:
So sth. like this?:
```bash
if ! curl --connect-timeout 15 --retry 10 --max-time 30 https://spamassassin.heinlein-support.de/$(dig txt 1.4.3.spamassassin.heinlein-support.de +short | tr -d '"' | tr -dc '0-9').tar.gz --output /tmp/sa-rules-heinlein.tar.gz; then
  if [[ "${HASH_SA_RULES}" -eq 0 ]]; then
    echo "Error Heinlein's Spamassassin Rules couldn't be fetched."
    exit 1
  fi
fi
```

dragoangel:
> So sth. like this?:
> 
> ```shell
> if ! curl --connect-timeout 15 --retry 10 --max-time 30 https://spamassassin.heinlein-support.de/$(dig txt 1.4.3.spamassassin.heinlein-support.de +short | tr -d '"' | tr -dc '0-9').tar.gz --output /tmp/sa-rules-heinlein.tar.gz; then
>   if [[ "${HASH_SA_RULES}" -eq 0 ]]; then
>     echo "Error Heinlein's Spamassassin Rules couldn't be fetched."
>     exit 1
>   fi
> fi
> ```

I don't see where and what you plan to set in HASH_SA_RULES, but generally yes, something like this

realizelol:
In the head part of the file there's already a test:
```bash
# Hash current SA rules
if [[ ! -f /etc/rspamd/custom/sa-rules ]]; then
  HASH_SA_RULES=0
else
  HASH_SA_RULES=$(cat /etc/rspamd/custom/sa-rules | md5sum | cut -d' ' -f1)
fi
```
https://github.com/mailcow/mailcow-dockerized/blob/e91a09bfbf3a31c6642db55ffbc5a0cbe049a508/data/Dockerfiles/dovecot/sa-rules.sh#L6C1-L11C3

dragoangel:
Yeah, thanks, not looked at existing part

realizelol:
Okay, the code has been updated on my staging branch.

dragoangel:
LGTM

realizelol:
There's actually a new commit in staging: https://github.com/mailcow/mailcow-dockerized/commit/af626d98d31637b17ed11a4f96aecaa3e3c67082

`www.` seems to be legit now:
```
curl --connect-timeout 15 --retry 10 --max-time 30 https://www.spamassassin.heinlein-support.de/$(dig txt 1.4.3.spamassassin.heinlein-support.de +short | tr -d '"' | tr -dc '0-9').tar.gz --output /tmp/sa-rules-heinlein.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  115k  100  115k    0     0   526k      0 --:--:-- --:--:-- --:--:--  528k
if gzip -t /tmp/sa-rules-heinlein.tar.gz; then echo legit; fi
legit
```

So the leftover part is just the information that the file couldn't be downloaded.

Does it make any sense or should the PR just being closed by now?

DerLinkman:
Ah lel, did not saw this pr... sorry but came across this issue a few days ago by myself so i already fixed it in staging...

I'm so sorry.

PR #5804, opened 2024-03-26
Allow prompt-less install on low-resource systems

Ayowel:
Make it possible to generate config unattended on low resource systems by bypassing prompts if `SKIP_SOLR` or `SKIP_CLAMD` is non-empty.
This was tested with the mailcow-ansiblerole repository with the following patch:

```patch
diff --git a/tasks/main.yml b/tasks/main.yml
index a1263f2..bf8825a 100755
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -28,6 +28,8 @@
     MAILCOW_HOSTNAME: "{{ mailcow__hostname }}"
     MAILCOW_TZ: "{{ mailcow__timezone }}"
     MAILCOW_BRANCH: "{{ mailcow__git_version }}"
+    SKIP_SOLR: "{{ mailcow__config_skip_solr }}"
+    SKIP_CLAMD: "{{ mailcow__config_skip_clamd }}"
   args:
     executable: /bin/bash
     chdir: "{{ mailcow__install_path }}"
```

### DIFF
--- a/generate_config.sh
+++ b/generate_config.sh
@@ -147,40 +147,44 @@ done
 
 MEM_TOTAL=$(awk '/MemTotal/ {print $2}' /proc/meminfo)
 
-if [ ${MEM_TOTAL} -le "2621440" ]; then
-  echo "Installed memory is <= 2.5 GiB. It is recommended to disable ClamAV to prevent out-of-memory situations."
-  echo "ClamAV can be re-enabled by setting SKIP_CLAMD=n in mailcow.conf."
-  read -r -p  "Do you want to disable ClamAV now? [Y/n] " response
-  case $response in
-    [nN][oO]|[nN])
-      SKIP_CLAMD=n
+if [ -z "${SKIP_CLAMD}" ]; then
+  if [ ${MEM_TOTAL} -le "2621440" ]; then
+    echo "Installed memory is <= 2.5 GiB. It is recommended to disable ClamAV to prevent out-of-memory situations."
+    echo "ClamAV can be re-enabled by setting SKIP_CLAMD=n in mailcow.conf."
+    read -r -p  "Do you want to disable ClamAV now? [Y/n] " response
+    case $response in
+      [nN][oO]|[nN])
+        SKIP_CLAMD=n
+        ;;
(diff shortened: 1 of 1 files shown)


milkmaker:

Ayowel:
This is not stale, please review/comment

milkmaker:

Ayowel:
It would have activity if it was reviewed at some point, which it is not

DerLinkman (commented):
Hightly against it... we have ressource requiremts for reasons... yes it might be "unessecary" to have these prompts. What is happening when you run the generate_config.sh script manually not via ansible without any parameters? Does it still mention the low end system informations?

Ayowel:
> Does it still mention the low end system informations?

Yes it does in the normal use-case. As said in the description, the only change is IF the environment variables `SKIP_CLAMD` or `SKIP_SOLR` are set to non-empty values (which they should not be by default in any environment, so them being set is an active decision made by the user). In the case where they are defined - and this case only - the corresponding prompts are skipped and the provided value is used. The exact same behavior is present in the script if `MAILCOW_HOSTNAME` is set in the environment when calling the script (no prompt).
The specific names can be changed, but the behavior update is needed as mailcow **can't** be installed on "low"-resources systems with ansible at the moment (and the install simply stays pending forever, there is no error and no feedback to the user)

I can change the PR so that the changeset is ~4 lines to make it easier to review if needs be, but this would introduce bash variable usage (`${MY_VAR:-default_value}`) that does not appear to be used anywhere else in the script so I wasn't sure it would be accepted.

DerLinkman (approved):
Now i got it.

Yeah then it's fine for me. 

I completely understand the use inside Ansible then.

Ayowel:
Great. As an aside : is the ansible repository still being maintained ? The last activity appears to be from december last year, and there are a few very easy to review PR since then that have had no answer.

DerLinkman:
The ansible stuff is maintained not from us directly... it is upon the mailcow repos however.

I cannot say this 100% my last thing was: Yes but maybe the main developer has not much time recently...

Issue #5797, opened 2024-03-20
docker-compose.yml: `version` is obsolete


bersierj:
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... I understand that not following the below instructions will result in immediate closure and/or deletion of my issue.
- [X] ... I have understood that this bug report is dedicated for bugs, and not for support-related inquiries.
- [X] ... I have understood that answers are voluntary and community-driven, and not commercial support.
- [X] ... I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

### Description

```plain text
With docker version 25.05 the version 2.1 in docker-compose.yml will be obsolete
```

### Logs:

```plain text
docker-compose.yml: `version` is obsolete
```

### Steps to reproduce:

```plain text
N/A
```

### Which branch are you using?

master

### Which architecture are you using?

x86

### Operating System:

rocky linux 9

### Server/VM specifications:

12GB/4 Core

### Is Apparmor, SELinux or similar active?

yes

### Virtualization technology:

kvm

### Docker version:

25.0.5

### docker-compose version or docker compose version:

v2.25.0

### mailcow version:

2024-02

### Reverse proxy:

Nginx

### Logs of git diff:

```plain text
N/A
```

### Logs of iptables -L -vn:

```plain text
N/A
```

### Logs of ip6tables -L -vn:

```plain text
N/A
```

### Logs of iptables -L -vn -t nat:

```plain text
N/A
```

### Logs of ip6tables -L -vn -t nat:

```plain text
N/A
```

### DNS check:

```plain text
N/A
```

nabeeltahir785:
Please remove the version line from the `docker-compose.yml` file and verify that everything still works correctly on your end. Docker has made this line obsolete, which is likely the reason you're facing this issue

conjunrad:
@nabeeltahir785 from which version of docker this line obsolete?

bersierj:
If line is removed the warning go away.
the warning start with version 25.05

Cl-Chris:
Compose doesn't use version to select an exact schema to validate the Compose file. See [https://github.com/compose-spec/compose-spec/blob/master/04-version-and-name.md](url)
[...]

PR #5845, opened 2024-04-16
remove version from docker-compose.yml

iamspido:
With docker version 25.05 the version 2.1 in docker-compose.yml will be obsolete.
docker-compose.yml: `version` is obsolete

Fix for https://github.com/mailcow/mailcow-dockerized/issues/5797


milkmaker:
Thanks for contributing!

JiiPee74:
This is only warning and removing it from config file may break older docker installations, so I would leave it for now.

tomudding:
> This is only warning and removing it from config file may break older docker installations, so I would leave it for now.

The `version` attribute has been optional since the new Compose specs released at the end of 2020, see https://github.com/docker/compose/releases/tag/1.27.0-rc1 and https://github.com/compose-spec/compose-spec/commit/649d3b0d8627a50f81e268678dca93811c3934b6.

Any recent version of Mailcow requires Compose v2 to be able to work, which has the above change since its inception. So if someone is still using a version of Compose that uses/only supports the v1 spec there are likely more serious issues with that system that require attention.

DerLinkman:
I'll check what @tomudding said today. And if so (what i think because i trust him :D) it will be merged. But yeah if this is optional since Compose v2 we can safely merge this as you need compose v2 anyways to run mailcow nowadays...

DerLinkman:
Jup works without version: even in the first 2.0.0 version of compose. Alright, then it will be merged :)

DerLinkman (approved):

PR #5870, opened 2024-05-05
Fix broken dependencies for dovecot image

icedream:
- Fix python 3.11 vs 3.12 breakage because of edge repository usage
- Fix gosu armhf dependency (@MAGICCC do we even want to support this properly?)

### DIFF
--- a/data/Dockerfiles/dovecot/Dockerfile
+++ b/data/Dockerfiles/dovecot/Dockerfile
@@ -7,6 +7,9 @@ ARG GOSU_VERSION=1.16
 ENV LANG C.UTF-8
 ENV LC_ALL C.UTF-8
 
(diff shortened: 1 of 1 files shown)


milkmaker:
Thanks for contributing!

icedream:
Tested with

```
docker build --platform=armhf --rm --no-cache data/Dockerfiles/dovecot
docker build --platform=aarch64 --rm --no-cache data/Dockerfiles/dovecot
docker build --platform=amd64 --rm --no-cache data/Dockerfiles/dovecot
```

DerLinkman:
**mailcow is not armhf compatible, only ARM64 and x64.**

Won't merge this as this change is obsolelet with Alpine 3.20 as the perl package is bundled inside Alpine 3.20 natively without the need of the edge repo

PR #5871, opened 2024-05-05
Add option to send authorization header to webhook

cedric-lamalle:
This PR adds a new option `WATCHDOG_NOTIFY_WEBHOOK_AUTH_HEADER` to send an header for authorization when calling watchdog webhook.
This is useful as some webhooks don't accept Basic Authentication that can be put in the URL. I didn't force the header's name as some applications requires fancy header names as `X-API-KEY` for instance. Can be used too to "hide" Basic Authentication in the headers.
I have tested it with a custom image in our production environment and it's working fine.


milkmaker:

Issue #5645, opened 2024-01-17
Authentication very slow (7-11 sec)


mkrecek234:
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... I understand that not following the below instructions will result in immediate closure and/or deletion of my issue.
- [X] ... I have understood that this bug report is dedicated for bugs, and not for support-related inquiries.
- [X] ... I have understood that answers are voluntary and community-driven, and not commercial support.
- [X] ... I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

### Description

```plain text
Since a few months, the authentication of Mailcow UI is very slow. In the logs, I can see that this event takes about 7sec, sometimes 11sec.

Seen on Google Chrome and Safari in the same way.
```

### Logs:

```plain text
17.01.2024, 22:35:51	success	2AD8BE	luxadmin	admin	2a09:bac2:2a7e:1e1::30:4c	"verified_totp_login"
17.01.2024, 22:35:43	info	622707	unauthenticated	unauthenticated	2a09:bac2:2a7e:1e1::30:4c	"awaiting_tfa_confirmation"
```

You can see that there are 8 seconds between awaiting_tfa_confirmation and verified_totp_login.
This issue also exists if there is no two-factor login what I can see with other logins. 

### Steps to reproduce:

```plain text
I did a usual install on Ubuntu and always updated Mailcow to the latest version over time.
Steps to reproduce: Just login
```

### Which branch are you using?

master

### Operating System:

Ubuntu 22.04

### Server/VM specifications:

AMD EPYC 7282 16-Core Processor, 6 cores, 16 GB

### Is Apparmor, SELinux or similar active?

no

### Virtualization technology:

None

### Docker version:

24.0.7, build afdd53b

### docker-compose version or docker compose version:

1.29.2, build unknown

### mailcow version:

2024-01

### Reverse proxy:

Apache

### Logs of git diff:

```plain text
No
```

### Logs of iptables -L -vn:

```plain text
# Warning: iptables-legacy tables present, use iptables-legacy to see them
Chain INPUT (policy ACCEPT 0 packets, 0 bytes)
 pkts bytes target     prot opt in     out     source               destination         
[...]

PR #5880, opened 2024-05-17
Fix blocking last logins fetching

PierrePlt:
(Probably)Fixes https://github.com/mailcow/mailcow-dockerized/issues/5645

The problem was that fetching the last logins source IP's country of origin can sometime take a LONG time (up to a few minutes depending on the speed of the requests to `dfdata.bella.network`).

When the country <-> IP relationship isn't cached, this can cause the /user page to take from a few seconds to a few minutes to load for users with a long connection history.

This was caused by a variable declared in `headers.inc.php` that was never used.
Removing this variable solves the issue with loading the `/user` page but when changing tab the user would still be confronted by a long loading time since the background request to `api/v1/get/last-login` would block the loading of other pages.

The second fix is to only load the last logins when the recent logins menu is loaded.

Not sure if this is the issue the original author had, but this fixes the issue discussed on https://github.com/mailcow/mailcow-dockerized/issues/5645  on the last few days

Known issue: When `IP_SHORTCOUNTRY` is not cached, opening the `/user` page, sliding down until the recent logins section is visible, and then opening another tab needing to make an API request will confront the user with a longer than usual waiting time since the request made by the second tab will be blocked by the request to `api/v1/get/last-login`

### DIFF
--- a/data/web/inc/header.inc.php
+++ b/data/web/inc/header.inc.php
@@ -49,7 +49,6 @@ $globalVariables = [
   'app_links' => customize('get', 'app_links'),
   'is_root_uri' => (parse_url($_SERVER['REQUEST_URI'], PHP_URL_PATH) == '/'),
   'uri' => $_SERVER['REQUEST_URI'],
-  'last_login' => last_login('get', $_SESSION['mailcow_cc_username'], 7, 0)['ui']['time']
 ];
 
 foreach ($globalVariables as $globalVariableName => $globalVariableValue) {
--- a/data/web/js/site/user.js
+++ b/data/web/js/site/user.js
@@ -676,5 +676,5 @@ jQuery(function($){
   onVisible("[id^=wl_policy_mailbox_table]", () => draw_wl_policy_mailbox_table());
   onVisible("[id^=sync_job_table]", () => draw_sync_job_table());
   onVisible("[id^=app_passwd_table]", () => draw_app_passwd_table());
-  last_logins('get');
+  onVisible("[id^=recent-logins]", () => last_logins('get'));
 });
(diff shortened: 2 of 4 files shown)


mkuron:
Thanks. It does seem like eliminating the bella.network requests significantly speeds up the login.

T0biii:
https://dfdata.bella.network/:
> [09.10.2022] Added new API endpoint only for flag lookup ([/country/thomas.bella.network](https://dfdata.bella.network/country/thomas.bella.network))


maybe we should use also use the country Endpoint, or do we need the other infos?
The Json Output of `country` is a bit smaller and different than for `lookup`

country (1,2 kB):
image

lookup (1,5 kB):
image

PierrePlt:
I tried the endpoint, it doesn't seem to work with IPv6 sadly...\
Regarding querying speed, it also doesn't seem to be a lot faster than the standard endpoint, it's a shame because yes as far as I know we only need `shortcountry`.

T0biii:
maybe @Thomas2500 can say something about IPv6 and maybe the speed?

Thomas2500:
Hello @T0biii and @PierrePlt!
Thanks for the hint. The parser in the backend had an error where an IPv6 address was interpreted as an IPv4 address and thus too strongly shortened. So every IPv6 within the address space `::0` to `::fffe:ffff:ffff` was found. The error has now been fixed and IPv6 addresses should now be displayed correctly.

Regarding the speed of the API:
The old API path `https://dfdata.bella.network/lookup/` uses the old method to retrieve data, which includes many additional lookups such as DNS resolution, rDNS, ASN lookup and other background checks.
The new method with `https://dfdata.bella.network/country/` only resolves the country and skips all other checks. In tests, the query speed was reduced from approx. 700ms to 100ms. The change should be particularly noticeable for queries like this one. I have just created a pull request #5886.

DerLinkman:
Will fixed within next release. Thanks to @Thomas2500

PierrePlt:
It most likely won't be fixed in the next release.\
@Thomas2500's fix allows for an increase in performance when fetching IP location.\
This PR removes the unnecessary fetching of all the locations when loading the `/user` page.

DerLinkman (approved):
From my side it looks fine. Sorry for closing this to early that were two seperate things for the same topic...

Issue #2077, opened 2018-12-07
Wildcard aliases


levinuss:
**Is your feature request related to a problem? Please describe.**
I try to minimize the spam I get by using a separate mail address for each account I create on the internet. For example, I use archlinux@mydomain.tld for my archlinux account[s] and netflix@mydomain.tld for Netflix. I then manually create an alias for each of those mail addresses to my main account. That's not very comfortable because I have to log in to the admin panel (password + 2FA) and then manually go to the aliases to create a new one every time I register a new account on the internet.

**Describe the solution you'd like**
It would be very cool to be able to use wildcards in the aliases. That way I could keep my Catchall pointed to the 'trash' account and could also do this scheme (creating an additional alias for every account) with multiple persons on one domain. For example, I could set up *.levinus@mydomain.tld as an alias and then use archlinux.levinus@mydomain.tld for archlinux and the mail would automatically be forwarded to my main account. The next user could set up his own *.otheruser@mydomain.tld and so on.

**Additional context**
I don't know if this is actually possible, just wanted to ask. A problem could be that the * seems to be allowed in normal mail addresses, so we cannot use the asterisk as a wildcard.

Greetings,
Levin

hachre:
You can address `*@example.com` by simply using `@example.com` as the alias name. What you suggest for multiple users isn't easily possible, but you could set up subdomains with MX records (one for each user) and then use `@user1.example.com` and `@user2.example.com` and so on. So something like `archlinux@levinus.example.com` would work.

mkuron:
You can already use levinus+archlinux@mydomain.tld and levinus+netflix@mydomain.tld if you set up levinus@mydomain.tld. Anything else cannot easily be done due to the way Dovecot manages aliases.

levinuss:
I know Cachalls and also the plus signs. The problem with the plus signs is that they could possibly get removed in a spammer database because they know of that possibility. A solution with subdomains is an interesting idea, I will try that. If my idea is not possible sad but I will have to accept that :)

strarsis:
A wildcard alias like postmaster@* and abuse@* would be nice so emails to these aliases for all domains are send to the primary postmaster email address.

SnejPro:
> Anything else cannot easily be done due to the way Dovecot manages aliases.

@mkuron Why is Dovecot involved in this? Aren't aliases handled by postfix?
[...]

PR #5881, opened 2024-05-20
Feature: Catch-All on user level

SnejPro:
Use Catch All with wildcard within the local part of the e-mail address.
With this pull request you can add an user-level catch-all.

E.g.: When you add the alias `.user@example.org` all emails to `*.user@example.org` will be forwared to the goto-address.

See https://github.com/mailcow/mailcow-dockerized/issues/2077

### DIFF
--- a/data/Dockerfiles/postfix/postfix.sh
+++ b/data/Dockerfiles/postfix/postfix.sh
@@ -244,7 +244,12 @@ password = ${DBPASS}
 hosts = unix:/var/run/mysqld/mysqld.sock
 dbname = ${DBNAME}
 query = SELECT goto FROM alias
-  WHERE address='%s'
+  WHERE
+    (
+      address = SUBSTR( '%s', INSTR('%s', ".") )
+      OR
+      address = '%s'
+    )
     AND (active='1' OR active='2');
 EOF
 
--- a/data/web/inc/functions.mailbox.inc.php
+++ b/data/web/inc/functions.mailbox.inc.php
@@ -777,14 +777,28 @@ function mailbox($_action, $_type, $_data = null, $_extra = null) {
               );
               return false;
             }
+            # Get possible conflicting wildcard address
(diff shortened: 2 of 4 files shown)


mkuron:
This seems to solve a highly specific problem. Perhaps that's better addressed by modifying these files in your local Mailcow install rather than on the master version? Plus addressing (`user+something@example.org`) is much more commonly used for what you are trying to achieve and is well supported by Mailcow. I could imagine that there are about as many people who want `*.user@example.org` aliases as there are people who would only want `.user@example.org` as a non-wildcard.

Furthermore, we do some alias handling in Lua code for rspamd -- e.g. for looking up per-user thresholds and whitelists. Your Postfix-only solution does not take that into account. And there are probably some other places aware of plus-addressing and custom aliases that would need to be adapted too.

SnejPro:
> Plus addressing (`user+something@example.org`) is much more commonly used for what you are trying to achieve and is well supported by Mailcow

My personal problem is that i already have a lot of catch all adresses with this style. But i agree this is my personal problem.

However, I suspect that plus addresses do not work in part because e-mail programs or web forms do not allow the plus sign in an e-mail address.

> as there are people who would only want `.user@example.org` as a non-wildcard.

Currently mailcow does not support `.user@example.org` because php's FILTER_VALIDATE_EMAIL does not allow email addresses with leading dots.
So it should not break any existing installation.

> Furthermore, we do some alias handling in Lua code for rspamd -- e.g. for looking up per-user thresholds and whitelists. Your Postfix-only solution does not take that into account. And there are probably some other places aware of plus-addressing and custom aliases that would need to be adapted too.

I searched for` `alias` in the repository: https://github.com/search?q=repo%3Amailcow%2Fmailcow-dockerized+%60alias%60&type=code
If I make the necessary changes, would the PR then be considered for acceptance?

milkmaker:

PR #5886, opened 2024-05-28
Switch IP2Country lookup backend to shortened version

Thomas2500:
Hello!

Since the mailcow project only uses the short country function to display the country flag of the source IP, the new backend https://dfdata.bella.network/country/<IP/Domain> can be used. The lookup has been significantly accelerated and has now dropped from 700ms to 100ms on average.
Repeated requests are cached so that the search time is now only a few milliseconds.

This should improve the lookup speed mentioned in #5880.

Best regards,
Thomas


milkmaker:
Thanks for contributing!

DerLinkman (approved):
Looks good to me!

DerLinkman (approved):
Thank you!

PR #5895, opened 2024-06-05
Allow hard-coding IPV4 and IPV6 rather than it being detected

cmer:
This allows hardcoding `IPV4` and `IPV6` rather than being automatically detected when renewing SSL certificates.

This is handy in scenarios where outbound traffic comes from a different IP than what would normally be expected.

### DIFF
--- a/data/Dockerfiles/acme/acme.sh
+++ b/data/Dockerfiles/acme/acme.sh
@@ -220,14 +220,12 @@ while true; do
   if [ -z "${IPV4}" ]; then
     IPV4=$(get_ipv4)
   else
-    IPV4=${IPV4}
     log_f "Using pre-defined IPV4 address: ${IPV4}"
   fi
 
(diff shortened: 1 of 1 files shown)


milkmaker:
Thanks for contributing!

DerLinkman (changes requested):
Where does this IPs are set so the container can get them?

It needs at least a declaration inside the docker-compose.yml for the Static IPv4/IPv6 adress to be used as a variable inside the container...

cmer:
They are environment variables. If not set (this is the default), the IP is automatically resolved. Otherwise, if set, it uses the hardcoded IP in the environment variables.

dragoangel:
What is the purpose of this change? What are you trying to solve? If you using SNAT properly this should be never the case. Seems to me like a dirty attempt to fix wrong network configuration. For very broken setups there is already option to skip ip check.

milkmaker:

PR #5902, opened 2024-06-10
nextcloud: add deprecation notice once script start

DerLinkman:
This PR adds a deprecation notice for the nextcloud helper-script to mailcow as preparation for the removal of this script in December 2024.

### DIFF
--- a/helper-scripts/nextcloud.sh
+++ b/helper-scripts/nextcloud.sh
@@ -2,6 +2,39 @@
 # renovate: datasource=github-releases depName=nextcloud/server versioning=semver extractVersion=^v(?<version>.*)$
 NEXTCLOUD_VERSION=28.0.6
 
+display_warning() {
+    local message=("$@")
+    local max_length=0
+
+    for line in "${message[@]}"; do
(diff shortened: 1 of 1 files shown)


Braintelligence:
@DerLinkman Quick question: The [Nextcloud AiO docker solution](https://github.com/nextcloud/all-in-one/tree/main) looks really good. Maybe Mailcow could "support" this kind of setup by providing information in the docs on how to serve Nextcloud AiO through the `nginx-mailcow` container and how to connect the userbase?

That way mailcow is out of the responsibility concerning php-fpm etc. and would only need to provide a proper nginx config, which can be seen here: https://github.com/nextcloud/all-in-one/blob/main/reverse-proxy.md

Thanks!

DerLinkman:
Hi,

nope we will drop any support of running nextcloud within mailcows services in any way. It is not recommended to let anything run besides mailcow on one server anyways...

Braintelligence:
I see, thanks for answering. Guess I'll need to split some VMs up this year 😄.

PR #5934, opened 2024-07-03
restore: remove tty requirement from restore process to allow for automated restores

muhlba91:
i am automating the entire infrastructure creation using Pulumi as an IaC tool. for that, i tried to automate restores for an empty installation using

```
MAILCOW_BACKUP_LOCATION=/opt/backup/mailcow THREADS=4 /opt/mailcow/helper-scripts/backup_and_restore.sh restore << EOF
${backupNumber}
0
y
y
EOF
```

unfortunately, this leads to multiple `the input device is not a TTY` errors for `docker` commands because `-t` expects a "real" TTY to be present whereas the input piping doesn't provide one.

removing `-t` from the `-it` docker commands solves the problem. i tried manual and automated restores with this change and both worked without any issues.

i'd very much welcome if this change can be integrated.

### DIFF
--- a/helper-scripts/backup_and_restore.sh
+++ b/helper-scripts/backup_and_restore.sh
@@ -199,7 +199,7 @@ function restore() {
     case "$1" in
     vmail)
       docker stop $(docker ps -qf name=dovecot-mailcow)
-      docker run -it --name mailcow-backup --rm \
+      docker run -i --name mailcow-backup --rm \
         -v ${RESTORE_LOCATION}:/backup:z \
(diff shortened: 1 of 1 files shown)


milkmaker:
Thanks for contributing!

DerLinkman (commented):
Did you also test this with non automated use? Does it still work then?

muhlba91:
yes, as mentioned in the description i tried it by running it manually as well and it was working and restoring it without an issue.

DerLinkman:
Sorry overread it... yeah looks fine for me.

DerLinkman (approved):

PR #5968, opened 2024-07-23
Update Docker Compose file naming convention

iamspido:
Change docker-compose.yml to compose.yaml to align with Docker's new standard file naming convention.
See: https://docs.docker.com/compose/compose-application-model/#the-compose-file


Habetdin:
Renaming these files is not enough because there are references to them in the scripts and documentation.

Also, Compose supports `docker-compose.yml` naming for backwards compatibility of earlier versions, so there might be no real reason to change it?

iamspido:
Thank you for your feedback. You're absolutely right that simply renaming the files is not sufficient, as there are likely references to them in scripts and documentation that would need to be updated as well.

While I'm not personally familiar with the specific scripts and cross-references in the repository, I agree that it would be beneficial for someone with more in-depth knowledge of the codebase to handle these necessary adjustments.

Regarding the point about backward compatibility for docker-compose.yml, you're correct that this naming convention is still supported. However, it's worth noting that we don't have a specific end-of-support date for this backward compatibility. Given that this change has been in place since January, and considering the uncertainty around how long the old naming convention will be supported, I believe it's prudent to implement this change in the near future.

Making this update now would ensure our project aligns with Docker's new standard file naming convention and potentially prevent issues down the line. It's generally a good practice to stay current with industry standards when possible.

DerLinkman:
Until they (docker compose) don't force it i'll be highly against it, so a no from me as @Habetdin said it is highly backed into docs etc and require unecessary work...

PR #5971, opened 2024-07-28
Do not add MAILCOW_WHITE on failed DMARC

dragoangel:
We should not allow spoofing for whitelisted domains, especially when whitelists are controlled by user who usually not deep in emails authorizations mechanisms.

### DIFF
--- a/data/conf/rspamd/local.d/composites.conf
+++ b/data/conf/rspamd/local.d/composites.conf
@@ -21,6 +21,10 @@ FREEMAIL_TO_UNDISC_RCPT {
 SOGO_CONTACT_EXCLUDE {
   expression = "(-WHITELISTED_FWD_HOST | -g+:policies) & ^SOGO_CONTACT & !DMARC_POLICY_ALLOW";
 }
+# Remove MAILCOW_WHITE symbol for senders with broken policy recieved not from fwd hosts
+MAILCOW_WHITE_EXCLUDE {
+  expression = "^MAILCOW_WHITE & (-DMARC_POLICY_REJECT | -DMARC_POLICY_QUARANTINE | -R_SPF_PERMFAIL) & !WHITELISTED_FWD_HOST";
+}
 # Spoofed header from and broken policy (excluding sieve host, rspamd host, whitelisted senders, authenticated senders and forward hosts)
 SPOOFED_UNAUTH {
   expression = "!MAILCOW_AUTH & !MAILCOW_WHITE & !RSPAMD_HOST & !SIEVE_HOST & MAILCOW_DOMAIN_HEADER_FROM & !WHITELISTED_FWD_HOST & -g+:policies";


DerLinkman:
Hi,

yeah sounds reasonable.

Thank you!

krzsztf1:
Hi,
What are your thoughts on domains that use p=none (DMARC_POLICY_SOFTFAIL)?

Given that many users may whitelist senders based on the visible From: address, would it make sense to exclude MAILCOW_WHITE in such cases?

Also, a question: What is the reason for including R_SPF_PERMFAIL in this rule? Isn’t a successful DMARC verification sufficient to whitelist the sender according to the user’s preference?

For example, below is an entry from the Rspamd log showing an incoming message with a valid DKIM signature. However, the sender’s domain has two SPF records, which caused an R_SPF_PERMFAIL. While this is clearly a misconfiguration on the sender’s side, DKIM validation is still correct:

```
2025-04-24 12:55:34 #3626642(rspamd_proxy) <C756FC>; proxy; rspamd_task_write_log: id: <CAOEgGfoVaQn=Ss6G_ebXjMAA_hbJtNVXwSv8aeyBrWuchQB5dA@mail.gmail.com>, qid: <C756FCD50E>, ip: 209.85.167.177, from: <REMOVED@diplomacy.pl>, (default: F (no action): [-0.80/15.00] [DMARC_POLICY_ALLOW(-0.50){diplomacy.pl;none;},R_DKIM_ALLOW(-0.20){diplomacy.pl:s=google;},MIME_GOOD(-0.10){multipart/alternative;text/plain;},ARC_NA(0.00){},ASN(0.00){asn:15169, ipnet:209.85.128.0/17, country:US;},DKIM_TRACE(0.00){diplomacy.pl:+;},FREEMAIL_CC(0.00){gmail.com;},FROM_EQ_ENVFROM(0.00){},FROM_HAS_DN(0.00){},MIME_TRACE(0.00){0:+;1:+;2:~;},MISSING_XM_UA(0.00){},NEURAL_HAM(0.00){-0.988;},PREVIOUSLY_DELIVERED(0.00){REMOVED;},RCPT_COUNT_TWO(0.00){2;},RCVD_COUNT_ONE(0.00){1;},RCVD_IN_DNSWL_NONE(0.00){209.85.167.177:from;},RCVD_TLS_LAST(0.00){},RWL_MAILSPIKE_POSSIBLE(0.00){209.85.167.177:from;},R_SPF_PERMFAIL(0.00){empty SPF record;},TO_DN_SOME(0.00){},TO_MATCH_ENVRCPT_SOME(0.00){}]), len: 5536, time: 416.309ms, dns req: 30, digest: <a95f6d7f0bfc25c628251a71d0222816>, rcpts: <REMOVED>, mime_rcpts: <REMOVED>
```

SPF lookup result for the domain:
```
$ dig diplomacy.pl TXT +short
"v=spf1 a mx ip4:138.201.252.19 ~all"
"v=spf1" "include:_spf.google.com" "~all"
```

dragoangel:
If a user has 2 SPF records, he breaks his domain, that's it, this is not valid settings and such mail should not be whitelisted. It should be one record listening all ips. About DMARC p=none, again - that's sender policy, if he wants so - okay, but if his spf will fail too - we not passing him, that's the point :) when you whitelist somebody by domain you mean you trust their domain, no matter what policy they have - trust should work, if you see that policy is none - you can not trust domain and as result not add them to whitelist but instead whitelist their ip or whatever

krzsztf1:
Surely, when p=none is used, I cannot trust the domain and should instead whitelist it by IP, for example. However, a regular user who can customize their whitelist by adding any domain might not be aware of this.

They might add the domain of a trading partner (e.g., geis.pl) because, for some reason, legitimate emails from that domain are being marked as spam or rejected, and they want to prevent that.

As a result, Rspamd adds the MAILCOW_WHITE symbol with -1998 points, which works — emails are no longer marked as spam or rejected. But at some point, an attacker sends them an email with a forged From header using the same whitelisted domain, from a server they control (SPF passes). Since the domain in the From header is on the user's whitelist, the message receives a large negative score (e.g., -999 points) and lands in the user's inbox.

I'm wondering — isn't this a legitimate risk for users?

```
[Metric: default]
Action: no action
Spam: false
Score: -996.11 / 15.00
Symbol: ARC_NA (0.00)
Symbol: ARC_SIGNED (0.00)[cow.int.removed.pl:s=dkim:i=1]
Symbol: ASN (0.00)[asn:47544, ipnet:XX.XX.XX.XX/24, country:PL]
Symbol: BCC (0.00)
Symbol: DMARC_POLICY_SOFTFAIL (0.10)[geis.pl : SPF not aligned (relaxed), No valid DKIM, none]
Symbol: FORGED_SENDER (0.00)[ceo@geis.pl, notscamatall@removed2.pl]
Symbol: FORGED_W_BAD_POLICY (3.00)
Symbol: FROM_HAS_DN (0.00)
Symbol: FROM_NEQ_ENVFROM (0.00)[ceo@geis.pl, notscamatall@removed2.pl]
Symbol: MAILCOW_WHITE (-999.00)
Symbol: MIME_GOOD (-0.10)[text/plain]
Symbol: MIME_TRACE (0.00)[0:+]
Symbol: MX_GOOD (-0.01)[]
Symbol: PREVIOUSLY_DELIVERED (0.00)[test@cow.int.removed.pl]
Symbol: RCPT_COUNT_ONE (0.00)[1]
Symbol: RCPT_MAILCOW_DOMAIN (0.00)[cow.int.removed.pl]
Symbol: RCVD_COUNT_THREE (0.00)[4]
Symbol: RCVD_NO_TLS_LAST (0.10)
Symbol: RECEIVED_HELO_LOCALHOST (0.00)
Symbol: R_DKIM_NA (0.00)
Symbol: R_SPF_ALLOW (-0.20)[+ip4:XX.XX.XX.XX/32]
Symbol: SINGLE_SHORT_PART (0.00)
Symbol: TO_DN_NONE (0.00)
Symbol: TO_MATCH_ENVRCPT_ALL (0.00)
Message-ID: 20250509170816.3520757@proxy2.removed2.pl
```

dragoangel:
This is the point of p=none, if you as administrator don't want such thing
to ever happen you have couple of options:
1. Administrative: do not allow users to manage WL and/or rspamd actions on
their own by removing corresponding ACL from them.
2. Organizational: instead of revoking rights educate your coworkers to not
apply any ACLs without understanding the reasons and prefer to contact
SysAdmin in first place for a help
3. Force actions: configure rspamd to forcibly quarantine or reject emails
with non desired symbols
4. Modify current composite to fit your needs - mailcow is platform with
sane defaults, but it's doesn't mean that all defaults we have can fit
everyone's needs and expectations.

I can assume most people needs simpler whitelist mechanism that works, but
some can prefer stricter rules, it's impossible to create system that would
apply both at once here and from my personal view current defaults are okay
for most. I personally don't allowed users to apply any WLs and all, they
have other risks: we not split emails per RCPT so as result if multi-rcpt
email is received and only 1 person applied WL - it still will be delivered
to everyone.

PR #5987, opened 2024-08-06
Bug Fix: Check `mailcow.conf` exists before work with it

h3ssan:
## Modified Files

- `update.sh` script

## Description

In `update.sh` file, there's this part:

```bash
chmod 600 mailcow.conf
source mailcow.conf

detect_docker_compose_command

[[ ! -f mailcow.conf ]] && { echo "mailcow.conf is missing! Is mailcow installed?"; exit 1;}
```

Which as it looks, first chmod, then source the `mailcow.conf`, and then it checks if `mailcow.conf` exists or not.

The modified version is:

```bash
[[ ! -f mailcow.conf ]] && { echo "mailcow.conf is missing! Is mailcow installed?"; exit 1;}

chmod 600 mailcow.conf
source mailcow.conf

detect_docker_compose_command
```

## Test

- I have tested it manually

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -404,12 +404,13 @@ while (($#)); do
   shift
 done
 
+[[ ! -f mailcow.conf ]] && { echo "mailcow.conf is missing! Is mailcow installed?"; exit 1;}
+
 chmod 600 mailcow.conf
 source mailcow.conf
 
 detect_docker_compose_command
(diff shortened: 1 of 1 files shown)


DerLinkman (approved):
Looks fine to me!

PR #6000, opened 2024-08-11
[Rspamd] Fix bayes config

dragoangel:
Add hint about classifier name, and add missing learn_condition

### DIFF
--- a/data/conf/rspamd/local.d/statistic.conf
+++ b/data/conf/rspamd/local.d/statistic.conf
@@ -1,11 +1,13 @@
 classifier "bayes" {
+    # name = "custom";  # 'name' parameter must be set if multiple classifiers are defined
+    learn_condition = 'return require("lua_bayes_learn").can_learn';
+    new_schema = true;
     tokenizer {
       name = "osb";
     }
     backend = "redis";
     min_tokens = 11;
     min_learns = 5;
-    new_schema = true;
     expire = 2592000;
     statfile {
       symbol = "BAYES_HAM";
--- a/data/conf/rspamd/local.d/statistic.conf
+++ b/data/conf/rspamd/local.d/statistic.conf
@@ -8,7 +8,7 @@ classifier "bayes" {
     backend = "redis";
     min_tokens = 11;
     min_learns = 5;
-    expire = 2592000;
+    expire = 7776000;
     statfile {
       symbol = "BAYES_HAM";
       spam = false;


dragoangel:
Missing `learn_condition` in `statistic.conf` will lead to overlearning: [Statistics required parameters](https://rspamd.com/doc/configuration/statistic.html#:~:text=The%20default%20function%20must%20be%20set%20if%20you%20have%20not%20written%20your%20own.%20Omitting%20learn_condition%20from%20statistic.conf%20will%20lead%20to%20losing%20protection%20from%20overlearning.)

@DerLinkman bayes should be dropped manually or will need to wait for it to get expire. I think first one is more preferred. We can link it to [existing docs](https://docs.mailcow.email/manual-guides/Rspamd/u-e-rspamd-work-with-spamdata/#reset-bayes-data)

Having 30 days for expire I think can be increased to 90 days, as with Rspamd 3.9 new bayes will eat in 3.5 times lower space, we can safely extend expire in 3 times due to it. This will also align bayes expire with local fuzzy, which is also 90 days.

DerLinkman (approved):
Looks fine

DerLinkman:
> Missing `learn_condition` in `statistic.conf` will lead to overlearning: [Statistics required parameters](https://rspamd.com/doc/configuration/statistic.html#:~:text=The%20default%20function%20must%20be%20set%20if%20you%20have%20not%20written%20your%20own.%20Omitting%20learn_condition%20from%20statistic.conf%20will%20lead%20to%20losing%20protection%20from%20overlearning.)

Thanks for pointing this out.

> @DerLinkman bayes should be dropped manually or will need to wait for it to get expire. I think first one is more preferred. We can link it to [existing docs](https://docs.mailcow.email/manual-guides/Rspamd/u-e-rspamd-work-with-spamdata/#reset-bayes-data)

Yeah we should keep the dropping schemes as they are now, it's fine for me and i don't see why we would have to change it. If i understood it correctly the bayes are expiring after 90 days (after your pr merge now) or if a user removes them automatically?

> Having 30 days for expire I think can be increased to 90 days, as with Rspamd 3.9 new bayes will eat in 3.5 times lower space, we can safely extend expire in 3 times due to it. This will also align bayes expire with local fuzzy, which is also 90 days.

Sounds reasonable.

dragoangel:
About part 2 with dropping bayes I mean that we need to ask (or highly recommend to) users in update notes to drop their bayes once after update.

Reason is that over learning will not be removed by itself. Bayes expires module - expires only tokens that are non "core" one, they will be put to TTL of 90d, and insufficient stat token will put TTL of 10d (default), core tokens will be "persisted", and will not expire at all. Over learned tokens usually will be over learned one.

DerLinkman:
> About part 2 with dropping bayes I mean that we need ask users to drop their bayes once after update.

Is that necessary? Or is it enough if we write this into our update posts only?

dragoangel:
> Is that necessary? Or is it enough if we write this into our update posts only?

It's highly recommended. If people don't follow it, everything will work too, they just continue to leave with what they already overlearned. Overlearned keys is hard (almost impossible) to relearn.

So update post is enough, but with **bolt** font :)

PR #5945, opened 2024-07-09
Don't expose SMTP/IMAP if announced "not provided" via SRV

SailReal:
Fixes #5944

Issue #5944, opened 2024-07-09
Do not publish IMAP/SMTP in autoconfig.php if disabled via SRV record


SailReal:
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... I understand that not following the below instructions will result in immediate closure and/or deletion of my issue.
- [X] ... I have understood that this bug report is dedicated for bugs, and not for support-related inquiries.
- [X] ... I have understood that answers are voluntary and community-driven, and not commercial support.
- [X] ... I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

### Description

In https://github.com/mailcow/mailcow-dockerized/blob/master/data/web/autodiscover.php#L164-L185 we only publish TLS ports. In https://github.com/mailcow/mailcow-dockerized/blob/master/data/web/autoconfig.php#L35-L48 and https://github.com/mailcow/mailcow-dockerized/blob/master/data/web/autoconfig.php#L73-L86 we publish TLS and plain ports.

In https://github.com/mailcow/mailcow-dockerized/blob/master/data/web/autoconfig.php#L62 and https://github.com/mailcow/mailcow-dockerized/blob/master/data/web/autoconfig.php#L51 we already check if POP3/s is enabled via SRV record so checking against SRV records wouldn't even be something new.

To address https://www.usenix.org/system/files/sec21-poddebniak.pdf I "disabled" plain ports via 

```
_imaps._tcp           SRV   0 1 993 mail.example.com
_smtps._tcp           SRV   0 1 465 mail.example.com
_submissions._tcp     SRV   0 1 465 mail.example.com

_submission._tcp      SRV  0 0 0   .
_smtp._tcp            SRV  0 0 0   .
_imap._tcp            SRV  0 0 0   .
_pop3._tcp            SRV  0 0 0   .
_pop3s._tcp           SRV  0 0 0   .
```

but due to the code in `autoconfig`, they are still published for this endpoint only.

In the docs in https://docs.mailcow.email/getstarted/prerequisite-dns/#the-advanced-dns-configuration is stated

> SRV records specify the server(s) for a specific protocol on your domain. If you want to explicitly announce a service as not provided, give "." as the target address (instead of "mail.example.org."). Please refer to RFC 2782.

[...]

PR #6014, opened 2024-08-16
Expose the same protocols in autoconfig and autodiscover

SailReal:
## Contribution Guidelines

### Short Description

This PR streamlines the exposure of protocols in autoconfig and autodiscover by dropping dynamic config based on SRV values for autoconfig, for more information see https://github.com/mailcow/mailcow-dockerized/issues/5944 and https://github.com/mailcow/mailcow-dockerized/pull/5945 .

## Did you run tests?

No

### DIFF
--- a/data/web/autoconfig.php
+++ b/data/web/autoconfig.php
@@ -39,17 +39,6 @@ header('Content-Type: application/xml');
          <username>%EMAILADDRESS%</username>
          <authentication>password-cleartext</authentication>
       </incomingServer>
-      <incomingServer type="imap">
-         <hostname><?=$autodiscover_config['imap']['server']; ?></hostname>
-         <port><?=$autodiscover_config['imap']['tlsport']; ?></port>
-         <socketType>STARTTLS</socketType>
-         <username>%EMAILADDRESS%</username>
-         <authentication>password-cleartext</authentication>
-      </incomingServer>
-
-<?php
-$records = dns_get_record('_pop3s._tcp.' . $domain, DNS_SRV); // check if POP3 is announced as "not provided" via SRV record
-if (count($records) == 0 || $records[0]['target'] != '') { ?>
       <incomingServer type="pop3">
          <hostname><?=$autodiscover_config['pop3']['server']; ?></hostname>
          <port><?=$autodiscover_config['pop3']['port']; ?></port>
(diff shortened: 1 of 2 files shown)


data/web/autodiscover.php
```diff
@@ -171,6 +171,16 @@
...
+      <Protocol>
+        <Type>POP3</Type>
+        <Server><?=$autodiscover_config['pop3']['server'];?></Server>
+        <Port><?=$autodiscover_config['pop3']['port'];?></Port>
+        <DomainRequired>off</DomainRequired>
+        <LoginName><?=$email;?></LoginName>
+        <SPA>off</SPA>
+        <SSL>on</SSL>
+        <AuthRequired>on</AuthRequired>
+      </Protocol>
```
mkuron, line 183: This is inside the `$autodiscover_config['autodiscoverType'] == 'imap'` section, which I don't think makes sense. Does Outlook even support having both IMAP and POP3 in the same discovery? Which one does it pick?

data/web/autoconfig.php
```diff
@@ -39,51 +39,20 @@
...
-      <incomingServer type="imap">
-         <hostname><?=$autodiscover_config['imap']['server']; ?></hostname>
-         <port><?=$autodiscover_config['imap']['tlsport']; ?></port>
-         <socketType>STARTTLS</socketType>
-         <username>%EMAILADDRESS%</username>
-         <authentication>password-cleartext</authentication>
-      </incomingServer>
-
-<?php
-$records = dns_get_record('_pop3s._tcp.' . $domain, DNS_SRV); // check if POP3 is announced as "not provided" via SRV record
```
mkuron, line 51: I think being able to hide POP3 was useful. In fact, I was the person who added this code seven years ago. However, I do understand that you want to deal with situations where the DNS might unreliable. So instead of unconditionally presenting both POP3S and IMAPS, perhaps we could either have a configuration switch somewhere, or, perhaps more conveniently, simply query our database and hide the protocol if it is disabled for all mailboxes (and all their app passwords) of the domain.

mkuron (changes requested):

SailReal:
Since the 4th of July on https://community.mailcow.email/d/3803-disable-plain-ports-starttls-and-force-tls, https://github.com/mailcow/mailcow-dockerized/issues/5944 and https://github.com/mailcow/mailcow-dockerized/pull/5945 I have been trying to explain and fix the problem of autoconfig and autodiscover exposing different protocols, as well as the option to disable potentially less secure STARTTLS.

For the fourth time, we are starting the discussion all over again, after trying to find out exactly how you want it to be implemented (long discussion in #5945), after reverting the merged previous one. I'm sorry, I don't see any way that we can address it in a way that you all agree on, other than what we all want to have as an option in the backend, but I see that in the distant future.

mkuron:
My intention was not to decline this pull request, quite the contrary. I was merely pointing out that the change to autodiscovery.php had a logic error and the functionality removed from autoconfig.php could be added in a different way without much of a hassle.

PR #6019, opened 2024-08-16
Refactor/Change Dockerfiles cmd from shell to exec form

h3ssan:
## Contribution Guidelines

## What does this PR include?

In this PR, I modified the `Dockerfile`s to make sure all Dockerfiles uses the same `exec form` instead of old `shell form` in CMD instructions. Here's the modified components:

- dockerapi
- postfix
- sogo
- unbound
- watchdog

### Short Description

The update is to make the whole project use [`exec form of CMD (click for ref)`](https://docs.docker.com/reference/dockerfile/#shell-and-exec-form) instead of the old shell form, to make the project concise. Because some Dockerfile instructions still use the old shell form and other use the newer exec form.

###  Affected Containers

- dockerapi
- postfix
- sogo
- unbound
- watchdog

## Did you run tests?

Yes.

### What did you tested?

I tested the affected containers, mentioned above.

I build the containers that I modified and make sure they all works and in healthy state.

### DIFF
--- a/data/Dockerfiles/dockerapi/Dockerfile
+++ b/data/Dockerfiles/dockerapi/Dockerfile
@@ -24,4 +24,4 @@ COPY main.py /app/main.py
 COPY modules/ /app/modules/
 
 ENTRYPOINT ["/bin/sh", "/app/docker-entrypoint.sh"]
-CMD exec python main.py
+CMD ["python", "main.py"]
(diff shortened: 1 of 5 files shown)


DerLinkman (approved):
Looks fine to me!

But we won't rebuild the containers extra for this, as it is only a "cosmetic" thing to change and is not critical for the operation.

Thank you for your contribution!

h3ssan:
> Looks fine to me!
> 
> But we won't rebuild the containers extra for this, as it is only a "cosmetic" thing to change and is not critical for the operation.
> 
> Thank you for your contribution!

No problem, it's not a bug, just a refector.

PR #6032, opened 2024-08-20
[Web] Do not show Apps dropdown, when there is only single item

feldsam:
## Contribution Guidelines

### Short Description

Do not show Apps dropdown in top nav, if there is only single item

###  Affected Containers

- web

### What did you tested?

Tested, working good

### DIFF
--- a/data/web/templates/base.twig
+++ b/data/web/templates/base.twig
@@ -89,24 +89,33 @@
           </ul>
         </li>
         {% endif %}
-        {% if mailcow_apps or app_links %}
-        <li class="nav-item dropdown">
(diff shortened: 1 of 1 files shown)


FreddleSpl0it:
This does not work as expected. If I only have Webmail as App Link it works, but if I set this in `vars.local.inc.php`
```
$MAILCOW_APPS = array(

);
```

And add only one custom App Link via UI, it still shows a Dropdown.

feldsam:
Yes, this case is not programed, but for installations with default settings, where is only SOGo, it works.

DerLinkman:
I would like to make it clean when implementing this so it should be matching all types of links (ui or php file)

feldsam:
yes, agreed

feldsam:
Need to look into this again

patschi:
Is this still something being worked at?

PR #6155, opened 2024-11-12
Fix #2752 - Allow domain recipients for address rewrite

### LINKED ISSUE #2752: Adress Rewrite for whole Domains
Hi,

I tried to "Address Rewrite" a whole Domain. I knew Postfix is able to do that because a Friend of mine is also doing that. The Domain is fully Relayed to another Server so all the Mails need to be rewritten to another Domain.

Mailcow supports only rewrite to a single Mail Address in the UI, but Postfix supports also a rule like:
@example.com -> @new-example.com

I know it works as i then added this Rule to the SQL DB and all Mail gets Domain Rewritten.

PseudoResonance:
## Contribution Guidelines

### Short Description

Fix for #2752 to allow address rewriting to entire domains, instead of being limited to a single email.
I simply copied the old address validation code and replaced the new address validation so that they both share the same code.

###  Affected Containers

 - nginx

### What did you tested?

I inputted a number of different valid/invalid domain and email combinations to ensure the validation works as intended.

Works as described in #2752 and either exact emails or whole domains can be added as a new address.


milkmaker:

PseudoResonance:
I rather dislike these stale bots...

FreddleSpl0it (changes requested):
Please also apply this logic to `$new_dest`
```
      if (substr($old_dest, 0, 1) == '@') {
        $old_dest = substr($old_dest, 1);
      }
```

After adding a domain as `$new_dest`, an `@` is prepended. Editing this entry will only work if the `@` is removed.

PseudoResonance:
> After adding a domain as `$new_dest`, an `@` is prepended. Editing this entry will only work if the `@` is removed.

I think I implemented all of `old_dest`'s logic to `new_dest` as well. It now removes the prefixing `@` when editing it and properly handles it when included.

Additionally, I updated what localizations I could to indicate that an email or domain is valid. I simply used the old dest translation, and updated the new dest one to match. I checked them with Google translate and I believe I got all the grammar correct in the languages I could check, however I'm only fully confident about English, Japanese and Spanish. The rest I do not speak and just had to use the internet.

A few languages I couldn't confidently update though, as the old_dest translation is inaccurately not mentioning domains as being valid, or were missing the translation. For example, Ukrainian, Simplified Chinese, Taiwanese.

I can remove them though if you want them in another commit or something.

FreddleSpl0it (approved):

PR #6225, opened 2024-12-20
Reduce sa rules download retry limit to 5

alyx161:
## Contribution Guidelines

### Short Description

Reduces the retry limit for the sa rules download to a more reasonable 5 retries to prevent running in a timeout condition.
Fixes https://github.com/mailcow/mailcow-dockerized/issues/6018

###  Affected Containers

- dovecot

## Did you run tests?

Yes

### What did you tested?

Startup and general functionality

### DIFF
--- a/data/Dockerfiles/dovecot/sa-rules.sh
+++ b/data/Dockerfiles/dovecot/sa-rules.sh
@@ -11,7 +11,7 @@ else
 fi
 
 # Deploy
-if curl --connect-timeout 15 --retry 10 --max-time 30 https://www.spamassassin.heinlein-support.de/$(dig txt 1.4.3.spamassassin.heinlein-support.de +short | tr -d '"' | tr -dc '0-9').tar.gz --output /tmp/sa-rules-heinlein.tar.gz; then
+if curl --connect-timeout 15 --retry 5 --max-time 30 https://www.spamassassin.heinlein-support.de/$(dig txt 1.4.3.spamassassin.heinlein-support.de +short | tr -d '"' | tr -dc '0-9').tar.gz --output /tmp/sa-rules-heinlein.tar.gz; then
   if gzip -t /tmp/sa-rules-heinlein.tar.gz; then
     tar xfvz /tmp/sa-rules-heinlein.tar.gz -C /tmp/sa-rules-heinlein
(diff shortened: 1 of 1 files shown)


DerLinkman:
Not merging this as this problem is on your side

alyx161:
@DerLinkman If this is a problem on my side, then why are others complaining about it in the issue?

Or in other words, the original pr did not fixed the issue.

DerLinkman:
Do you know what this pr do?

It is lowering the tries the SA Rules will try to download.

On your and on a the other persons machines they cannot be downloaded at all, thats why dovecot is taking so long to start. Why? Inspect your machine. On a clean setup it works without a charm.

We can argue about, that this is not a clean solution and yes that is right. An idea i have back in mind is to remove that section from the startup completely and putting this in a background job so dovecot can start always. But again, this will not solve your core issue which is some kind of network related, which we won't analyse for you though.

alyx161:
@DerLinkman 

I'm a bit stunned about your reply.
This has nothing to do with my network or setup.

You yourself added https://github.com/mailcow/mailcow-dockerized/pull/6025 as a response to a scenario in which the SA rules can't be downloaded if the remote server is down. Your pr includes 10 retires with a timeout of 30 seconds each, so it takes at least 5 minutes until the container continues to startup in this scenario.

The issue is, since the container is waiting 5 minutes for a obviously dead server to reply, the container doesn't reach a healthy state in time and restarts before it even is able to finish the 10 retires.

Or in other words, your original pr doesn't have the desired affect, since the container is still crashing and restarting if the remote server is down.

You are right, that a background taks approach probably would be the best idea.
But in this pr I stick with your original idea of a timeout.

The easiest fix for your initial pr seemed to just reduce the retries to a more reasonable number.
And this is what this pr is about.

I know it's not a important contribution.
I literally just changed one number.
It's also a very hard edge case since it only applies to a scenario where the remote server is down/unavailable.
But it still improves the experience in such a scenario.

DerLinkman:
Yeah, i understand. Sorry for harsh reply i'm pretty busy currently.

Let me rethink this pr: i think as a quick workaround until we/i have found a solution which is downloading it in the background ready will be the easiest way.

DerLinkman (approved):
Looks fine, Dovecot needs to be rebuild for this to change effect (notice to myself)

Issue #6228, opened 2024-12-26
Synchronization failure of IMAP Port has been changed


Boldie:
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... I understand that not following the below instructions will result in immediate closure and/or deletion of my issue.
- [X] ... I have understood that this bug report is dedicated for bugs, and not for support-related inquiries.
- [X] ... I have understood that answers are voluntary and community-driven, and not commercial support.
- [X] ... I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

### Description

```plain text
After doing a fresh install, I want to transfer my old mail data using synchronize. For this to succeed I have changed the ports for the new mailcow installation to be prefixed with a 9 for eg. IMAP is now on 9143.
After starting the sync I found out the port change is not respected by synchronization mechanism which will lead to an aborted sync. The port for the new server seems not to be recognized while calling the external tool.

Inspecting the command line form the logs shows the --port2 option is missing.
```

### Logs:

```plain text
Host1: Socket successfully converted to SSL
Host1: success login on [imap.xxxx.de] with user [usernameA] auth [LOGIN] or [LOGIN]
Host2: connecting and login on host2 [localhost] port [143] with user [xxx@xxxx.de*fffff@mailcow.local]
Host2 failure: can not open imap connection on host2 [localhost] with user [xxx@xxxx.de*fffff@mailcow.local]: Unable to connect to localhost:  Connection refused
```

### Steps to reproduce:

```plain text
1. Install Mailcow
2. Change IMAP Port to something different then 143
3. Add a Mailbox and Domain
4. Configure a synchronization from a different server
5. -> The sync breaks because he can not connect to Port 143 in the new mailcow installation.
```

### Which branch are you using?

master

### Which architecture are you using?

x86

### Operating System:

Uneccesary

### Server/VM specifications:

Uneccesary

### Is Apparmor, SELinux or similar active?

Uneccesary

### Virtualization technology:

Uneccesary

### Docker version:

Uneccesary

### docker-compose version or docker compose version:

Uneccesary

### mailcow version:

Uneccesary

### Reverse proxy:

Uneccesary

### Logs of git diff:

```plain text
Uneccesary
```

### Logs of iptables -L -vn:

[...]

PR #6229, opened 2024-12-26
Fix imapsync to support other ports for IMAP Server than 143

Boldie:
## Contribution Guidelines

* [x ] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree them

### Short Description

In case of configuring a different port for the builtin IMAP server the synchronization fails. The reason is there was a missing definition for the port of the destination machine (--port2 parameter to imapsync). This will now be added if the IMAP_PORT environment is set.

###  Affected Containers

- dovecot

## Did you run tests?

Yes, for my setup.

### What did you tested?

I tested if sync now works for my config with a dovecot on Port 9143. The port is now passed correctly.

The sync does no longer fail.


Boldie:
Refs  #6228

PR #6265, opened 2025-01-26
Added availability to redirect HTTP to HTTPS and  a variable "REDIREC…

Scuruba:
…T_HTTP" to configure this option

## Contribution Guidelines

### Short Description

A updated NGINX config to redirect HTTP requests to HTTPS. This redirect should be activated/deactivated over the mailcow.conf, via the variable "REDIRECT_HTTP=y/n"

###  Affected Containers

- NGINX

### What did you tested?

I tested and also actually run the NGINX config itself, like in the PR.
But I couldn't get the part with the variable configuration to work! I don't have enough knowledge for this, so I tried to use the variable “DISABLE_IPv6” as a guide.

NGINX config itself was successfully tested, the variable part doesn't work.


DerLinkman:
This would basically obsolete the Docs page here right?

https://docs.mailcow.email/manual-guides/u_e-80_to_443

@FreddleSpl0it what do you think?

FreddleSpl0it:
https://docs.mailcow.email/manual-guides/u_e-80_to_443 should still work, so I don’t understand why we should change it. I don’t see any advantage.

Clete2:
> https://docs.mailcow.email/manual-guides/u_e-80_to_443 should still work, so I don’t understand why we should change it. I don’t see any advantage.

It doesn’t work for me. I didn’t look into it yet but can confirm that curl -v to the http endpoint returns a 200, not a 301. The docs at minimum need to be updated for the new NGINX config

Scuruba:
> https://docs.mailcow.email/manual-guides/u_e-80_to_443 should still work, so I don’t understand why we should change it. I don’t see any advantage.

I think redirect HTTP should be the standard, and not be an option.
But the config doesn't work because you call the port 80 two times, at first with the standard config, and then the second time in the redirect.conf. NGINX warn me every time, because I declared the sides two times. (with nginx -t).  NGINX uses the first entry of the port 80 call, so the redirect will not work, you land on the http site.

FreddleSpl0it:
I'm not sure if we should make it a standard, but I reviewed the issue with redirect.conf not working and can now confirm it.
After that, I agree that we should move it to the Jinja2 template and trigger it via an ENV variable.

Please add the new ENV variable to `generate.sh` and `update.sh`.

`generate.sh`
https://github.com/mailcow/mailcow-dockerized/blob/f92ddd86c543d115aab78bf5b3fe7e6414136a52/generate_config.sh#L255-L268

`update.sh`
https://github.com/mailcow/mailcow-dockerized/blob/f92ddd86c543d115aab78bf5b3fe7e6414136a52/update.sh#L896-L942

Scuruba:
Okay thanks @FreddleSpl0it, I added the variable to both scripts.
I made the default as no, for the ones who don't want it as a default.

But the problem that it does not work for me via the variable remains, does anyone have an idea what the problem could be?

FreddleSpl0it:
Looks good at first sight. Did you rebuild the Nginx Docker image?

Scuruba:
I removed the old `data/Dockerfiles/nginx/bootstrap.py` , `docker-compose.yml` and `data/conf/nginx/templates/nginx.conf.j2` files, copied the new ones and added the variable REDIRECT_HTTP=y to `mailcow.conf`

Then recreated the containers with `docker compose up --build --force-recreate --no-deps -d`
Nothing changed.

Scuruba:
I played a bit around, and the expression `{% if REDIRECT_HTTP is defined %}` is false!
But the expression `{% if DISABLE_IPv6 is defined %}` is true.

In portainer, in the nginx-container I see the ENVs are both listed and also follows the mailcow.conf entry (y/n). Looks like that not docker is the problem, but maybe Jinja2.
The `bootstrap.py` file looks correct for me.

And also 
```
docker compose exec nginx-mailcow python3
>>> import os
>>> print(os.environ['REDIRECT_HTTP'])
```
print y or n as an answer

FreddleSpl0it:
Thanks for the PR.
I had to make some other changes to Nginx, so I added it in along the way.
https://github.com/mailcow/mailcow-dockerized/commit/e645f931dc04c8b8754927d90275a2e77a03931d

PR #6270, opened 2025-01-28
Update generate_config.sh version checking for wider compatibility

digitalhen:
fix: replace `grep -oP` with `grep -oE` for broader compatibility

The `-P` option (Perl-compatible regex) is not supported in all versions of `grep`, particularly the default BSD `grep` on macOS. This change replaces `-P` with `-E` (extended regex), which is more widely available and ensures compatibility across different environments.

Tested on macOS and Linux.

## Contribution Guidelines

### Short Description

This PR replaces `grep -oP` with `grep -oE` to improve compatibility across different environments. The `-P` option (Perl-compatible regex) is not supported in all versions of `grep`, particularly the default BSD `grep` on macOS. The script fails in its current format on MacOS. Using `-E` (extended regex) ensures broader support while maintaining the intended functionality.

### Affected Containers

None (this is a script change, not a change to Docker containers).

### What did you test?

Tested the updated command on both macOS and Linux to ensure it extracts the major version correctly.

### What were the final results? (Expected, actual)

**Expected:** The command should return only the major version number from `docker --version`.  
**Actual:** The command correctly returns the major version number on both macOS and Linux.

### DIFF
--- a/generate_config.sh
+++ b/generate_config.sh
@@ -26,7 +26,7 @@ for bin in openssl curl docker git awk sha1sum grep cut; do
 done
 
 # Check Docker Version (need at least 24.X)
-docker_version=$(docker -v | grep -oE '\d+\.\d+\.\d+' | head -n 1 | cut -d '.' -f 1)
+docker_version=$(docker version --format '{{.Server.Version}}' | cut -d '.' -f 1)
 
 if [[ $docker_version -lt 24 ]]; then
   echo -e "\e[31mCannot find Docker with a Version higher or equals 24.0.0\e[0m"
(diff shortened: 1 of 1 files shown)


DerLinkman:
You know that we do not support macOS right?

digitalhen:
This was the only line preventing it working on my Mac. Would be great if it could be included -- would diversify where the container can run -- even if unsupported.

DerLinkman:
On which Linux did you tested this?

digitalhen:
I tested on CentOS, but I see it doesn't work on Ubuntu. A more universal approach would be this. This works on Ubuntu, MacOS, CentOS.

`docker version --format '{{.Server.Version}}' | cut -d '.' -f 1`

DerLinkman:
Yes your last approach works better it seems.

Can you update your pr?

DerLinkman (approved):
Looks good

PR #6288, opened 2025-02-01
Example configuration for AIO config: mailcow +caddy +borg +mailman

midzelis:
## Contribution Guidelines

## What does this PR include?

This contains an all-in-one example of using mailcow with Caddy as reverse proxy, Borg, for backups, and Mailman, for a list manager. 

* docker-compose.caddy-mailman-borg-example.override.yml

This override makes the following configuration changes: 
1) Caddy, configured with auto container-reload script. This uses a custom caddy Dockerfile (included), to configure caddy with necessary modules, included. example conf files included. 
2) Borg - backs up all volumes, databases, and the mailcow-dockerized root (conf files). Example conf files included. 
3) Mailman - example conf files included. 

If you choose to use this override, a user should also add these configuration directives in their `mailcow.conf` file:

```
HTTP_PORT=180 # listen on alt port
HTTP_BIND=127.0.0.1 # only listen on private loopback

HTTPS_PORT=1443 # listen on alt port
HTTPS_BIND=127.0.0.1 # only listen on private loopback

CADDY_BIND=0.0.0.0
CADDY_HTTP_PORT=80
CADDY_HTTPS_PORT=443

BORG_PASSPHRASE=some-long-secure-phrase

HYPERKITTY_API_KEY=some-long-secure-key
DJANGO_KEY=some-long-secure-key
MAILMAN_DB_NAME=mailmandb
MAILMAN_DB_USER=mailmandbuser
MAILMAN_DB_PASS=some-long-secure-key
MAILMAN_DOMAIN=lists.example.org
MAILMAN_ADMIN_USER=mailman
MAILMAN_ADMIN_EMAIL=mailman@example.org

```
### Short Description

A lot of this was mostly derived from existing guides. However, I found some a little out of date, and I didn't like some of the advise they were suggesting. 

i.e. the Caddy example used a script and a cronjob to copy certs. Instead, I used caddy events to remove need for polling, the certificate is shared using shared volume mounts, and it uses a modified script from the acme folder to use the dockerapi endpoints to restart the dovecot/postfix containers. 

The borg backup guide was ok, but I didn't like that it mounted the mysql sock. Just use the hostname instead. This was also modified to include the postgres database used by mailman. Additionally, it also bind-mounts the root mailcow-dockerized folder, so all config data is also backed-up.

The mailman guide was also ok, but it suggested to apache as a reverse proxy. I prefer Caddy, since it handles ACME/SSL automatically, and also does reverse-proxy very well. Additionally, this guide suggested to install apache not as a container, but as app installed directly on the host - would prefer to use a container solution instead. It also suggested to install mailman, using their own dockerfiles, and creating an external network to allow the two docker-compose stacks to communicate with each other. Instead, I added all of the mailman docker compose configuration files directly to the override - you do not need to download/clone mailman from git either. The example docker-compose includes everything. 

Additionally, I needed to custom build mailman's dockerfiles for aarch64 support, since I'm using Heztner cloud's arm servers. 

TODOs: 
* Caddy could replace the need for nginx completely, but that was a much bigger change, and I didn't really see the need to do that. I'm ok with the inefficiency for now. 
* Mailman also supports mariadb/mysql - but their examples were using postgres. Considering that mariadb's container doesn't support multiple databases in a single container, I didn't care too much to standardize on a single db. I think postgres does support multiple dbs per container, but not sure if mailcow can support postgres. If so, using postgres would probably be the db i would choose to standardize on, but thats just me. 

###  Affected Containers

This does add anything to any existing containers. These are new instructions/examples for users to follow if they wanted to configure an all-in-one server with mailcow, borg, and mailman. 

## Did you run tests?

Deployed to my production instance.

### DIFF
--- a/data/Dockerfiles/caddy/Dockerfile
+++ b/data/Dockerfiles/caddy/Dockerfile
@@ -1,0 +1,14 @@
+# This dockerfile builds a custom caddy process, configured with optional 
+# modules for caddy-uwsgi-transport (for mailman) and caddy-events-exec for 
+# reloading dovecot/postfix when new certificates are issued
+# to build, do 'docker compose build' before running 'docker compose up -d'
+ARG CADDY_VERSION=2.9.1
+FROM caddy:${CADDY_VERSION}-builder AS builder
+
+RUN xcaddy build \
+    --with github.com/BadAimWeeb/caddy-uwsgi-transport \
+    --with github.com/mholt/caddy-events-exec
+
+FROM caddy:${CADDY_VERSION}-alpine
+RUN apk add jq
+COPY --from=builder /usr/bin/caddy /usr/bin/caddy
--- a/data/conf/borgmatic/etc/config.yaml
+++ b/data/conf/borgmatic/etc/config.yaml
@@ -1,0 +1,35 @@
+source_directories:
+    - /mnt/source/vmail
+    - /mnt/source/crypt
+    - /mnt/source/redis
+    - /mnt/source/rspamd
+    - /mnt/source/postfix
+    - /mnt/source/docker-mailman
+repositories:
+    - path: ssh://root@example.org/./mailcow
+      label: backup1
+exclude_patterns:
+    - '/mnt/source/postfix/public/'
+    - '/mnt/source/postfix/private/'
+    - '/mnt/source/rspamd/rspamd.sock'
+
+keep_hourly: 24
+keep_daily: 7
+keep_weekly: 4
+keep_monthly: 6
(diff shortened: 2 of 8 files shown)


midzelis:
Upstream contribution here: https://github.com/maxking/docker-mailman/pull/743

MAGICCC:
That's maybe something for https://docs.mailcow.email/

DerLinkman:
Yep definately. I'll won't merge this inside the normal project. Sorry :(

Issue #6042, opened 2024-08-21
2024-08a break dns resolution


osos:
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... I understand that not following the below instructions will result in immediate closure and/or deletion of my issue.
- [X] ... I have understood that this bug report is dedicated for bugs, and not for support-related inquiries.
- [X] ... I have understood that answers are voluntary and community-driven, and not commercial support.
- [X] ... I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

### Description

```plain text
After sudo ./update.sh unbound failed with "unable to ping 1.1.1.1 and so on.

Works fine outside docker.

Now nothing can get access to internet due to
```

### Logs:

```plain text
unbound-mailcow-1    | 2024-08-21T04:53:22.654519511Z 2024-08-21 06:53:22: Healthcheck: DNS Resolution Failed on attempt 3 for hub.docker.com! Trying again...
unbound-mailcow-1    | 2024-08-21T04:53:22.659959708Z 2024-08-21 06:53:22: Healthcheck: DNS Resolution not possible after 3 attempts for hub.docker.com... Gave up!
unbound-mailcow-1    | 2024-08-21T04:53:22.665430854Z 2024-08-21 06:53:22: Healthcheck: Too many DNS failures (1 failures allowed, you got 3 failures), marking Healthcheck as unhealthy...

postfix-mailcow-1    | 2024-08-21T04:53:39.359919355Z Aug 21 06:53:39 1e2aa1500bac postfix/dnsblog[444]: warning: dnsblog_query: lookup error for DNS query 27.162.32.193.bl.spameatingmonkey.net: Host or domain name not found. Name service error for name=27.162.32.193.bl.spameatingmonkey.net type=A: Host not found, try again
postfix-mailcow-1    | 2024-08-21T04:53:39.359967340Z Aug 21 06:53:39 1e2aa1500bac postfix/dnsblog[447]: warning: dnsblog_query: lookup error for DNS query 27.162.32.193.bl.ipv6.spameatingmonkey.net: Host or domain name not found. Name service error for name=27.162.32.193.bl.ipv6.spameatingmonkey.net type=A: Host not found, try again
[...]

PR #6311, opened 2025-02-17
Fix `check_dns()` hard-coded IP address in ..`unbound/healthcheck.sh`

andyraddatz:
## Contribution Guidelines

### Short Description

Fix to hard-coded IP address that causes `healthcheck.sh` to fail for unbound during `docker compose up -d`

## Did you run tests?

Yes

### What did you tested?

Failed to pass healthceck on startup because of DNS errors before change.

Healthcheck succeeded and I was able to log in successfully after `docker compose up -d`

### DIFF
--- a/data/Dockerfiles/unbound/healthcheck.sh
+++ b/data/Dockerfiles/unbound/healthcheck.sh
@@ -50,7 +50,7 @@ local failures=0
 for domain in "${domains[@]}" ; do
     success=false
     for ((i=1; i<=3; i++)); do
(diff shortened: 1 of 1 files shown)


andyraddatz:
Fixes https://github.com/mailcow/mailcow-dockerized/issues/6042

DerLinkman:
Will not merge this. This "fix" is changing the internal healthcheck to outside dns resolution which is breaking the sense of this.

Again: The healthcheck should check if a dns resolution is possible over localhost inside outbound which is then asking the own unbound resolver, not the resolver from the docker network, which is the host network.

PR #6332, opened 2025-02-26
compose: use ghcr.io for new/current mailcow docker images

DerLinkman:
## Contribution Guidelines

### Short Description

This PR include the swap of our ownly build docker images from Docker Hub to Github Container Registry (or short ghcr.io).

Due to Docker Hubs announcement regarding ratelimit an IP to 10 Pulls per Hour (and mailcow in total uses 18 containers) we would ran into a bottleneck, when all containers need an update (rarely the case but possible).

That's why we switch the shipment for the latest images beginning with 2025-02 to Github instead (until they come up with funny ratelimit ideas which are currently not the case (at least in my mind)

###  Affected Containers

All except not our images, which are:

- mariadb
- ofelia
- ipv6nat
- redis
- memcached

### What did you tested?

The normal mailcow startup.

As only the image location changed it was important to test the pull too.

The pull of the new images worked as expected, also from a not dev machine.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,7 @@
 services:
 
     unbound-mailcow:
-      image: mailcow/unbound:1.23
+      image: ghcr.io/mailcow/unbound:1.23
       environment:
         - TZ=${TZ}
         - SKIP_UNBOUND_HEALTHCHECK=${SKIP_UNBOUND_HEALTHCHECK:-n}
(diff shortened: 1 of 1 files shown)


patschi:
Note: The reason for this switch were the new [rate limits by Docker Hub](https://docs.docker.com/docker-hub/usage/) of 10 pulls per IP per hour. As one mailcow update/install would exceed this, we have decided to migrate over to the GitHub Container Registry.

PR #6365, opened 2025-03-10
dkim: Add support for 3072 and 4096 bit RSA keys

### LINKED ISSUE #6364: DKIM: Support 3072 and 4096 bit RSA keys
### Summary

Currently, DKIM keys of arbitrary size can be uploaded manually, but generating keys is limited to 1024 and 2048 bit keys. Also, uploaded 3072 bit keys are incorrectly labeled as 2048 bit keys in the UI.

This issue proposes to allow generation of 1024, 2048 (default), 3072 and 4096 bit RSA keys directly in the UI and to correctly label keys of all valid key sizes.

### Motivation

Users may wish to choose larger key sizes for improved security.

### Additional context

[RFC 8031, section 3.2](https://datatracker.ietf.org/doc/html/rfc8301#section-3.2) states that

> Verifiers MUST be able to validate signatures with keys ranging from 1024 bits to 4096 bits.

marvinruder:
## Contribution Guidelines

### Short Description

This PR implements generation of 3072 and 4096 bit RSA keys for DKIM directly in the UI and correctly labels 3072 bit RSA keys.

Fixes #6364

###  Affected Containers

- php-fpm-mailcow

### What did you tested?

- Generated 3072 and 4096 keys using the UI
- Used new key size options in domain template modals
- Created new domain with new key size option selected

- Keys were successfully generated and labelled
- Domain templates were successfully adapted
- Key was automatically generated for the new domain

image

### DIFF
--- a/data/web/api/openapi.yaml
+++ b/data/web/api/openapi.yaml
@@ -409,7 +409,7 @@ paths:
                   description: a list of domains for which a dkim key should be generated
                   type: string
                 key_size:
-                  description: the key size (1024 or 2048)
+                  description: the key size (1024, 2048, 3072 or 4096)
                   type: number
               type: object
       summary: Generate DKIM Key
--- a/data/web/inc/functions.dkim.inc.php
+++ b/data/web/inc/functions.dkim.inc.php
@@ -241,7 +241,7 @@ function dkim($_action, $_data = null, $privkey = false) {
           $dkimdata['length'] = "1024";
         }
         elseif (strlen($dkimdata['pubkey']) < 564) {
-          $dkimdata['length'] = "2048"
+          $dkimdata['length'] = "2048";
         }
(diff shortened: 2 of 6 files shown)


marvinruder:
@DerLinkman While I support having an opt-in option to use larger keys for users who prefer those, I am not sure whether 4096 bit keys are a sensible default for everyone. RFC 8301 requires verifiers to validate those signatures, but that RFC is relatively new and some older mail servers may not yet be compliant. Specifically, [this article](https://wander.science/articles/recommendations-dkim-signing/) states that Cisco Secure Email Gateway does not support keys larger than 2048 bits. I would agree with their conclusion that

> 2048 bits is a safe choice between security and compatibility

DerLinkman:
Hm hm ok, good concern!

Yeah i'll rollback the default to 2048 then

DerLinkman (approved):

PR #6398, opened 2025-03-25
feat(olefy): Allow disabling Olefy

### LINKED ISSUE #6389: Allow disabling Olefy
### Summary

Analysis of Office documents as done by Olefy is not an integral part of mailcow. I propose to add a `SKIP_OLEFY` configuration parameter that works like e.g. `SKIP_CLAMD` and disables the Olefy container as well as health checks, Watchdog checks and container information on the dashboard.

### Motivation

Disabling containers providing unused features utilizes less system resources.

### Additional context

_No response_

marvinruder:
## Contribution Guidelines

### Short Description

This PR adds a `SKIP_OLEFY` configuration parameter that disables the Olefy container as well as health checks, Watchdog checks and container information on the dashboard.

* Fixes #6389

###  Affected Containers

- olefy-mailcow
- rspamd-mailcow
- php-fpm-mailcow
- watchdog-mailcow

### What did you tested?

Create a new mailcow instance within a `docker:dind` container with `SKIP_OLEFY=y`.

- Olefy container log shows `SKIP_OLEFY=y, skipping Olefy...`
- Admin dashboard does not show the Olefy container
- Watchdog logs do not show entries for Olefy container


FreddleSpl0it (changes requested):
Thank you. But I think we also have to disable it in rspamd. The File `data/conf/rspamd/local.d/external_services.conf` specifies the Olefy Service. The best would be to delete the file and let rspamd create it at container startup in `data/conf/rspamd/override.d/external_services.conf`.

MAGICCC:
I guess you may also need to add in update.sh something following just for SKIP_OLEFY
```
  elif [[ "${option}" == "DISABLE_NETFILTER_ISOLATION_RULE" ]]; then
    if ! grep -q "${option}" mailcow.conf; then
      echo "Adding new option \"${option}\" to mailcow.conf"
      echo '# Prevent netfilter from setting an iptables/nftables rule to isolate the mailcow docker network - y/n' >> mailcow.conf
      echo '# CAUTION: Disabling this may expose container ports to other neighbors on the same subnet, even if the ports are bound to localhost' >> mailcow.conf
      echo 'DISABLE_NETFILTER_ISOLATION_RULE=n' >> mailcow.conf
    fi
```

marvinruder:
> I guess you may also need to add in update.sh something following just for SKIP_OLEFY
> 
> ```
>   elif [[ "${option}" == "DISABLE_NETFILTER_ISOLATION_RULE" ]]; then
>     if ! grep -q "${option}" mailcow.conf; then
>       echo "Adding new option \"${option}\" to mailcow.conf"
>       echo '# Prevent netfilter from setting an iptables/nftables rule to isolate the mailcow docker network - y/n' >> mailcow.conf
>       echo '# CAUTION: Disabling this may expose container ports to other neighbors on the same subnet, even if the ports are bound to localhost' >> mailcow.conf
>       echo 'DISABLE_NETFILTER_ISOLATION_RULE=n' >> mailcow.conf
>     fi
> ```

Yes, that seems useful. I was looking for a similar instruction for `SKIP_CLAMD`, but did not find one. I might add it for both parameters then.

FreddleSpl0it (approved):

jman-schief:
hey @marvinruder I've deployed the new Mailcow update that contains this patch, thanks for working on this.

If I understand correctly, the intent in #6389 was that `SKIP_OLEFY=y` would avoid downloading and running the container, correct? In my case the Docker image is downloaded again and started. I am using an existing Mailcow installation, not a new one.

Am I missing anything?

Thanks

EDIT: fixed freudian slip :sweat_smile:

marvinruder:
> hey @marvinruder I've deployed the new Mailcow update that contains this patch, thanks for working on this.
> 
> 
> 
> If I understand correctly, the intent in #6389 was that `SKIP_OLEFY=n` would avoid downloading and running the container, correct? In my case the Docker image is downloaded again and started. I am using an existing Mailcow installation, not a new one.
> 
> 
> 
> Am I missing anything?
> 
> 
> 
> Thanks

`SKIP_OLEFY=y` (not `=n`, as it says _SKIP_) will still download and start the container, but the container will only print `Skipping Olefy` and do nothing. `SKIP_OLEFY=n` is the default setting, it will run the container normally and not skip anything. This behavior is equivalent to the already existing `SKIP_CLAMD` variable, and I assume this is the best you can do in “disabling” containers using environment variables only.

If you want to remove the container completely, you can override it within your `docker-compose.override.yaml` file. Note that this may still not prevent the update script from pulling the images, but no containers will be created.

jman-schief:
ok, so if the image is downloaded, the container created and started, what is *"less system resources"* referring to in the mentioned issue?

marvinruder:
> ok, so if the image is downloaded, the container created and started, what is *"less system resources"* referring to in the mentioned issue?

A regular Olefy container uses lots of CPU time and memory to do its tasks, a disabled Olefy container uses no CPU time, as it does nothing, and uses only as much memory as it takes a Python runtime to print a message and then idle forever.

PR #6460, opened 2025-04-04
[Web] Improve clarity of LDAP SSL/TLS settings

FreddleSpl0it:
## Contribution Guidelines

### Short Description

This PR adds Tooltips for the LDAP SSL/TLS options. 

###  Affected Containers

- PHP-FPM


patschi:
In regards to: `TLS is recommended over SSL`. I'm not so sure about that. There are pages [1][2][3] which suggest and state seeing SSL (via TCP/636) as deprecated and prefer StartTLS approach. 

But StartTLS has the main security-wise disadvantage, that it starts with plain-text communication before upgrading to TLS, and can also be forced to downgrade. Effectively reducing security. So I also found a discussion of OpenLDAP where they are looking to get the dedicated port (SSL) recommended [4][5].

So I'm not sure what is true now. I'd change it to more neutral and leave this up to the customer - the dedicated SSL port seems however more secure, as there is now downgrade possibility which favors security.

Also: I think the "SSL" option does not necessarily refer to the old SSL transport encryption and also includes TLS. It's just some 'legacy' naming.

---
[1] https://www.openldap.org/faq/data/cache/605.html
[2] https://ldaprecord.com/docs/core/v3/configuration/#ssl--tls:~:text=in%20Active%20Directory.-,TLS%20is%20recommended%20over,SASL,-This%20boolean%20option
[3] https://kb.symas.com/configuration/overview-of-tls-and-ldaps
[4] https://unix.stackexchange.com/questions/607560/why-is-ldap-with-starttls-preferred-over-ldaps
[5] https://lists.openldap.org/hyperkitty/list/openldap-technical@openldap.org/thread/IXEZAGCOTL5CTSOFNFDE65LURQET6ZZV/#LZ4N3JHGKZVDXRIXNRAM4SHZ44B5U4BP

FreddleSpl0it:
@patschi i removed the TLS is recommendations from the tooltips

anolting:
Please revert this PR. The discussion about TLS over SSL was finished years ago against SSL. SSL has seen the last specification update in 1996. It has various design issues (POODLE attack against SSL 3.0) and as of this attack  it was lastly declared as deprecated by IETF in June 2015 with RFC 7568. TLS is the successor of SSL and is still under development. TLS 1.0 and 1.1 also have mayor security issues and shouldn't used any more. TLS 1.2 will be deprecated in near future as it has a number of disadvantages against 1.3 for instance it allows a number of already deprecated declared encryption protocols, it does not provide PFS (Perfect Forward Security) due to not supporting ECDHE or DHE, not supporting encrypted ServerHello and so on. All this are reworked and added to TLS 1.3.
Therefore this PR will spread false information about the security of TLS and SSL.

smarsching:
@anolting I do not quite agree. Yes, SSL is dead, no one is using it any longer.

However, this discussion was not about SSL vs TLS, but about LDAPS (a.k.a. LDAP over TLS) or LDAP with STARTTLS. I agree that optimally, the term SSL should be avoided completely, but I am also afraid that calling one option “Use TLS” and the other option “Use STARTTLS” may confuse users who are not aware of the intricacies of LDAP (or other protocols, where there is the TLS vs STARTTLS option).

I think that we have kind of a chicken and egg problem here: A lot of software and documentation is still using the term SSL when actually TLS is meant. Just look how often you read the term “SSL certificate” and how rarely “TLS certificate” (and even more rarely the actually correct term “X.509” certificate). But if one software starts using different terms, it can easily lead to confusion, in particular in the context of TLS vs STARTTLS…

PR #6483, opened 2025-04-12
Allow additional domains in OAuth2 redirect URLs

PseudoResonance:
## Contribution Guidelines

### Short Description

Currently with OAuth2, you must set a redirect URL that the OAuth provider will redirect users to after authentication.

If you want mailcow to be available on multiple domains (for example a dedicated SOGo subdomain), OAuth2 login will redirect back to the primary domain, leaving the additional domain unusable.

This lets you define additional redirect URLs, and attempts to match them based on your current domain to send to the provider.

###  Affected Containers

- nginx
- php-fpm

### What did you tested?

Adding/editing identity providers with 0, 1 and 2 additional URLs specified.

Logging in with OAuth2 from 2 domains pointing to mailcow.

I was able to login from both my primary, and webmail domains. The OAuth provider got the correct redirect URL depending on which domain I started on, and redirected me back to the appropriate URL.

### DIFF
--- a/data/web/api/openapi.yaml
+++ b/data/web/api/openapi.yaml
@@ -5847,6 +5847,7 @@ paths:
                           client_id: "mailcow_client"
                           client_secret: "*"
                           redirect_url: "https://mail.mailcow.tld"
+                          redirect_url_extra: ["https://extramail.mailcow.tld"]
                           version: "26.1.3"
                           default_template: "Default"
                           mappers:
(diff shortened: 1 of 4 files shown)


PseudoResonance:
This somewhat goes together with #6482, as my end goal is to have 2 simple URLs for my parents to use, one for SOGo and another for mailcow. They primarily want to use SOGo, but may need to access mailcow occasionally, so I want to have the ease of navigation to SOGo with a single link, without making it harder to access mailcow.

I wasn't totally sure how to implement this without affecting more standard setups though. Hopefully this is acceptable, but of course I'm happy to make any changes if necessary.

FreddleSpl0it:
Thank you this is a great enhancement.

FreddleSpl0it (approved):

PR #6499, opened 2025-04-26
Bulgarian language added

Angelov9004:
This pull request adds full Bulgarian (bg-bg) translation to mailcow.  
I have added the new `lang.bg-bg.json` file and updated `vars.inc.php` to include the language option.

Thank you for considering this contribution! 🇧🇬


milkmaker:
Thanks for contributing!

MAGICCC:
Nice thanks
Could you rebase + sort [data/web/inc/vars.inc.php](https://github.com/mailcow/mailcow-dockerized/pull/6499/files#diff-e27671f283e93d0f8cab7769916665b2b3b085dc540999b3a4dee9651a202656) alphabetically?

krassle:
What happened here? `lang.bg-bg.json` got created (#6463) but it's empty...

@Angelov9004 
Why did you close your PR?

I 'd like to vote for reopening and merging, as the translation is indeed very good.

MAGICCC:
@krassle the json file got created in https://github.com/mailcow/mailcow-dockerized/pull/6463 because I created it in weblate since I received an email that they want to translate the language.
But if the translation is fine we could use it anyway.

krassle:
@MAGICCC 
It appears the original author is not interested in their contribution any longer, so I'd suggest using the translation as is.

I have no experience with Weblate, but just created an account and if we could somehow reverse-convert this json lang file back to weblate format, it would be really helpful. And if not, I might simply "copy-paste" it string by string... :sweat: 

I've just started using mailcow, but I'm going to try updating / maintaining the translations for the Bulgarian language in the future.

Thanks! :+1:

Issue #6104, opened 2024-10-14
Moving mails by functions.quarantine.inc.php to inbox failed


vigeland:
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... I understand that not following the below instructions will result in immediate closure and/or deletion of my issue.
- [X] ... I have understood that this bug report is dedicated for bugs, and not for support-related inquiries.
- [X] ... I have understood that answers are voluntary and community-driven, and not commercial support.
- [X] ... I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

### Description

```plain text
In some special cases functions.quarantine.inc.php quick_release does not work
I am not a php programmer, but debugging the file I am able to say 
$detail_row['msg'] = preg_replace('/^X-Spam-Flag: (.*)/', 'X-Pre-Release-Spam-Flag $1', $detail_row['msg']); 
sometimes not work for me. 
Dumping $detail_row['msg'] before and after call, show the same "X-Spam-Flag: YES" so moving the mail to inbox does not work. The mail is sent twice to junk folder.
It occurs for one user on every mail.

Using the last docker release. 2024-08a / x86
```

### Logs:

```plain text
X-Spam-Score: 8.5 (++++++++)
X-Spam-Flag: YES
X-Spam-Report: Action: add header

and 

X-Spam-Score: 8.5 (++++++++)
X-Spam-Flag: YES
X-Spam-Report: Action: add header
```

### Steps to reproduce:

```plain text
-
```

### Which branch are you using?

master

### Which architecture are you using?

x86

### Operating System:

Debian 6.1.90-1 

### Server/VM specifications:

8 GB / 4 core

### Is Apparmor, SELinux or similar active?

no

### Virtualization technology:

KVM

### Docker version:

 26.1.2

### docker-compose version or docker compose version:

docker-compose version 1.29.2

### mailcow version:

2024-08a

### Reverse proxy:

ngnix

### Logs of git diff:

```plain text
-
```

### Logs of iptables -L -vn:

```plain text
-
```

### Logs of ip6tables -L -vn:

```plain text
-
```

### Logs of iptables -L -vn -t nat:

```plain text
-
```

### Logs of ip6tables -L -vn -t nat:

```plain text
-
```

[...]

PR #6506, opened 2025-04-29
[Fix] Moving mails by functions.quarantine.inc.php to inbox failed

mrclschstr:
## Contribution Guidelines

### Short Description

This PR fixes a small bug in the regex of the function used to quickly release mails from the spam folder. Further details can be found in issue #6104

###  Affected Containers

To be honest, I don't know for sure which container the file is being used by. The directory `/data/web` is used by the following two containers:

- `php-fpm-mailcow`
- `nginx-mailcow`

Does the version have to be increased for both containers? I still need some help here.

### What did you tested?

I have primarily tested moving emails recognized as spam to the inbox via my mail client (IMAP).

Moving false positive spam mails back to my inbox now works again without any problems.

### DIFF
--- a/data/web/inc/functions.quarantine.inc.php
+++ b/data/web/inc/functions.quarantine.inc.php
@@ -169,7 +169,7 @@ function quarantine($_action, $_data = null) {
           }
         }
         elseif ($release_format == 'raw') {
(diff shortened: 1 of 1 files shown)


FreddleSpl0it (approved):

FreddleSpl0it:
Thank you for the contribution!
You don't have to increase the image tags when editing PHP files.
And even if increasing is needed, we'll handle that during the build process of the images anyway.

PR #6444, opened 2025-04-01
Fix Broken Whitelisting/Blacklisting Functionality in mailcow UI

PhoenixPeca:
## Contribution Guidelines

* [x] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree them

## What does this PR include?

Include the local resolver IP in the python code instead of blindly relying on resolver config from the base image OS to have it (Alpine)..

### Short Description

Without this PR, the whitelist and blacklist functionality from mailcow UI doesnt work:
image

Container netfilter-mailcow log:
image

###  Affected Containers

- netfilter-mailcow

## Did you run tests?

Yes, locally... This PR now allows netfilter to get the list of blacklist and whitelist from mailcowUI via fail2ban, see screenshot below.
image

### What did you tested?

netfilter-mailcow container tested to see if Exception still occurs after this change... This update fixed it.

### What were the final results? (Awaited, got)

Got the expected behavior... Bug fixed.

PR #6520, opened 2025-05-09
Fix Broken Whitelisting/Blacklisting Functionality in mailcow UI

PhoenixPeca:
## Contribution Guidelines

## What does this PR include?

Include the local resolver IP in the python code instead of blindly relying on resolver config from the base image OS to have it (Alpine)..

### Short Description

I'm reopening this PR as per https://github.com/mailcow/mailcow-dockerized/pull/6444#issuecomment-2863132636

Without this PR, the whitelist and blacklist functionality from mailcow UI doesnt work:
image

Container netfilter-mailcow log:
image

###  Affected Containers

- netfilter-mailcow

## Did you run tests?

Yes, locally... This PR now allows netfilter to get the list of blacklist and whitelist from mailcowUI via fail2ban, see screenshot below.
image

### What did you tested?

netfilter-mailcow container tested to see if Exception still occurs after this change... This update fixed it.

Got the expected behavior... Bug fixed.

### DIFF
--- a/data/Dockerfiles/netfilter/main.py
+++ b/data/Dockerfiles/netfilter/main.py
@@ -320,7 +320,8 @@ def isIpNetwork(address):
   return True
 
 def genNetworkList(list):
-  resolver = dns.resolver.Resolver()
+  resolver = dns.resolver.Resolver(configure=False)
+  resolver.nameservers = ['127.0.0.1']
   hostnames = []
   networks = []
   for key in list:


FreddleSpl0it:
Can you explain why this fixes the Problem?
My understanding is, that without defining a DNS Server, it will use the DNS Servers from `/etc/resolv.conf` and since the netfilter container runs in `network_mode: "host"` it will use the hosts `/etc/resolv.conf`.

PhoenixPeca:
Hi @FreddleSpl0it , based on my previous testing, this functionality got broken because of the recent upgrade of netfilter's alpine base image from the last major release I have installed... Probably some alpine system level config or behavior changes that caused the `main.py` script to not detect the DNS resolver correctly...

See this screenshot--it shows DNS nameserver error because it cannot detect it:

image

However, it seems that this issue is not happening in older versions of alpine and mailcow.. Also I pointed out here that the recent netfilter downgrade might have fixed it: https://github.com/mailcow/mailcow-dockerized/pull/6444#issuecomment-2768094467

This means this PR may only be relevant if we intend to upgrade netfilter's alpine again..

PhoenixPeca:
My investigation:

1. It seems that the python [resolver class](https://dnspython.readthedocs.io/en/latest/resolver-class.html), uses the `/etc/resolv.conf` as default reference for the nameserver info..
2. My default docker install does not set a fallback DNS resolver, so when container is started, the `/etc/resolv.conf` file inside the netfilter container is empty.

PR #6523, opened 2025-05-09
Feat/6522-shorten-service-names

teodorescuserban:
## Contribution Guidelines

### Short Description

Shortens the service and container names, but keeping the inital longer name as alias.

###  Affected Containers

All of them.

- acme
- clamd
- dockerapi
- dovecot
- memcached
- mysql
- netfilter
- nginx
- ofelia
- olefy
- php-fpm
- postfix
- redis
- rspamd
- sogo
- unbound
- watchdog

### What did you tested?

Yes. Create the stack from scratch.
Everything seems to be working normally.

Unfortunately, even though some usual things work, there are scripts that rely on the actual service name having "-mailcow" as suffix (update.sh, reload-configurations.sh,  watchdog.sh, and so on).

Leaving this as draft PR :\ 
Maybe this is not even the real source, maybe the source is mailcow _not_ dockerized...

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,6 +1,6 @@
 services:
 
-    unbound-mailcow:
+    unbound:
       image: ghcr.io/mailcow/unbound:1.24
       environment:
         - TZ=${TZ}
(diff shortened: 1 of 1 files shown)


DerLinkman:
Why should we merge this?

patschi:
Sounds to be rude. To be honest, it would be much cleaner and how "the name should be". The container names in a stack always get a corresponding prefix, e.g. `mailcowdockerized` - ending in full names like `mailcowdockerized-postfix-mailcow-1`. 

Without the `-mailcow` suffix, it would look more clean like `mailcowdockerized-postfix-1`.

With `aliases:` it should not have any impact, but not sure and it could be indeed a major breaking change.

teodorescuserban:
Unfortunately, it looks like the `-mailcow` suffix is used in more than a couple of scripts, so probably the PR needs a lot more work.

PR #6546, opened 2025-05-16
[fix] Clamav score for FP CLAM_SECI_SPAM symbol

dragoangel:
## Contribution Guidelines

* [ ] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree them

## What does this PR include?
Decrease CLAM_SECI_SPAM score to 1 instead of 5 to low quality of database and significant false-positive rate

### Short Description
`spam_marketing.ndb` is very false positive database that listing hex values of emails without proper separation, f.e. it contains hashes like `SecuriteInfo.com.Spam-120111:0:*:726a40676d61696c2e636f6d` which decoded as `rj@gmail.com` and doesn't contain any limitations, meaning that any email body that contains `rj@gmail.com` in any way will be treated as spam and get 5 score. This total nonsence. In addition `SecuriteInfo.com` do not answer on questions by email or contact form to exclude some hashes from their databases based on my personal experience.

###  Affected Containers

- rspamd

## Did you run tests?
no, it not necessary for this small change

### DIFF
--- a/data/conf/rspamd/local.d/composites.conf
+++ b/data/conf/rspamd/local.d/composites.conf
@@ -76,7 +76,7 @@ ENCRYPTED_CHAT {
 CLAMD_SPAM_FOUND {
   expression = "CLAM_SECI_SPAM & !MAILCOW_WHITE";
   description = "Probably Spam, Securite Spam Flag set through ClamAV";
-  score = 5;
+  score = 1;
 }
 
 CLAMD_BAD_PDF {


data/conf/rspamd/local.d/composites.conf
```diff
@@ -76,7 +76,7 @@ ENCRYPTED_CHAT {
 CLAMD_SPAM_FOUND {
   expression = "CLAM_SECI_SPAM & !MAILCOW_WHITE";
   description = "Probably Spam, Securite Spam Flag set through ClamAV";
-  score = 5;
+  score = 1;
```
DerLinkman, line 79: Lowering it is ok, but i would lower it to 2 instead of 1.

DerLinkman (changes requested):

data/conf/rspamd/local.d/composites.conf
dragoangel: I by myself checked how that db works, honestly dropped it immediately. It is not worth a score of 2, and honestly it is not worth even a score of 1. This db has conditions like "jr@gmail.com", which will fire on anything, like abramsjr@gmail.com, anythingjr@gmail.comunity... They not even put a space or <> in their db even they could lower FP by that, they could do this DB in YARA format which would not be so FP as YARA allows nice regexes, but no..., it's one total FP, and they are not responding to email to fix issues in their DB or any FP cases they put there.

dragoangel: On every site this spammarketing db listed as highly false positive and not recommended for a use :), Google about it a bit and take this db in decoded format, you will understand.

milkmaker:

Issue #5121, opened 2023-03-09
watchdog (unbound): nagios check_dns plugin SEGFAULT


anokfireball:
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... I understand that not following the below instructions will result in immediate closure and/or deletion of my issue.
- [X] ... I have understood that this bug report is dedicated for bugs, and not for support-related inquiries.
- [X] ... I have understood that answers are voluntary and community-driven, and not commercial support.
- [X] ... I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

### Description

```plain text
Ever since my last mailcow update, watchdog has been spitting out alerts about unbound-mailcow. After investigating these today, it seems like `/usr/lib/nagios/plugins/check_dns` is at fault: It reliably segfaults on every single call.
```

### Logs:

```plain text
[root@mail ~]# docker logs $(docker ps -qf name=watchdog-mailcow) -n 100
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
DNSSEC check succeeded
[...]

Issue #5033, opened 2023-01-30
Segfault in /usr/lib/nagios/plugins/check_dns in Watchdog container


Lorak-mmk:
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... I understand that not following the below instructions will result in immediate closure and/or deletion of my issue.
- [X] ... I have understood that this bug report is dedicated for bugs, and not for support-related inquiries.
- [X] ... I have understood that answers are voluntary and community-driven, and not commercial support.
- [X] ... I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

### Description

```plain text
I see A LOT of segfaults for `/usr/lib/nagios/plugins/check_dns` in my journalctl.
See example journalctl entries in "Logs" section.


I can see that it's being run here, by watchdog container: https://github.com/mailcow/mailcow-dockerized/blob/640f535e9995f27fcc1541a60f0a56648d0d66cc/data/Dockerfiles/watchdog/watchdog.sh#L272

When I try to execute the command myself in the container, I also get the segfault (see gdb session in "Logs" section), but it seems that the binary has no symbols, so the stacktrace is quite useless.
```

### Logs:

```plain text
Crash in journalctl:

Jan 30 22:02:47 vps770689.ovh.net systemd-coredump[3126064]: [🡕] Process 3126059 (check_dns) of user 0 dumped core.
                                                             
                                                             Module /usr/lib/nagios/plugins/check_dns without build-id.
                                                             Module /lib/libcrypto.so.3 without build-id.
                                                             Module /lib/ld-musl-x86_64.so.1 without build-id.
                                                             Module linux-vdso.so.1 with build-id 5cbcbeaed17f6205501c99ad3b84bfa3cb42f965
                                                             Stack trace of thread 1646727:
                                                             #0  0x00005576fbe22ffc n/a (/usr/lib/nagios/plugins/check_dns + 0x6ffc)
                                                             ELF object binary architecture: AMD x86-64
```

GDB session:
```
ea9217b7bd96:/# gdb /usr/lib/nagios/plugins/check_dns
GNU gdb (GDB) 12.1
Copyright (C) 2022 Free Software Foundation, Inc.
[...]

PR #6685, opened 2025-08-25
watchdog: use dig instead of check_dns

maxi322:
This issue is mentioned here: https://github.com/mailcow/mailcow-dockerized/issues/5033 and https://github.com/mailcow/mailcow-dockerized/issues/5121

I think it can occour on low memory servers. I myself experienced the problem when updating my mailcow server from debian 12 to 13 on a 4GB RAM machine. Before it was all fine, after the update check_dns would constantly segfault.

I observed quiete a significant memory usage when running check_dns on another machine.
Therefore using dig in a wrapper script should deliver similar experience and fix the segfault problem.

## Contribution Guidelines

### Short Description

Replace nagios check_dns with dig in a wrapper script to behave pretty much the same.

###  Affected Containers

- watchdog

### What did you tested?

The script itself. Building and using the container

The segfault problem was gone, and unbound should still be monitored as before.

### DIFF
--- a/data/Dockerfiles/watchdog/Dockerfile
+++ b/data/Dockerfiles/watchdog/Dockerfile
@@ -16,7 +16,6 @@ RUN apk add --update \
   fcgi \
   openssl \
   nagios-plugins-mysql \
-  nagios-plugins-dns \
   nagios-plugins-disk \
   bind-tools \
   redis \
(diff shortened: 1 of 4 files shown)


DerLinkman (approved):
Thanks!

PR #6686, opened 2025-08-26
pf/php: add mta-sts support (outbound)

### LINKED ISSUE #1809: MTA-STS - secured mailtransport
Just as info:

Since some days MTA-STS has been officially published as [RFC 8461](https://tools.ietf.org/html/rfc8461)

MTA-STS (the full name SMTP Mail Transfer Agent Strict Transport Security) is a new standard that aims to improve the security of SMTP by enabling domain names to opt into strict transport layer security mode that requires authentication (valid public certificates) and encryption (TLS).

DerLinkman:
## Contribution Guidelines

### Short Description

This PR includes outbound MTA-STS Support for mailcow based on the BSI "E-Mail Security Year 2025" where mailcow is contributing to.

It will also allow Admins to setup MTA-STS by editing the desired Domain in mailcow UI.

mailcow will automatically create such TXT file with the desired values and will request a SSL Certificate for it.

> [!CAUTION]
> Users who have setup MTA-STS manually within mailcow may be adviced to switch to the build in one or keeping the setting in Domain edit untouched.

To get the requested DNS Value visit the updated DNS Tool in the Domain section.

Postfix has also updated to include https://github.com/Zuplu/postfix-tlspol inside the container to validate MTA-STS on destination Domains but still prefering DANE.

Fixes #1809 

###  Affected Containers

- postfix-mailcow
- php-fpm-mailcow
- nginx-mailcow

## Did you run tests?

Yes.

### What did you tested?

Outbound/Inbound Mailing, MTA-STS TXT File + DNS Record generation + validation

Postfix will correctly detect MTA-STS or DANE of the Destination Address and is sending the Mails with a forced TLS encryption to the destination Server.

If no Policy has been found the mail will normally been send via StartTLS.

The MTA-STS Record will generate properly and will change according to a change done via Domain Edit.


DerLinkman:
> Will the Mailcow docs be updated with a section on this?

Yes! Within the next few days.

CLechleitner42:
### TLDR:
An update to Mailcow 2025-09b made it temporarily impossible to send e-mails to some particular domains, and because of the error messages in the logs (and the prominent tlspol warning in the release notes) I suspected the MTA-STS integration and hence this PR at first.
In the end it was a MTU problem, but others might fall in the same trap and follow the same train of thought, so I document this here after all, hope this is OK.
I have some other observations around the new feature too, see at the bottom of this comment.

### Full version:

After updating to Mailcow 2025-09b, i.e. `ghcr.io/mailcow/postfix:1.81` and `ghcr.io/mailcow/postfix-tlspol:1.0`, we could not send e-mails to `@gmail.com` (and one other domain), and the error messages pointed at the new `postfix-tlspol` feature the release notes talk about:
~~~
2025-09-15T19:17:21+02:00 localhost postfix[868]: Sep 15 19:17:21 4387585e3a11 postfix/smtp[531]: warning: socketmap:inet:postfix-tlspol:8642:QUERY lookup error for "gmail.com"
2025-09-15T19:17:21+02:00 localhost postfix[868]: Sep 15 19:17:21 4387585e3a11 postfix/smtp[531]: warning: smtp_tls_policy_maps, next-hop destination "gmail.com": policy table lookup error
2025-09-15T19:17:21+02:00 localhost postfix[868]: Sep 15 19:17:21 4387585e3a11 postfix/smtp[531]: warning: TLS policy lookup for gmail.com/gmail-smtp-in.l.google.com: client TLS configuration problem
~~~

For some of the target domains I solved the problem by removing `socketmap:inet:postfix-tlspol:8642:QUERY` from `smtp_tls_policy_maps` in  `data/conf/postfix/main.cf` (which was only present in our prod server, not on our test server, why?).

However I still could not get through to a particular other domain, until I found the root cause: Our Hetzner vSwitch network (which allows arbitrary sharing of world IPs between multiple hardware nodes) requires a MTU of just 1400 (instead of usual 1500) and `docker-compose` ignores MTU settings in `/etc/docker/daemon.json` and creates networks with default MTU of 1500.

Our solution was to enforce that MTU on the `mailcow-network`: In our `docker-compose.override.yml` (which mostly changes logging to local rsyslogd) I added this at the end:
~~~
networks:
  mailcow-network:
    driver_opts:
      com.docker.network.driver.mtu: "1400"
~~~
Then of course
~~~
docker-compose  stop
docker network rm mailcowdockerized_mailcow-network
docker-compose  up -d
~~~
Which gave me `Error response from daemon: network aeb3...8ac3 not found` which was resolved with another `update.sh` run.

Voila.

### Additional observations

Additional observation 1: The additional container `postfix-tlspol` as well as the according part of `smtp_tls_policy_maps` was only present on our own prod server, not on our test server. I have not checked our customer mail server which will not get the update before a grace period of a few days at least. But I have no idea why on setup gets the tlspol stuff and another does not.

Additional observation 2: The DNS proposal for domains does not contain anything related to STS, as I would have expected from previous comments in this PR.

Additional observation 3: The MTA-STS tab in the domain settings is confusing. I suspect this is for e-mails sent to this domain, but the German text does not make clear where the information goes, nor does it make meaningful suggestions for what to put in the mandatory age and MX fields.

I also have checked https://docs.mailcow.email/getstarted/prerequisite-dns/#optional-dmarc-statistics, nothing there about MTA-STS yet.

### A note on the curse of MTUs

Besides Hetzner vSwitches, VPNs and SD-WANs sometimes require reduced MTUs.

3 facts makes these situations extra tricky:
- `docker` ignores MTU of ethernet interfaces and defaults docker networks to 1500
- `docker-compose` ignoring the `mtu` setting in docker's `daemon.json` and defaults to 1500, which is OK for internal communication but not for communication over ethernets with lower MTU
- many networking things still work, lke `ping` (unless you enforce large packets) or simple `telnet` stuff like `http` (probably dependend on the software used and the discovered path mtu), but some TLS stuff fails by sloooooowly running into timeouts or application errors after ages

mrclschstr:
> Our solution was to enforce that MTU on the `mailcow-network`: In our `docker-compose.override.yml

Isn't this already described here? https://docs.mailcow.email/getstarted/install/#mtu-not-equal-to-1500-eg-openstack

CLechleitner42:
> > Our solution was to enforce that MTU on the `mailcow-network`: In our `docker-compose.override.yml
> 
> Isn't this already described here? https://docs.mailcow.email/getstarted/install/#mtu-not-equal-to-1500-eg-openstack

Kind of.

But our initial change in `docker-compose.yml` got lost somewhere along the way, and that wasn't a problem for some time - until this new feature brought it up again. So as written, maybe others suffer the same fate and maybe search for solutions the same way and land here.

Also, `docker-compose.override.yml` is a far better place for any local adaptions because it won't be touch by any update or git actions.

Konstrukteur:
> tiny issue, my dns setup is still working, but differs from what mailcow expects. Eg:
> 
> Name	Type	Correct	state
> _mta-sts.mydomain.org	TXT	v=STSv1;id=20250913122400;	**?** v=STSv1; id=171727622660Z;
> I would like to see checkmarks :)
> 
> I have automated dnssetup with dnscontrol, which queries mailcow API already for DKIM Records. I fail to see how to query that timestamp for mta-sts.

I suspect you had an existing MTA-STS implementation, since mailcow reports an existing id of 171727622660Z. With the new MTA-STS implementation mailcow now creates dynamic ids for each domain. You have to update your DNS entry with the data from the "Correct" column. Mailcow will read this and check it against its value. If both are identical you will get a check.

Instead I suspect we will not get a green check for the _smtp._tls row, because mailcow has now way to verify the correctness of the provided email. It would still be nice to have mailcow report the actual value from the DNS entry .... just like for DMARC by the way ...

Konstrukteur:
> > > Will the Mailcow docs be updated with a section on this?
> > 
> > 
> > Yes! Within the next few days.
> 
> I don't want to be rude, but when can we expect the updated documentation? I'm looking forward to finally being able to use MTA-STS.

There has been a section on the documentation since 12.09.2025
https://docs.mailcow.email/post_installation/setup-mta-sts/?h=mta

It should cover most basic questions

mrclschstr:
> There has been a section on the documentation since 12.09.2025

Thank you very much, I overlooked that. Am I correct in understanding that I don't need to make any changes to the `mailcow.conf` file for MTS-STS to work properly? From the previous comments, it sounded to me as if you also had to adjust options (`ADDITIONAL_*`) there?

Konstrukteur:
> > There has been a section on the documentation since 12.09.2025
> 
> Thank you very much, I overlooked that. Am I correct in understanding that I don't need to make any changes to the `mailcow.conf` file for MTS-STS to work properly? From the previous comments, it sounded to me as if you also had to adjust options (`ADDITIONAL_*`) there?

In my case I deleted my mta-sts.* parts and it works. There are people that reported they needed them. I would try without and report back.

My understanding is still one does not need it, Mailcow creates these endpoints automatically just like imap, smtp, pop3, ….

Root-Core:
> what are you using the mail.* subdomain for? In my case for SOGO or the mailcow UI depending from path.

Same.

> one would have to renew all certificates I suspect.

I haven't tried it yet, but does mailcow renew the cert with the `mta-sts` subdomain included?
Or does it issue a second cert for this subdomain? @DerLinkman

Konstrukteur:
In my case I already had an mta-sts implementation running and existing LE certificates. I deleted mta-sts from my MAILCOW.conf and it is working.

What might have happened though is that mailcow did not try to create new certs, because it already had them. At least the ACME log did just report that it used what it found....

I will have to wait for when the certs are up for a renewal and see, if they are renewed or not.

It could well be that I will have to add it to ADDITIONAL_SAN again.

mrclschstr:
> So apparently mailcow does not take care of provisioning functional endpoints with certs automatically ...

At least for the `autodiscover` and `autoconfig` subdomains, they are automatically added to the certificate. So my question is simple: Does mailcow also do this for the `mta-sts` subdomain? If not, this would at least be inconsistent with the current behavior.

mrclschstr:
> Who tries?

I'm hoping more for a statement from @DerLinkman ;-)

Root-Core:
I don't know exactly why, but my certificate is now also valid for the `mta-sts` sub domain.
I rebooted the host, it seems mailcow realized it's missing a domain.

Maybe the DNS checker added some kind of flag or something.

-----

I had previously a setup, where the `mts-sts` domain was handled by another host and proxied to the mail server, with an valid wildcard cert. With the CNAME it is a bit cleaner and more reliable.

PR #6704, opened 2025-09-02
[Postfix] Prevent unwanted unauth relaying to other mailservers

offlinehoster:
In case of unwanted spamfloods your mailcow will relay incoming mails from other mailservers because of a wrong rcpt_to or from: header. I this case you have can secue your mailcow by adding those 2 restrictions, one for external relays and also for the internal usecase, when you may accidently use the wrong relay adress with your auth. 

In my case I learned that this will deny round about 20k spammails from outside, who tried to use my mailcow as "openrelay" because they used a wrong from/rcpt_to address and my mailcow thought..ok I will relay it back to the "primary" spam destination....

## Contribution Guidelines

### Short Description

Add 2 more settings to the smtpd_relay_restrictions key to make sure that no other auth or unwanted unauth relaying will happen under a wrong name/permission.

###  Affected Containers

- postfix-mailcow

## Did you run tests?
Yes

### What did you tested?

I tested from internal/external by script to try to relay to external mailserver an undelivered mail -> denied (expected)
->  Sender address rejected: User unknown in virtual mailbox table; 

The default mailcow installaiton is now not anylonger possible to relay a mail from external to another external mailserver and so it's not an openrelay anymore. With the default config you do not have those 2 settings which mean that you could possible use "any" default mailcow to flood by relaying any other mail instance as you want to flood.
You're also not able to relay a mail from your own mailcow from another "domain" which does not belong to your own mailcow account. You can only relay with your own credentials and your own "domains/accounts".


milkmaker:
Thanks for contributing!

milkmaker:
Thanks for contributing!

DerLinkman:
Violating Contrib Rule 3, please reopen based on staging

Issue #6747, opened 2025-09-14
Google Safe Browsing flags mailcow pages as dangerous


thokich:
### Contribution guidelines

- [x] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [x] ... I understand that not following the below instructions will result in immediate closure and/or deletion of my issue.
- [x] ... I have understood that this bug report is dedicated for bugs, and not for support-related inquiries.
- [x] ... I have understood that answers are voluntary and community-driven, and not commercial support.
- [x] ... I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

### Description

```plain text
Since versions 2025-09a and 2025-09b, Google Safe Browsing has started flagging the
mailcow user domain login page as dangerous and blocks the entire domain.

Previously, version 2025-07 was installed, where this problem did not occur.

Google has now blocked my entire domain for the second time because of the user login page.
```

### Logs:

```plain text
-
```

### Steps to reproduce:

```plain text
Install 2025-09a/b and wait
```

### Which branch are you using?

master

### Which architecture are you using?

x86

### Operating System:

Debian 6.1.147-1 (2025-08-02) 

### Server/VM specifications:

8GB, 4 Cores

### Is Apparmor, SELinux or similar active?

no

### Virtualization technology:

kvm

### Docker version:

Server: Docker Engine - Community  Engine:   Version:          28.4.0

### docker-compose version or docker compose version:

Docker Compose version v2.39.2

### mailcow version:

2025-09b

### Reverse proxy:

no

### Logs of git diff:

```plain text
-
```

### Logs of iptables -L -vn:

```plain text
-
```

### Logs of ip6tables -L -vn:

```plain text
-
```

### Logs of iptables -L -vn -t nat:

```plain text
-
```

### Logs of ip6tables -L -vn -t nat:

```plain text
-
```

### DNS check:

```plain text
-
```

thokich:
<img width="738" height="470" alt="Image" src="https://github.com/user-attachments/assets/41d3e9cc-b1f0-4488-8059-dbae7aa8ad61" />

<img width="735" height="416" alt="Image" src="https://github.com/user-attachments/assets/b2891843-afaf-4083-b5fe-94f06a411bb0" />

thokich:
Hello,
[...]

PR #6758, opened 2025-09-22
[SOGo][Web] SOGo URL Encryption support

FreddleSpl0it:
## Contribution Guidelines

### Short Description

This PR adds **SOGo URL Encryption** support.
[https://www.sogo.nu/files/docs/SOGoInstallationGuide.html](https://www.sogo.nu/files/docs/SOGoInstallationGuide.html#:~:text=auxiliary%20mail%20accounts.-,S,and%20the%20DAV%20url%20will%20change.%20Default%20value%20is%20SOGoSuperSecret0.,-D)

Related:
- https://github.com/mailcow/mailcow-dockerized/issues/6747

###  Affected Containers

- SOGo

### What did you tested?

After a successful login and redirection to SOGo, the user's email address in the URL is encrypted.


docker-compose.yml
```diff
@@ -213,6 +213,7 @@ services:
         - ALLOW_ADMIN_EMAIL_LOGIN=${ALLOW_ADMIN_EMAIL_LOGIN:-n}
         - IPV4_NETWORK=${IPV4_NETWORK:-172.22.1}
         - SOGO_EXPIRE_SESSION=${SOGO_EXPIRE_SESSION:-480}
+        - SOGO_URL_ENCRYPTION_KEY=${SOGO_URL_ENCRYPTION_KEY:-SOGoSuperSecret0}
```
patschi, line 216: IIRC the default `SOGoSuperSecret0` did not work in your tests? Should we maybe pick some other default value, or leave it blank and refuse to start if not appropriately set? Otherwise it might use unsecure keys without anyone noticing.

generate_config.sh
```diff
@@ -436,6 +436,10 @@ MAILDIR_SUB=Maildir
 # SOGo session timeout in minutes
 SOGO_EXPIRE_SESSION=480
 
+# SOGo URL encryption key (exactly 16 characters, limited to A–Z, a–z, 0–9)
+# This key is used to encrypt email addresses within SOGo URLs
+SOGO_URL_ENCRYPTION_KEY=$(LC_ALL=C </dev/urandom tr -dc A-Za-z0-9 2>/dev/null | head -c 16)
```
patschi, line 441: In file https://github.com/mailcow/mailcow-dockerized/blob/ef0f366d1cc9f60a6b17f92390616301b3475535/data/Dockerfiles/sogo/bootstrap-sogo.sh#L31-L32 we seem to recommend using openssl because "seems to hang here occasionally". So if possible we should use same openssl command here too. (Potentially it seeds and generates better, more-random chars)

patschi (commented):
Review test?

docker-compose.yml
FreddleSpl0it: In my tests, I had a leading #, which resulted in 17 chars instead of the required 16 (128 bits).
It must be exactly 16 chars, otherwise it throws errors.

generate_config.sh
FreddleSpl0it: I would use `$(LC_ALL=C </dev/urandom tr -dc A-Za-z0-9 2>/dev/null | head -c 16)` here, since it is already used in several other places during `mailcow.conf` generation. I'm not entirely sure why it was implemented this way, but I would address using `openssl` for this in a separate PR.

docker-compose.yml
patschi: ACK. Makes sense. 

But honestly I'm still not a fan of silently using some hard-coded secret when `mailcow.conf` isn't updated for whatever reasons. Should we maybe add a `if [ "$SOGO_URL_ENCRYPTION_KEY" = "SOGoSuperSecret0"] && echo "Define encryption key. && exit 1` (pseudo-code) during SOGo-startup to make user aware to configure key properly?

FreddleSpl0it: If the Encryption Key is empty, SOGo will also default to `SOGoSuperSecret0`. That's why I also set it in docker-compose.yml
https://www.sogo.nu/files/docs/SOGoInstallationGuide.html#:~:text=auxiliary%20mail%20accounts.-,S,and%20the%20DAV%20url%20will%20change.%20Default%20value%20is%20SOGoSuperSecret0.,-D

patschi: I know. My point is just that we should never fallback to some insecure, hard-coded keys even if SOGo does. I wouldn't see this as best-practise.

Current implementation has the potential of insecure, default key to be used silently. We should error-out if `SOGO_URL_ENCRYPTION_KEY` not specified.

PR #6778, opened 2025-09-25
Add hostname fields to all containers

### LINKED ISSUE #4505: Change Hostname in Container
### Summary

At the moment the hostname in the container are called "localhost"
It it possible to use the serverhostname?

### Motivation

For logging via syslog it would be great, to use the correct hostname and not the name "localhost"

### Additional context

_No response_

patschi:
## Contribution Guidelines

### Short Description

This adds hostnames to all docker containers. It does NOT change the DNS hostnames the services use to communicate between them and are solely for hostname within the containers - as in, they use now static hostnames instead of the docker ID. This might make it easier to identify the container, especially for logging.

Worked fine:
```text
$ for c in $(docker ps -q -a); do docker exec $c hostname; done
watchdog
acme
nginx
ofelia
rspamd
postfix
dovecot
mysql
postfix
clamd
php
redis
sogo
controller
unbound
netfilter
memcached
olefy
```

Closes #4505

###  Affected Containers

All except rspamd. rspamd already had this.

### What did you tested?

Start the stack, check for errors.

No errors found.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,6 +10,7 @@ services:
         - ./data/conf/unbound/unbound.conf:/etc/unbound/unbound.conf:ro,Z
       restart: always
       tty: true
+      hostname: unbound
       networks:
         mailcow-network:
           ipv4_address: ${IPV4_NETWORK:-172.22.1}.254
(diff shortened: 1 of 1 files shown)


patschi:
Haven't seen it being changed, no. Stack stays up and DNS is resolved as before.

Also docker docs only state:
https://docs.docker.com/reference/compose-file/services/#hostname
```
hostname declares a custom host name to use for the service container. It must be a valid RFC 1123 hostname.
```

No mention of any influence of DNS/PTR whatsoever.

PR #6791, opened 2025-09-28
Fix enabling of ipv6 when updating

SpaghettiBorgar:
## Contribution Guidelines

### Short Description
A quick fix in `ipv6_controller.sh` to make sure `/etc/docker` exists before trying to create `daemon.json` as it is not guaranteed to exist by default.

###  Affected Containers
None

### What did you test?
Running `update.sh` with non-existing `/etc/docker` directory and agreeing to create a new configuration file.

Successfully created the config.


patschi (approved):
LGTM.

`dirname` seems to be part of `coreutils`, which is already a required dependency as per our docs.

DerLinkman (approved):

PR #6801, opened 2025-10-01
[Postfix] update postscreen_access.cidr

milkmaker:
This PR updates the postscreen_access.cidr using GitHub Actions and [helper-scripts/update_postscreen_whitelist.sh](https://github.com/mailcow/mailcow-dockerized/blob/master/helper-scripts/update_postscreen_whitelist.sh)


patschi (approved):
LGTM

PR #6867, opened 2025-10-23
[Web] Disable login UI on autoprotocol domains

DiscoNova:
## Contribution Guidelines

### Short Description

This PR adds a new configuration option to disable the login page on autodiscover and autoconfig domains (autodiscover.\*, autoconfig.\*). When enabled, these domains will return a 404 error instead of displaying the login interface for endpoints that haven't been defined elsewhere. This is useful because autodiscover/autoconfig domains are intended solely for automatic client configuration and should not expose the login interface.

For me, disabling the login-interface on the mentioned domains solved the problem I was facing with [issue #6747](https://github.com/mailcow/mailcow-dockerized/issues/6747) ... at least for now.

The implementation includes:
- Backend logic to store and retrieve the configuration setting via Redis
- Checkbox for enabling the option in the admin customization panel
- Login page check that returns 404 when accessing from autodiscover/autoconfig domains if the option is enabled
- Multi-language support (English, German, French, Dutch, Finnish, Swedish)

n.b. While I am a native Finnish/Swedish speaker myself with English as third language, and have studied French and German decades ago, my “Dutch” is actually learnt from Flemish speakers in Belgium, so some TLC review of the translations would be highly appreciated to avoid unintended idiosyncrasies and achieving technical consistency.

###  Affected Containers

- php-fpm-mailcow (PHP-FPM container)

## Did you run tests?

I tested the feature manually, yes.

### What did you tested?

Following components were tested:
- Admin panel UI: Verified the new checkbox appears correctly in the customization settings
- Configuration storage: Confirmed the setting is properly saved to and retrieved from Redis
- Login page behavior: Tested accessing the login page via autodiscover.* and autoconfig.* domains with the option both enabled and disabled
- Language translations: Verified all language strings display correctly in the UI

**Expected behavior:**
- When the option is disabled (default): Login page displays normally on all domains including autodiscover/autoconfig
- When the option is enabled: Login page returns 404 error on autodiscover/autoconfig domains, but works normally on other domains

**Actual results:**
All tests passed and the feature works as expected:
- The checkbox appears in the admin panel under customization settings
- The setting persists correctly in Redis
- When enabled, autodiscover.example.com and autoconfig.example.com return HTTP 404 (for root endpoint access)
- When disabled or on regular (non-autoprotocol) domains, the login page displays normally
- All language translations render properly in the interface

### DIFF
--- a/data/web/index.php
+++ b/data/web/index.php
@@ -27,6 +27,12 @@ elseif (isset($_SESSION['mailcow_cc_role']) && $_SESSION['mailcow_cc_role'] == '
   exit();
 }
 
+$host = strtolower($_SERVER['HTTP_HOST'] ?? '');
+if (str_starts_with($host, 'autodiscover.') || str_starts_with($host, 'autoconfig.')) {
+  http_response_code(404);
+  exit();
+}
+
 require_once $_SERVER['DOCUMENT_ROOT'] . '/inc/header.inc.php';
 $_SESSION['return_to'] = $_SERVER['REQUEST_URI'];
 $_SESSION['index_query_string'] = $_SERVER['QUERY_STRING'];


FreddleSpl0it:
Thank you for the PR.
I think there is no need to make it customizable. We should always return 404.
@patschi What do you think?

DiscoNova:
> I think there is no need to make it customizable. We should always return 404.

That would indeed simplify the implementation quite a bit; just leaving a few of the lines in `index.php` and dropping everything else that is maybe just adding unnecessary complexity ... the reason I made this optional is rather that since autodiscover/autoconfig aren't really standards as such, but rather proprietary protocols, I haven't found an actual "SHALL NOT"-guidelines _forbidding_ using the domains for anything else for either of them - it is rather a community consensus that the domains should only be used for providing configuration data :-/

I also didn't want to make this into a breaking change which some users might - for some reason - currently be relying on. Those are always a pain to figure out when you upgrade a system and something you've grown to using (whether advisable or not) just stops working without a warning.

patschi:
`str_starts_with` sounds good indeed - great idea. 

I'd say rebase & force-push - to keep the discussion in this PR too. Should do the trick.

DiscoNova:
I've removed all the unnecessary changes that enabled making this feature optional (UI, localisations, backend storage) and left the essentials.

At one point I noticed diving deep into professional paranoia rabbithole trying to handle various imaginary edge cases that are very unlikely to ever be seen in real life and decided I was overthinking and should take a step back :D 

The current version of the code makes one memory allocation for the lowercase hostname, which could be avoided at the cost of increased code complexity - but I think this implementation with a much more straight-forward approach is a good compromise between performance and future maintainability.

During my latest testing I did notice that the current server implementation actually responds to both protocol requests on both domains (i.e. you can request autoconfig-endpoints using autodiscovery-domain and vice versa) but that issue is way beyond the scope of this PR ... and probably of little consequence one way or the other.

DiscoNova:
I understand this PR has only been open a few weeks (sometimes I've waited months for review etc. - often enough to completely forget what I was thinking about when submitting one:) but I wanted to ask if there is still something you folks would want changed / addressed in this pull request context? @FreddleSpl0it @patschi

FreddleSpl0it:
@DiscoNova Looks good to me. I don’t have time for development right now, but I’ll likely get back to your PR in late December.

FreddleSpl0it (approved):

Issue #304, opened 2017-05-21
End users creating alias


cuddylier:
At present it doesn't seem possible for end users to create permanent alias in the end user panel, only in the admin panel. It would be nice if end users could create alias themselves up to the limit set for their domain.

stevesbrain:
If implemented, I feel it should be a per-domain settings as to whether this is enabled for users, or just domain administrators. I imagine certain domains would want to be restricted so additional aliases cannot be created, such that people cannot impersonate others etc (I imagine this would be more relevant when in a small business or semi-public setting of course). Just my $0.02, but I agree, it would be nice to have the choice :)

Quiwy:
Hi, I'm are highly interested in this feature too, it would be tremenendous to see it implemented.

PR #6888, opened 2025-11-01
Allow making spam aliases permanent

PseudoResonance:
## Contribution Guidelines

### Short Description

Allows setting user spam aliases to have no expiry.

image

Translations for the aliases in English, Spanish and Japanese.

###  Affected Containers

- php-fpm
- nginx
- postfix

## Did you run tests?

Yes

### What did you tested?

Tested creating/modifying aliases back and forth between expiry dates and no expiry. Tested automatic deletion of old aliases and ensuring permanent ones are unaffected.

Users can choose to make their spam aliases last indefinitely for use on long-term, but still spammy accounts. Aliases are fully configurable as needed and normal aliases still work as expected.


PseudoResonance:
Related to #3892 and #304.

I realize it might be a bit weird to implement the permanent aliases on the "temporary alias" user page, but it somehow seemed more appropriate to me than doing something else. Naturally I'm happy to reimplement it in a different way if desired.

DerLinkman (approved):
Thank you very much!

Issue #4478, opened 2022-02-18
OAuth2 avoid multiple confirmations when logging in multiple times


BigMichi1:
## Summary

Provide an option for an OAuth2 app to suppress the confirmation screen (or follow up confirmations for some days/forever) when an external app is performing a login in via oauth. what is quite useful the first time is annoying after some days especially when you are logging in multiple times per day

## Motivation

What are you about to solve or improve with this idea?
What would be the benefit for most users?
user will not be prompted again and again to authorize the login

## Additional context

add the Trusted checkbox to the current screen for an OAuth app similar to the one from GitLab
current screen in MailCow
![Bildschirmfoto vom 2022-02-18 16-15-11](https://user-images.githubusercontent.com/91786/154710186-c58ffc58-ef80-4df3-938e-41192127f5b6.png)

Example from GitLab where only the first login needs to be confirmed because of the Trusted checkbox
![Bildschirmfoto vom 2022-02-18 16-16-16](https://user-images.githubusercontent.com/91786/154710179-10600ea2-e7ac-40c0-b4af-d7feb28318ef.png)

waschbaer77:
We have the same "problem". It's annoying. This enhancement would help a lot.

TaylorChristie:
The normal Oauth2 flow should require authorization for the first installation, then afterwards no longer request it. Having the user authorize every time is annoying and breaks the continuity of our integrations. +1 for this issue.

Gri89:
I confirm, it would be very convenient.
I have authorization set up in several services, and to enter each one, login confirmation is required every time, which does not take much time, but is very annoying.

mfuryto:
I agree. This is an important enhancement when running several services and users

Fly7113:
I too second this enhancement. An option to configure this behaviour is mandatory.

lauff:
Wondering that this obvious requirement is still open for more than two years. Unfortunately, I am not so familiar with GitHub processes. Is anyone working on this?

Anyhow, thanks for the great work!

Warm regards, Markus

mfuryto:
I love all the good work you guys are doing. I wish I had the skills to contribute with this myself. Im doing a bump on this enhancement to get new attention to it because I know this would be a great fix for all mailcow admins not using an external IDP.

I think there could be two solutions:
1: There should be stored when the user click authorize, and then the flow would check before opening the authorize screen if the authorization is stored. If stored it bypass the authorization screen.
[...]

PR #6895, opened 2025-11-04
feat: Add support for trusted OAuth2 clients that skip the consent screen.

h44z:
## Contribution Guidelines

### Short Description

This PR adds support for marking OAuth2 clients as trusted. Trusted clients automatically skip the user consent screen during authorization, simplifying authentication flows for internal or pre-approved applications.

This is particularly useful when using Mailcow as an OAuth2 server for other applications such as Nextcloud or Seafile.
The feature request #4478 should also be solved by this PR.

###  Affected Containers

The change only affects the Mailcow web portal code and does not directly impact any containers.
To test the changes, overwrite the files in data/web/... and restart the php-fpm container.

## Did you run tests?

Yes, I am using the proposed changes on my Mailcow instance in production.

### What did you test?

 - I verified backward compatibility with existing Mailcow installations that do not include the new feature yet. For updated instances, OAuth2 clients remain untrusted by default, and the login and consent behavior is unchanged.
 - I tested both trusted and untrusted OAuth2 clients. For untrusted clients, the OAuth2 consent page continues to appear as expected. For trusted clients, users are redirected directly to the callback URL after logging in to Mailcow - no consent screen is shown in this case.
 - I validated the UI elements and translations (German and English). Creating and modifying clients works as intended, and switching a client between trusted and untrusted states functions seamlessly.

The Mailcow UI for managing OAuth2 clients now includes a 'Trusted' flag, allowing designated clients to skip the consent screen. The underlying database table for OAuth2 clients has been updated to include a new column for storing the trusted status. Existing OAuth2 clients remain untrusted and the consent behavior is unchanged.

Screenshots:
image
image
image


jakob-info:
I like this feature 👍

mfuryto:
Still waiting, using stable release :)

data/web/oauth/authorize.php
```diff
@@ -25,7 +29,7 @@
 }
 
 // print the authorization code if the user has authorized your client
-$is_authorized = ($_POST['authorized'] == '1');
+$is_authorized = ($trusted_client == 1 || $_POST['authorized'] == '1');
```
neorejalist:
```suggestion
$is_authorized = ($trusted_client == 1 || ($_POST['authorized'] ?? '') == '1');
```

To prevent `E_NOTICE ("Undefined index: authorized") in the logs`

h44z:
@DerLinkman @FreddleSpl0it any thoughts on this?

MMilesMM:
I would also really appreciate this feature :)

milkmaker:

h44z:
@DerLinkman @FreddleSpl0it can you maybe review this PR? This PR would be really helpful to avoid having to re-confirm the OAuth consent every time you log in...

mfuryto:
Would it make sense to keep trusted clients separate from remembered user consent?

A trusted client would be pre-approved by an administrator, while a normal client could remember consent per user and scope after the first approval.

Issue #7183 suggested using access tokens for this, but that would lose consent when tokens expire or are cleaned up. A separate persistent consent record may be safer.

Would you prefer discussing this here or in a new issue?

jakob-info:
In my opinion, the trusted flag is sufficient for now.

As far as I can tell, Mailcow's own OAuth only supports the profile scope, which includes your full name (if provided) and email address. The other fields, such as username and identifier, reuse the email address. Normal users expect an application they sign into via OAuth to receive their name and email address. Moreover, they can only sign into applications that their local admin has explicitly set up for them, such as their Nextcloud or WordPress instance. Therefore, separate consent for each user isn't needed.

mfuryto (commented):
I reviewed the current diff. Overall, the approach looks good and the change is nicely contained.

What works well:

- Existing clients remain untrusted by default.
- The authorization request is validated before the trusted-client check.
- Trusted clients only skip the consent screen; the existing authorization-code and token handling remains unchanged.
- The setting is supported when creating and editing clients.
- The earlier undefined `authorized` notice has been fixed.

I found two small code cleanups:

1. The submitted `trusted` value should be normalized before it is stored:

```php
$trusted = in_array($_data['trusted'] ?? null, [1, '1'], true) ? 1 : 0;
```

This should be used in both the add and edit paths.

2. `details_by_client_id` appears to be unused because `oauth2trusted()` performs its own database lookup. The unused action can be removed.

The branch also needs to be rebased onto the current staging branch. Staging already has a newer database schema version, which is causing the merge conflict. The schema version should be updated after the rebase.

A short test log covering trusted and untrusted clients, disabling trust again, and the default for existing clients would be useful.

The implementation otherwise looks reasonable to me.

PR #6934, opened 2025-12-01
Domain Wide Footer no longer breaks MIME or loses Attachments

shaneodoo:
When an email contains attachments or embedded images, it's structured as a multipart MIME message with special boundary markers that separate each part, the text body, HTML body, attachments, and images all sit between these boundaries. 

The rspamd.local.lua script that added the system footer,  wasn't properly detecting these multipart structures, so when it injected the footer, it would inadvertently break or modify these MIME boundaries also disrupting special characters like German Umlauts . Email clients couldn't parse the broken structure correctly, which resulted in attachments simply disappearing from the recipient's view or corrupting the MIME structure and falsely displaying incorrect characters.

This fix involves adding smart detection logic that identifies when an email is multipart before attempting to add the footer. For simple plain text or HTML-only emails, the code operates normally, adding the footer and adjusting encoding as needed. But for complex multipart messages like replying to or forwards with attachments, the code now takes a hands-off approach, it preserves the original MIME structure completely, lets rspamd's built-in footer function handle the insertion, and avoids touching any encoding headers or boundary markers. This way, the footer gets added without disrupting the carefully structured MIME format that email clients rely on to display attachments correctly.

## Contribution Guidelines

## What does this PR include?

rspamd.local.lua

### Short Description

Fixes the MIME structure so that when an attachment is added and the domain wide footer is enabled Rspamd does not mess with the MIME structure to include it. This mainly affects Outlook 2010-2019, but has been known to also affect Outlook 364 and New Outlook, and some other email clients like Mailbird. The corruption drops attachments when replies are sent or forwarded. 

###  Affected Containers

rspamed

### What did you tested?

Applied new rspamd.local.lua to data/conf/rspamd/lua folder, restarted container and ensured the logs had no issues. 

Testing original, replies and forward messages using multiple email clients confirms the fix works perfectly. Sent emails with multiple attachments now arrive with all files intact and the footer properly appended. Embedded images remain visible, and complex email threads with mixed content types maintain their structure. The solution is elegant in its simplicity, rather than trying to reconstruct MIME boundaries The added logic to detect and preserve what's already there has fixed the issue.

Complete success. The fix resolved both issues, character encoding for special characters and MIME boundary preservation for complex multipart messages. All test scenarios passed without any edge case failures.

### DIFF
--- a/data/conf/rspamd/lua/rspamd.local.lua
+++ b/data/conf/rspamd/lua/rspamd.local.lua
@@ -1,10 +1,10 @@
 rspamd_config.MAILCOW_AUTH = {
-	callback = function(task)
-		local uname = task:get_user()
-		if uname then
-			return 1
-		end
-	end
+  callback = function(task)
+    local uname = task:get_user()
+    if uname then
+      return 1
+    end
+  end
 }
 
 local monitoring_hosts = rspamd_config:add_map{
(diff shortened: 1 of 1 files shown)


shaneodoo:
Is someone able to progress this please? @DerLinkman @dragoangel

shaneodoo:
When can I expect to see progress with this. I have been running this update on my email server now for 2 months without issues.

dragoangel (changes requested):
Do not delete comments, not change formating without reason, please leave in PR only changes to implement your fix

milkmaker:

PR #6988, opened 2026-01-01
Prevent fail2ban from adding multiple netfilter rules for the same IP

furest:
## Contribution Guidelines

* [X ] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree them

### Short Description

An MUA makes multiple simultaneous failed authentication attempts to mailcow, this can result in the creation of more than one netfilter ban rule.
When one of the simultaneous attempts triggers an actual ban, the following attempts will not be prevented, as fail2ban does not instantly kill the established sessions.
The result is that fail2ban will still process the subsequent log entries, and because the attempt number stays higher than the treshold, it will create additional (duplicate) netfilter ban rules.
However, when unbanning the IP, the fail2ban daemon only unbans the IP address once, leaving behind extra netfilter rules that will never get deleted.

###  Affected Containers
- netfilter

### What did you tested?
- Using swaks to spam my mailcow instance with failed authentication attempts. Without the patch, 15 attempts will result in the creation of 6 netfilter ban rules. With this patch, only 1 rule is created.
- Testing the created functions independently. When IP w.x.y.z is banned, "isIPv4Banned" returns True. When the IP is not already banned, it will return False.
- Tested with both IPv4 and IPv6.

Awaited : More than 10 failed authentication attempts should result in the creation of a single nftables ban rule
Got : More than 10 failed authentication attempts resulted in the creation of a single nftables ban rule.

### Additional comment 
Since IPTables is going towards deprecation I don't see any reason to port this fix to IPTables.
I have nonetheless added "dummy" functions in the IPTables class to prevent any crash caused from calling non-existent functions.
If supporting IPTables is a strict requirement for this PR to be merged, please let me know.

### DIFF
--- a/data/Dockerfiles/netfilter/main.py
+++ b/data/Dockerfiles/netfilter/main.py
@@ -167,12 +167,18 @@ def ban(address):
     logger.logCrit('Banning %s for %d minutes' % (net, NET_BAN_TIME / 60 ))
     if type(ip) is ipaddress.IPv4Address and int(f2boptions['manage_external']) != 1:
       with lock:
-        logdebug("Calling tables.banIPv4(%s)" % net)
-        tables.banIPv4(net)
+        if not tables.isIPv4Banned(net):
+            logdebug("Calling tables.banIPv4(%s)" % net)
+            tables.banIPv4(net)
+        else:
+            logdebug("IPv4 %s is already banned." % net)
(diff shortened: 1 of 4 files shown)


milkmaker:

PR #7021, opened 2026-01-27
[Postfix] Configurable send permissions for alias addresses

FreddleSpl0it:
## Contribution Guidelines

### Short Description

This PR introduces a feature that gives admins control over whether alias addresses are permitted to send emails.

###  Affected Containers

- postfix-mailcow
- phpfpm-mailcow


miszterx:
May I ask regarding this PR?

I have installed this update, because i was really interested this new feature. 

I have an alias like:

group1@email.com
and there are 5 email inside the goto address list. Even inside this mailcow and outside the mailcow as well.

What I expected I can define a list of addresses who are allowed to send email to this alias.

Since I have installed this update I can see a tick option under the alias:

"If disabled, this alias can only receive mail. Use sender ACL to override and grant specific mailboxes permission to send."

So this new PR only in control of alias which mailboxes can use to send email and nothing more?

Issue #6998, opened 2026-01-06
MTA-STS Connection refused


BEBU88:
### Contribution guidelines

- [x] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### Checklist prior issue creation

- [x] I understand that failure to follow below instructions may cause this issue to be closed.
- [x] I understand that vague, incomplete or inaccurate information may cause this issue to be closed.
- [x] I understand that this form is intended solely for reporting software bugs and not for support-related inquiries.
- [x] I understand that all responses are voluntary and community-driven, and do not constitute commercial support.
- [x] I confirm that I have reviewed previous [issues](https://github.com/mailcow/mailcow-dockerized/issues) to ensure this matter has not already been addressed.
- [x] I confirm that my environment meets all [prerequisite requirements](https://docs.mailcow.email/getstarted/prerequisite-system/) as specified in the official documentation.

### Description

I'm running my mailcow UI at a different port. Therefore I've changed HTTPS_PORT in mailcow.conf. The connection for https://mta-sts.example.com/.well-known/mta-sts.txt on port 443 got refused. I switched the port to 443 for testing purposes and it worked as expected.

Unfortunately the error wasn‘t fixed with #6739.

In my opinion mailcow should always serve the mta-sts at https://mta-sts.example.com/.well-known/mta-sts.txt at Port 443 if enabled. Connections to the mailcow domain for the Ul should only be served at the HTTPS_PORT (actual behavior).

### Steps to reproduce:

1. Setup MTA-STS in mailcow
2. change HTTPS_PORT to something other than 443
3. Connection for mta-sts.example.com is refused
4. change HTTPS_PORT back to 443
5. MTA-STS is working as expected

### Logs:

```plain text
-
```

### Which branch are you using?

master (stable)

### Which architecture are you using?

x86_64

### Operating System:

Debian 13

### Server/VM specifications:

4GB 4 Core

### Is Apparmor, SELinux or similar active?

No

### Virtualization technology:

KVM

### Docker version:

-

### docker-compose version or docker compose version:

-

### mailcow version:

2025-12a

### Reverse proxy:

Buildin mailcow Proxy

### Logs of git diff:

```plain text

```

### Logs of iptables -L -vn:

```plain text
-
```

### Logs of ip6tables -L -vn:

```plain text
-
```

### Logs of iptables -L -vn -t nat:

```plain text
-
[...]

PR #7080, opened 2026-02-24
mta-sts.txt working with and without reverse proxy

monster010:
## Contribution Guidelines

### Short Description

#6998 solves the problem by adding a simple function that takes reverse proxy into account.

###  Affected Containers

- nginx-mailcow
- php-fpm-mailcow

### What did you tested?

- Accessing mta-sts.txt with reverse proxy
- Accessing mta-sts.txt without reverse proxy

I expected mta-sts.txt to work both with and without a reverse proxy.


flightlesstux:
Hey, thanks for the work on this — the idea of supporting reverse proxy setups via `X-Forwarded-Host` is great and definitely needed.

Just spotted a logic bug in the `get_requested_host()` function that will unfortunately cause the reverse proxy path to never execute:

```php
// Current (buggy) — condition is always false
if(!isset($_SERVER['HTTP_X_FORWARDED_HOST']) && !empty($_SERVER['HTTP_X_FORWARDED_HOST'])) {
```

`!isset` and `!empty` are contradictory here — if the header isn't set, it also can't be non-empty, so this block is unreachable. The `!isset` should be `isset`:

```php
// Fixed
if(isset($_SERVER['HTTP_X_FORWARDED_HOST']) && !empty($_SERVER['HTTP_X_FORWARDED_HOST'])) {
    $xfh = explode(',', $_SERVER['HTTP_X_FORWARDED_HOST'])[0];
    return trim($xfh);
}
```

Once that's flipped, the logic works as intended — falls back to `HTTP_HOST` when the forwarded header isn't present, and uses the first value from the comma-separated list when it is. 

One small note on security: trusting `X-Forwarded-Host` directly is fine here since this is only used to check the `mta-sts.` prefix and look up a domain that's already validated against the database downstream — no injection risk.

Great fix overall, just needs that one-character change!

monster010:
@flightlesstux Thanks for the heads-up, I must have somehow overlooked that.
Fixed.

milkmaker:

monster010:
I am surprised that pull requests submitted by the community months ago—which solve existing problems—have not yet been merged.

PR #7086, opened 2026-03-03
[SOGo] Build SOGo from source with security patches

FreddleSpl0it:
## Contribution Guidelines

### Short Description

This PR modifies the SOGo Docker image to build from source instead of using pre-compiled packages from Alinto's repository. The primary motivation is that Alinto does not include security patches in their official packages in a timely manner, leaving systems vulnerable to known security issues.

The implementation uses a source-based build with git cherry-pick support to apply critical security patches before official releases.

**Key changes:**
- Single-stage Dockerfile building
- Git cherry-pick mechanism for applying security patches to stable releases

**Benefits:**
- Apply security patches immediately without waiting for official packages
- Easy addition of future patches via `SOGO_SECURITY_PATCHES` ARG

###  Affected Containers

- sogo-mailcow

### What did you tested?

- Docker image builds successfully from source
- SOGo daemon starts and runs properly

### DIFF
--- a/data/Dockerfiles/sogo/Dockerfile
+++ b/data/Dockerfiles/sogo/Dockerfile
@@ -1,47 +1,128 @@
-FROM debian:bookworm-slim
+# SOGo built from source to enable security patch application
+# Repository: https://github.com/Alinto/sogo
+# Version: SOGo-5.12.4
+#
+# Applied security patches:
+# - 16ab99e7cf8db2c30b211f0d5e338d7f9e3a9efb: XSS vulnerability in theme parameter
+#
+# To add new patches, modify SOGO_SECURITY_PATCHES ARG below with space-separated commit hashes
+
+FROM debian:bookworm
 
 LABEL maintainer="The Infrastructure Company GmbH <info@servercow.de>"
 
 ARG DEBIAN_FRONTEND=noninteractive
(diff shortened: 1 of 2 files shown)


data/Dockerfiles/sogo/Dockerfile
```diff
@@ -1,47 +1,161 @@
...
-  supervisor \
-  syslog-ng \
-  syslog-ng-core \
-  syslog-ng-mod-redis \
-  dirmngr \
-  netcat-traditional \
-  psmisc \
-  wget \
-  patch \
+# Install dependencies, build SOPE and SOGo, then clean up (all in one layer to minimize image size)
```
Daniel15, line 23: The issue with doing it all in one layer is that your cache efficiency when rebuilding goes way down. You could use a multi-stage build to use multiple layers while building, without bloating the final image size with those layers. https://docs.docker.com/build/building/multi-stage/

FreddleSpl0it: then give it a try. I couldn't get Sogo to work in a multi-stage build because you have to copy so many library files and there's nothing documented.

Daniel15: Thanks for the reply! I'll give it a try if I get some free time.

Issue #7102, opened 2026-03-10
mysql/mariadb not healthy after upgrade to 2026-03


mjohnsonsu:
### Contribution guidelines

- [x] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### Checklist prior issue creation

- [x] I understand that failure to follow below instructions may cause this issue to be closed.
- [x] I understand that vague, incomplete or inaccurate information may cause this issue to be closed.
- [x] I understand that this form is intended solely for reporting software bugs and not for support-related inquiries.
- [x] I understand that all responses are voluntary and community-driven, and do not constitute commercial support.
- [x] I confirm that I have reviewed previous [issues](https://github.com/mailcow/mailcow-dockerized/issues) to ensure this matter has not already been addressed.
- [x] I confirm that my environment meets all [prerequisite requirements](https://docs.mailcow.email/getstarted/prerequisite-system/) as specified in the official documentation.

### Description

Watchdog is reporting errors with mysql after upgrade:
QUERY CRITICAL: TLS/SSL error: SSL is required, but the server does not support it
TLS/SSL error: SSL is required, but the server does not support it
QUERY CRITICAL: TLS/SSL error: SSL is required, but the server does not support it
TLS/SSL error: SSL is required, but the server does not support it
QUERY CRITICAL: TLS/SSL error: SSL is required, but the server does not support it
TLS/SSL error: SSL is required, but the server does not support it
QUERY CRITICAL: TLS/SSL error: SSL is required, but the server does not support it
TLS/SSL error: SSL is required, but the server does not support it
QUERY CRITICAL: TLS/SSL error: SSL is required, but the server does not support it
TLS/SSL error: SSL is required, but the server does not support it
QUERY CRITICAL: TLS/SSL error: SSL is required, but the server does not support it
Tue Mar 10 09:04:35 CET 2026 Container is running for less than 360 seconds, skipping action...
Tue Mar 10 09:04:35 CET 2026 MySQL/MariaDB health level: 20% (1/5), health trend: -4

### Steps to reproduce:

1. Update to 2026-03

### Logs:

```plain text
QUERY CRITICAL: TLS/SSL error: SSL is required, but the server does not support it
TLS/SSL error: SSL is required, but the server does not support it
QUERY CRITICAL: TLS/SSL error: SSL is required, but the server does not support it
[...]

PR #7103, opened 2026-03-10
[Watchdog] Fix Nagios MariaDB client SSL compatibility with Alpine 3.23

FreddleSpl0it:
## Contribution Guidelines

### Short Description

This PR fixes MySQL/MariaDB SSL compatibility issues introduced with Alpine 3.23, which ships MariaDB client 11.4.9. The new client version enforces SSL by default, causing watchdog health checks to fail with "TLS/SSL error: SSL is required, but the server does not support it" even when using Unix socket connections.

The fix adds a MySQL client configuration file for Nagios Plugins that disables SSL for Unix socket connections and updates all Nagios plugin calls to use this configuration.

Fixes:
- #7102
###  Affected Containers

- watchdog-mailcow


aminvakil:
Do you consider this serious enough to issue a release or should I just cherry-pick it?

PR #7140, opened 2026-03-13
fix(ofelia): pin ofelia image to v0.3.21 instead of :latest

flightlesstux:
## Summary

- `mcuadros/ofelia:latest` is the only unpinned image in `docker-compose.yml` — every other service uses an explicit version tag
- Using `:latest` is a supply chain risk: the image silently changes on every `docker compose pull`, making deployments non-reproducible and exposing the stack to unexpected upstream changes
- Pinned to `v0.3.21` (current latest release)

## Test plan

- [ ] Verify `docker compose pull ofelia-mailcow` pulls `mcuadros/ofelia:v0.3.21`
- [ ] Confirm ofelia-mailcow starts and scheduled jobs run as expected after the change


patschi:
Wrong branch, low-effort straight from Claude - even down to the individual comments in linked issue. Even though the PR would probably be valid, I don't want to support this behavior. Closing.

Issue #6974, opened 2025-12-16
Add SCIM provisioning


Fly7113:
### Summary

A nice feature to have, following the relatively recent introduction of SSO via OIDC/Keycloak/LDAP, would be a user provisioning feature to pair alongside the OIDC identity provider.

### Motivation

SCIM is a well established and thoroughly tested standard, that, along other things, would ease the load on the mailcow infrastructure since the periodic syncs are run by the identity provider, not requiring neither a cronjob on mailcow's side nor admin access to the Keycloak Admin REST API.
Another plus is that the configuration would be simpler in the case of Keycloak, since it wouldn't require a service account on the IdP side, and potentially the same for the generic OIDC backend, since it would be optional and on the mailcow's side it would mean to generate just an authorisation token for the SCIM API during the setup process.

Fly7113:
If it is helpful, I could attempt to extend the API to implement the SCIM endpoints.
I am unable to provide further assistance as I lack the confidence and knowledge to implement the internal logic for the authentication infrastructure.

Fly7113:
In the past months I found the time to dive deep into the topic.
The result of my work is found in #7148

PR #7148, opened 2026-03-17
[Web] Add SCIM 2.0 provider for IdP user provisioning

Fly7113:
## Contribution Guidelines

### Short Description

Implements a SCIM 2.0 (RFC 7643/7644) server endpoint so any Identity Provider — Keycloak, Entra ID, Okta, or any LDAP/OIDC IdP — can push user lifecycle events to mailcow in real time, independently of whichever login protocol is configured.

#### Protocol support

- Full Users CRUD: POST, GET, PUT, PATCH (RFC 7644 §3.5.2), DELETE
- SCIM DELETE is a soft-deactivate (active=0); mail data is never removed
- Filtering: filter=userName eq "..." on list endpoint
- Pagination: startIndex / count
- Discovery: ServiceProviderConfig, Schemas, ResourceTypes
- Groups: out of scope

#### Authentication & token management

Bearer tokens are generated in the admin UI (System > Configuration > Access > SCIM). The raw token is shown once at creation and never stored; only its SHA-256 hash is kept. Each token supports:
- Optional domain restriction (limits which mailboxes the token can manage)
- Optional mailbox template (applied on user creation)
- Optional IP allow-list / skip-IP-check flag
- Active/inactive toggle

#### Database schema

Two new tables added via the existing init_db migration mechanism:
- scim_tokens: stores token metadata and hashed credentials
- scim_maps: maps IdP externalId values to mailcow usernames per token The mailbox.authsource ENUM is extended with 'scim'.

#### Authsource & login design

SCIM is a provisioning protocol, not an authentication protocol. mailbox.authsource='scim' records who manages the user; login is handled by the globally configured IAM provider:

- Keycloak / Generic-OIDC: SCIM users pass through the existing verify-sso OIDC flow (identity_provider 'verify-sso' case).
- LDAP: SCIM users authenticate via ldap_mbox_login(), with full TFA support, matching the behaviour of authsource='ldap' users.
- No IAM configured: SCIM users cannot log in; the admin UI shows a warning on the SCIM configuration tab.

Attempting a password login as a SCIM user when an OIDC provider is configured returns a clear error directing the user to their IdP.

#### Claiming pre-existing users

A SCIM POST for a user who already has authsource='scim' (e.g. set manually by the admin to prepare a migration) is treated as a claim: attributes are updated, scim_maps is upserted, and 200 is returned. A SCIM POST for a user managed by a different authsource returns 409 with an actionable message explaining how to transfer ownership.

#### Admin UI

- New SCIM tab under System > Configuration > Access (alongside Identity Provider settings)
- Token table with active toggle and delete; one-time raw token modal
- Mailbox edit form gains a SCIM authsource option, shown only when SCIM tokens exist (or the mailbox is already set to SCIM)
- Contextual warning when no external IdP is configured for login

###  Affected Containers

- php-fpm-mailcow
- nginx-mailcow (nginx configuration template)
- sogo-mailcow (shared DB initialisation file)

### What did you tested?

- User creation, updating, deactivation via SCIM
- General SCIM protocol testing
- Migration between mailcow/external IdP and SCIM
- Logging-in in various scenarios
- Invalid SCIM requests (updating non-SCIM users to test the error handling)
- Admin UI test (create/update/delete SCIM tokens)

All the tests gave results conforming to the described behaviours.

### DIFF
--- a/data/conf/nginx/templates/sites-default.conf.j2
+++ b/data/conf/nginx/templates/sites-default.conf.j2
@@ -104,6 +104,10 @@ location ~ ^/api/v1/(.*)$ {
     try_files $uri $uri/ /json_api.php?query=$1&$args;
 }
 
+location ~ ^/scim/v2/(.*)$ {
+    try_files $uri $uri/ /scim.php?path=$1&$args;
+}
+
 location ~ ^/cache/(.*)$ {
     try_files $uri $uri/ /resource.php?file=$1;
 }
--- a/data/conf/nginx/templates/sites-default.conf.j2
+++ b/data/conf/nginx/templates/sites-default.conf.j2
@@ -104,7 +104,7 @@ location ~ ^/api/v1/(.*)$ {
     try_files $uri $uri/ /json_api.php?query=$1&$args;
 }
 
-location ~ ^/scim/v2/(.*)$ {
+location ~ ^/scim/v2(?:/(.*))?$ {
     try_files $uri $uri/ /scim.php?path=$1&$args;
 }
 
--- a/data/web/admin/system.php
+++ b/data/web/admin/system.php
@@ -86,6 +86,12 @@ $cors_settings['allowed_methods'] = explode(", ", $cors_settings['allowed_method
 $f2b_data = fail2ban('get');
 // mbox templates
 $mbox_templates = mailbox('get', 'mailbox_templates');
+// SCIM
+require_once $_SERVER['DOCUMENT_ROOT'] . '/inc/functions.scim.inc.php';
+$scim_tokens    = scim_token('get_all');
+$scim_new_token = $_SESSION['scim_new_token'] ?? null;
+$scim_base_url  = 'https://' . getenv('MAILCOW_HOSTNAME') . '/scim/v2/';
+unset($_SESSION['scim_new_token']);
 
 $template = 'admin.twig';
 $template_data = [
@@ -121,6 +127,9 @@ $template_data = [
   'is_https' => isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] === 'on',
   'iam_settings' => $iam_settings,
   'mbox_templates' => $mbox_templates,
+  'scim_tokens' => $scim_tokens,
+  'scim_new_token' => $scim_new_token,
+  'scim_base_url' => $scim_base_url,
   'lang_admin' => json_encode($lang['admin']),
   'lang_datatables' => json_encode($lang['datatables'])
 ];
--- a/data/web/admin/system.php
+++ b/data/web/admin/system.php
@@ -87,7 +87,6 @@ $f2b_data = fail2ban('get');
 // mbox templates
 $mbox_templates = mailbox('get', 'mailbox_templates');
 // SCIM
-require_once $_SERVER['DOCUMENT_ROOT'] . '/inc/functions.scim.inc.php';
 $scim_tokens    = scim_token('get_all');
 $scim_new_token = $_SESSION['scim_new_token'] ?? null;
 $scim_base_url  = getBaseUrl() . '/scim/v2/';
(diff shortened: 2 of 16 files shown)


Fly7113:
Fixes #6974

Fly7113:
Just a small testing update: in these past few days I had the chance to test this PR on my production system, alongside Authentik as SCIM and OIDC IdP. As far as I'm concerned all the features are working as expected, even in a fully production system having users with mixed `authsources`.

The are only two things of which I'm not fully convinced:
1. The deactivation of the user when a SCIM `DELETE` request is received: the user can be already just deactivated leveraging the `active` property of the user's JSON, so that scenario is already covered, and the user should be safe to be deleted on a DELETE request. From a spec compliance point of view the current implementation is not fully compliant because RFC 7644 states that a system MAY choose not to delete the specified resource, but such resource, in the case of other SCIM operations, should then be treated as if it was actually deleted. On the other hand a true deletion of the user on a DELETE request would be easier to implement with a perfect compliance.
2. As of now, the deactivation of the user via the `active` JSON parameter fully deactivates the user, but another alternative would be to put the user in the "soft" deactivation status that permits the user to still receive emails. I don't think this choice is relative to compliance to the RFCs, therefore an option would be to give both alternatives to the system admin and making the effect configurable from the admin UI.

Before moving forward with a decision I'd like to hear the consensus of the maintainers team.

Fly7113:
Sorry for pushing these new quite conspicuous two commits, but in these further days of testing I realised that the UI (and part of the logic about the IP check) was badly done, so I decided to work on it.
Now the SCIM tokens are modifiable, and the template mapping uses the same working principles of the ones in the OIDC template mapping, to have more consistency.
The input fields are now more robust (no more free text basically, just dropdown menus), and I added the strings in the lang.en-gb JSON which I had previously forgot (ups :D).

milkmaker:

Fly7113:
Definitely not to be closed.
@FreddleSpl0it (sorry if you are not the correct person to tag) could I get some feedback from the mailcow team? I am still waiting for a consensus on your end about the issues mentioned in my previous messages.
FYI in the past nearly *three months* I had the chance to use and test this feature in production without any issue at all.

milkmaker:

Fly7113:
This is getting kind of embarrassing… 😅
@MAGICCC @FreddleSpl0it is there even the slightest possibility of eventually seeing this merged?
As soon as I get the required feedback, I’ll rebase and implement.

Fly7113:
In the mean time, we onboarded in this PR another partner, that uses a different IdP.
We found a few issues here and there, so we ran a full audit of the implementation against RFC 7643 and 7644 and fixed everything it turned up.

One breaking change: the SCIM id is no longer the mailbox address but an opaque UUID stored in scim_maps, as RFC 7643 §3.1 requires an immutable server-assigned identifier. IdPs that cached the old ids will 404 on their next update and re-match; existing test installs need a re-import cycle.

A few highlights of the changes:
- full RFC 7644 filter grammar (the old single-regex filter made the advertised filter.supported=true untrue for everything else, including the externalId lookups Okta reconciles with);
- PATCH now supports valuePath paths, merges name components correctly and is atomic;
- Location on 201 and WWW-Authenticate on 401;
- individual /Schemas/{id} and /ResourceTypes/{id}; /.search; attributes/excludedAttributes;
- complete §7 attribute metadata; and the template selector moved out of the closed IETF enterprise URN into a declared mailcow extension;
- allow_claim is a new per-token flag: adopting a pre-created mailbox on POST is now opt-in, and off by default it returns the spec-mandated 409;
- the /Groups endpoint now replies to a GET with an empty collection, since there are a few IdPs (Authentik, Entra ID) that run a discovery pass without consulting /ResourceTypes (technically doesn't break spec compliance, since no MUSTs are broken)

Two notable fixes worth mentioning:
- every mailbox() call is now checked, so quota/ACL/validation failures return 400 instead of a 200 that the IdP recorded as success;
- boolean parsing is strict, because Entra sends "False" as a string and the previous cast turned it into true, leaving deactivated users able to log in.

The only single deviation is the DELETE action, which currently deactivates instead of destroying the mailbox (mail data is preserved, so the resource stays retrievable — §3.6). Still waiting for your call, and either way some code needs to be modified to have the implementation compliant.

PR #7162, opened 2026-03-23
Fix user webmail SSO entrypoints

zylos-luna:
## Summary

This fixes two related issues in the user webmail SSO flow:

1. The user-facing `Webmail` entrypoint currently links directly to `/SOGo/so`, which bypasses the intended SSO handoff path.
2. `sogo-auth.php` can reject a mailbox user's own SSO login because it only accepts alias-based lookup in paths where the mailbox itself should also be valid.

## Changes

- Change the user webmail link target to `/sogo-auth.php?login=%u`
- Change the user auth tab button to use `/sogo-auth.php?login={{ mailcow_cc_username }}`
- In `sogo-auth.php`, allow the currently logged-in mailbox user through the same SSO path without requiring alias lookup
- In the authenticated SOGo path, treat the current mailbox user as valid even if `user_get_alias_details()` is not the matching mechanism
- Fix the header link substitution guard to check the processed app entry consistently before replacing `%u`

## Why

In our reproduction, mailbox users could hit `Unauthorized` / broken webmail navigation because the user-facing entrypoint skipped the SSO helper, and the SSO helper itself was stricter than necessary for the mailbox owner case.

## Validation

I was able to reproduce the broken user webmail flow and confirm that these changes restore the expected SSO navigation path.

I could not run local PHP linting in this environment because `php` is not installed here.


milkmaker:
Thanks for contributing!

milkmaker:

PR #7205, opened 2026-04-28
[Postfix] Fix main.cf and extra.cf accumulation across container restarts

sfoerster:
## Contribution Guidelines

### Short Description

`postfix.sh` appends `extra.cf` to `main.cf` on every container start, relying on a `sed -i '/Overrides/q'` truncation to reset the dynamic section first. If the `# Overrides #` sentinel is ever absent from `main.cf`, the sed is a silent no-op and the dynamic section accumulates indefinitely — producing a bloated `main.cf` with hundreds of duplicate lines after repeated restarts.

Two fixes:
1. Ensure the sentinel exists before truncating; append it if missing so the reset never silently no-ops. Also tighten the sed pattern from `/Overrides/` to `/# Overrides #/` to avoid accidentally matching the `# User Overrides` line appended later in the same script.
2. Replace the `echo -e "...\n$(cat extra.cf)" > extra.cf` idiom for `myhostname` with a temp-file approach (`grep -v` + `mv`): race-free and produces no stray blank lines across restarts.

### Affected Containers

- postfix-mailcow

### What did you tested?

The exact startup logic from `postfix.sh` via a self-contained bash unit test at `helper-scripts/dev_tests/test_postfix_startup_idempotency.sh`. Covers: `main.cf` stability across 10 simulated restarts, self-healing of a missing sentinel, preservation of user `extra.cf` content, and hostname rotation. Docker image builds clean locally.

Expected `main.cf` and `extra.cf` to be stable after repeated restarts. All 9 test assertions pass. Image build succeeds.


milkmaker:

PR #7215, opened 2026-05-03
Feature: add signature management

tim-riep:
## Contribution Guidelines

### Short Description

Adds centrally managed, server-side **mailbox signatures** so company admins can define reusable templates and rule-based assignment instead of asking each user to maintain a signature in their mail client. Built next to the existing `domain_wide_footer` — it does not replace it.

**Concept**

- **Signature templates** (`signature_templates`): reusable HTML/plain bodies per domain, with name, description, skip-on-reply flag and active flag.
- **Signature rules** (`signature_rules`): bind a template to mailboxes via a `match_type`, ordered by `priority` (higher wins, lowest id breaks ties). Supported match types:
  - `domain` — every active mailbox of the domain
  - `mailbox_tag` — mailbox carries a given tag (re-uses existing `tags_mailbox`)
  - `mailbox_address` — exact address or glob pattern (`*`, `?`)
  - `custom_attribute` — key/value lookup against `mailbox.custom_attributes`
- **Template syntax**: simple Mustache-style placeholders `{{ key }}` (HTML-escaped), plus conditional sections `{{#key}}…{{/key}}` and inverse sections `{{^key}}…{{/key}}` so optional fields like *phone*, *mobile* or *position* disappear cleanly when the attribute isn't set on a mailbox. Standard variables: `name`, `email`, `local_part`, `domain`; every key from `mailbox.custom_attributes` is also available.

**Architecture**

Same shape as the existing `domain_wide_footer` (Lua prefilter + PHP dynmap), so no new container is needed:

```
SMTP submit → postfix → rspamd  ──HTTP──▶ nginx:8081/signature.php
                       └─ MOO_SIGNATURE prefilter (priority 0)
                          └─ lua_mime.add_text_footer(html, plain)
                             └─ task:set_message(rewritten body)
```

`signature.php` (new dynmap) calls `signature_resolve_for_mailbox()` from `functions.signatures.inc.php`, which evaluates active rules in priority order, picks the first matching template, renders both bodies through the Mustache-flavoured engine and returns JSON to the Lua callback.

**Web UI**

- New `/signatures` page (linked from the email dropdown for both `admin` and `domainadmin`).
- Per-domain card layout with each template shown alongside the rules attached to it; *Add template* and *Add rule* open Bootstrap modals with the calling domain or template pre-filled.
- Single-record edit pages reachable via `/edit/signature_template/<id>` and `/edit/signature_rule/<id>`.
- The rule edit form shows / hides `match_key` and `match_value` and switches their labels and placeholders based on the chosen `match_type` (e.g. `domain` rules need neither field, `custom_attribute` needs both).

**Permissions**

Mirrors the existing footer / CRUD pattern: `admin` sees all domains, `domainadmin` is restricted via `hasDomainAccess()` for every action, `user` has no access.

**i18n**

The new `signatures` top-level section and the `edit.signature_*` strings are translated into all 30 currently shipped locales. en-gb and de-de are native quality, the Romance / Germanic / Slavic / EU / Asian locales are checked translations and welcome refinement through Weblate. The freshly added `az-az` is intentionally left for Weblate to fill in alongside its other strings.

**Known limitation**

Same as `domain_wide_footer`: the recipient sees the signature, but the sender's IMAP `Sent` folder shows the original message — SOGo / IMAP clients save the local copy *before* the milter runs. Documenting this as expected behaviour for now; an optional SOGo-side rendering would be a follow-up.

### Affected Containers

- **nginx-mailcow** — serves the new `signature.php` dynmap on port 8081 (same setup as `footer.php`)
- **php-fpm-mailcow** — loads the new `functions.signatures.inc.php` via `prerequisites.inc.php`; runs the new dynmap PHP and the new web routes
- **rspamd-mailcow** — registers the new `MOO_SIGNATURE` Lua prefilter symbol
- **mysql-mailcow** — two new tables (`signature_templates`, `signature_rules`) created via `init_db_schema()` on first start (db version bumped)

No Dockerfile changes; nothing needs an image rebuild.

### What did you tested?

- **Schema migration**: fresh start created `signature_templates` and `signature_rules` with the expected columns, PKs and foreign keys; idempotent on subsequent starts.
- **Web UI** as both `admin` and `domainadmin`:
  - `/signatures` lists templates per domain with their attached rules.
  - Add template via modal (with the domain pre-filled from the card), edit via the single-record page, delete via the row action.
  - Add rule via modal (with the template and domain pre-filled), edit via the single-record page; the dynamic field show/hide and label switching based on `match_type` works for `domain`, `mailbox_tag`, `mailbox_address` and `custom_attribute`.
  - JSON API endpoints (`/api/v1/{add,edit,delete,get}/signature-{template,rule}`) return success / `route not found` as expected.
- **Mail flow**:
  - Sent through SOGo with a mailbox that has matching custom attributes. Recipient received the message with the rendered signature appended; confirmed via raw source.
  - Tested `match_type=domain`, `mailbox_tag`, `mailbox_address` (with `*@` glob) and `custom_attribute` (both with empty and explicit `match_value`).
  - Verified the conditional renderer: `{{#position}}…{{/position}}` blocks vanish when `position` is missing; `{{^position}}…{{/position}}` renders only then; trailing newlines of removed sections are consumed so no blank lines remain.
  - **Skip-on-reply**: with a message carrying `In-Reply-To`, the signature was correctly omitted.
- **Permissions**: `domainadmin` cannot list / edit / delete templates or rules of foreign domains via API or UI.
- **Sanity**: `php -l` clean on every changed PHP file; all 31 lang JSON files parse; no merge-conflict markers anywhere on the branch.

**Awaited:** signatures get rendered with the right per-mailbox values and conditional sections; rule resolution picks the highest-priority active rule; the UI lists everything grouped per domain with field hiding driven by the chosen match type; existing `domain_wide_footer` keeps working untouched.

**Got:** matches the above on every test case.

#### Signature Maintenance
image
image
image

#### Test Cases
Different Signature is appended based on tag
image
image


vaiket:
its okay

milkmaker:

PR #7255, opened 2026-05-24
Fix XML escaping for mobileconfig app passwords

garrettreid:
## Contribution Guidelines

## What does this PR include?
Small bug fix for mobileconfigs with app generated passwords containing ampersands.

### Short Description

Escape HTML special characters in passwords during mobileconfig generation. 

###  Affected Containers

- php-fpm-mailcow

### What did you tested?

I tested the mobile profile generation. Steps:
* Log in as a mail user, go to mailcow preferences
* Next to "Apple connection profile with app password", click the profile links repeatedly until an a profile with an & in the password is generated (~10x is typically sufficient)
* Open the configuration profile, confirm that it has &amp; instead of & inside the XML
* Open the configuration profile on macOS/iOS, confirm that it opens successfully.

Profiles with & in the password now open successfully.

### DIFF
--- a/data/web/mobileconfig.php
+++ b/data/web/mobileconfig.php
@@ -15,6 +15,10 @@ if (!isset($_SESSION['mailcow_cc_role']) || $_SESSION['mailcow_cc_role'] != 'use
 
 error_reporting(0);
 
+function mobileconfig_xml_escape($value) {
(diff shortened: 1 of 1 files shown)


mkuron:
Duplicate of #7212. That fix is a bit simpler, so I'm closing your pull request in favor of it.

PR #7270, opened 2026-06-02
Use string 'true' for the 'exact' query parameter to retrieve users f…

akrist:
## Contribution Guidelines

### Short Description

**Problem:** The _exact_ parameter value _true_ in the Keycloak user query in _keycloak_mbox_login_rest_ is converted to query parameter 1, which Keycloak interprets as “false”. Keycloak then performs a wildcard search (%$email%) and may return multiple users. This could allow a user to log in to another user’s account using their own password.

keycloak_users_exact_1

**Solution:**: Change exact parameter value from true (1) to 'true'.

###  Affected Containers

- php-fpm-mailcow

### What did you tested?

IMAP logins using Keycloak users who have set a _mailcow_password_ attribute.

All users were able to log in using their own password.

keycloak_users_exact_true


milkmaker:

Issue #7287, opened 2026-06-12
API: add/time_limited_alias silently fails with HTTP 200 + empty body when "description" is omitted (field missing from OpenAPI spec)


wryfi:
### Contribution guidelines

- [x] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### Checklist prior issue creation

- [x] I understand that failure to follow below instructions may cause this issue to be closed.
- [x] I understand that vague, incomplete or inaccurate information may cause this issue to be closed.
- [x] I understand that this form is intended solely for reporting software bugs and not for support-related inquiries.
- [x] I understand that all responses are voluntary and community-driven, and do not constitute commercial support.
- [x] I confirm that I have reviewed previous [issues](https://github.com/mailcow/mailcow-dockerized/issues) to ensure this matter has not already been addressed.
- [x] I confirm that my environment meets all [prerequisite requirements](https://docs.mailcow.email/getstarted/prerequisite-system/) as specified in the official documentation.

### Description

The `POST /api/v1/add/time_limited_alias` endpoint silently fails when the request body omits a `description` field. The API returns **HTTP 200 with an empty response body**, but no alias is created. A subsequent `GET /api/v1/get/time_limited_aliases/{mailbox}` then returns an empty JSON object.

The insert behind this endpoint requires a non-null `description` column:

```
$ docker compose exec mysql-mailcow \
  mariadb -u"$DBUSER" -p"$DBPASS" "$DBNAME" \
  -e "SHOW CREATE TABLE spamalias\G" -e "SELECT @@sql_mode\G"
*************************** 1. row ***************************
       Table: spamalias
Create Table: CREATE TABLE `spamalias` (
  `address` varchar(255) NOT NULL,
  `goto` text NOT NULL,
  `validity` int(11) DEFAULT NULL,
  `created` datetime NOT NULL DEFAULT current_timestamp(),
  `modified` datetime DEFAULT NULL ON UPDATE current_timestamp(),
  `description` text NOT NULL,
  `permanent` tinyint(1) NOT NULL DEFAULT 0,
  PRIMARY KEY (`address`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci ROW_FORMAT=DYNAMIC
*************************** 1. row ***************************
[...]

PR #7288, opened 2026-06-12
fixes for #7287: timelimited_alias API bugs

wryfi:
## Contribution Guidelines

## What does this PR include?

Fixes for #7287:

- update OpenAPI spec to correctly reflect available and required fields
- properly throw an exception and HTTP error when required `description` field is not provided
- fix an issue where provided `validity` was always overridden by code

### Short Description

This PR fixes the bugs described in #7287 and an additional bug setting `validity` that was discovered while fixing that issue.

###  Affected Containers

- php-fpm-mailcow
- nginx-mailcow

## Did you run tests?

No automated/runtime tests — mailcow has no test harness covering this API path, and I don't have a live instance to exercise the endpoint against. I ran static validation only and reasoned through the runtime behavior from the code.

### What did you tested?

- `php -l` (PHP 8.2) on the modified `data/web/inc/functions.mailbox.inc.php` — syntax check.
- Schema/format validation of `data/web/api/openapi.yaml` (YAML parse) and `data/web/lang/lang.en-gb.json` (JSON parse).
- Manual code trace of the `/add/time_limited_alias` path: confirmed the original empty-200 came from the `NOT NULL` `description` insert throwing an uncaught `PDOException`, which the global `exception_handler` records to `$_SESSION['return']` but never echoes (PHP halts after the handler), and that the `validity` else branch overwrites valid client input.

- `php -l` → awaited: no syntax errors. got: No syntax errors detected.
- `openapi.yaml` / `lang.en-gb.json` → awaited: parse clean. got: both valid.
- Behavioral changes are not runtime-verified. By inspection they should yield: missing `description` → `danger/description_missing` instead of empty `200`; any other insert failure surfaced via mysql_error; a valid `validity` now preserved instead of forced to `8760`. Runtime confirmation would need a live mailcow testing instance, which I do not currently have.


milkmaker:
Thanks for contributing!

wryfi:
filed against wrong branch

PR #7290, opened 2026-06-12
Fix #7150: passwd-verify.lua wipes auth cache during any nginx outage

### LINKED ISSUE #7150: Dovecot `passwd-verify.lua` wipes auth cache during any nginx restart or unavailability
### Contribution guidelines

- [x] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### Checklist prior issue creation

- [x] I understand that failure to follow below instructions may cause this issue to be closed.
- [x] I understand that vague, incomplete or inaccurate information may cause this issue to be closed.
- [x] I understand that this form is intended solely for reporting software bugs and not for support-related inquiries.
- [x] I understand that all responses are voluntary and community-driven, and do not constitute commercial support.
- [x] I confirm that I have reviewed previous [issues](https://github.com/mailcow/mailcow-dockerized/issues) to ensure this matter has not already been addressed.
- [x] I confirm that my environment meets all [prerequisite requirements](https://docs.mailcow.email/getstarted/prerequisite-system/) as specified in the official documentation.

### Summary

When `nginx:9082` is briefly unreachable from Dovecot — during a container restart, stack update, or any transient Docker network issue — `passwd-verify.lua` returns `PASSDB_RESULT_PASSWORD_MISMATCH` instead of `PASSDB_RESULT_INTERNAL_FAILURE`. This causes Dovecot to wipe the auth cache entry for affected users, resulting in hard authentication failures and client disruption. This affects any routine operation that restarts nginx, including `./update.sh`.

### Background

PR #6488 (merged 2025-05-08, shipped in 2025-05) introduced the HTTP-based auth path in `passwd-verify.lua`, making Dovecot dependent on `nginx:9082` for authentication. The PR deliberately chose `PASSDB_RESULT_PASSWORD_MISMATCH` over `PASSDB_RESULT_INTERNAL_FAILURE` on non-200/401 responses to ensure the auth cache is cleared on credential errors — this is correct behavior for actual HTTP error responses. However, the network error case was not handled: when the HTTPS request fails entirely (connection refused, DNS failure, timeout), `c` is an error string rather than an HTTP status code, but the code treats it identically to a password mismatch.

### Root Cause

In `data/conf/dovecot/auth/passwd-verify.lua`:
```lua
local b, c = https.request { ... }

if c ~= 200 and c ~= 401 then
    dovecot.i_info("HTTP request failed with " .. c .. " for user " .. request.user)
    return dovecot.auth.PASSDB_RESULT_PASSWORD_MISMATCH, "Upstream error"
end
```

When the HTTPS request fails at the network level, `c` is an error string (e.g. `"host or service not provided, or not known"`, `"connection refused"`, `"timeout"`), not an HTTP status code. The condition `c ~= 200` is true for any string, so network errors fall through to
`PASSDB_RESULT_PASSWORD_MISMATCH`, wiping the cache entry.

The existing comment in the code acknowledges this tradeoff explicitly:
```lua
-- Returning PASSDB_RESULT_PASSWORD_MISMATCH will reset the user's auth cache entry.
-- Returning PASSDB_RESULT_INTERNAL_FAILURE keeps the existing cache entry,
-- even if the TTL has expired. Useful to avoid cache eviction during backend issues.
```

The intent is clear — `INTERNAL_FAILURE` should be used for backend unavailability. The network error case simply was not handled.

### Proposed Fix

Distinguish network errors (where `c` is a string) from HTTP errors (where `c` is a number):
```lua
local b, c = https.request { ... }

if type(c) ~= "number" then
    dovecot.i_info("HTTP request failed with " .. tostring(c) .. " for user " .. request.user)
    return dovecot.auth.PASSDB_RESULT_INTERNAL_FAILURE, "Upstream unreachable"
end

if c ~= 200 and c ~= 401 then
    dovecot.i_info("HTTP request failed with " .. tostring(c) .. " for user " .. request.user)
    return dovecot.auth.PASSDB_RESULT_PASSWORD_MISMATCH, "Upstream error"
end
```

This preserves the intended behavior:
- **Network error** (nginx unreachable): return `INTERNAL_FAILURE`, keep cache entry intact
- **HTTP 4xx/5xx** (nginx reachable but rejecting): return `PASSWORD_MISMATCH`, wipe cache entry

### Impact

This affects any operation that restarts or briefly interrupts nginx, including:

- Running `./update.sh` (restarts all containers)
- `docker compose restart nginx-mailcow`
- Host reboots or VM snapshots
- nginx container crashes and automatic restarts
- Transient Docker network issues
- Intentional full stack restarts (e.g. stopping the stack for a consistent volume snapshot before offsite backup)

The last case is worth highlighting: stopping and restarting the entire stack is a legitimate and arguably more consistent alternative to the official `backup_and_restore.sh` script, which requires local disk space proportional to the backup size before data can be shipped offsite.
Users who take this approach — which was completely benign before 2025-05 when auth went directly to MySQL — will now have auth cache entries wiped on every backup cycle.

In all cases, any IMAP client that attempts to reconnect during the nginx unavailability window will have its auth cache entry wiped. iOS Mail clients configured via mobileconfig profile are particularly sensitive as they have no independent Keychain fallback — they may display password prompts hours later when the device wakes. macOS Mail is less affected due to independent Keychain credential caching.

Additionally, `https.TIMEOUT = 30` means auth workers can block for up to 30 seconds per attempt during a restart window, compounding the disruption under load.

### Version

Introduced: 2025-05 (PR #6488)  
Confirmed present in: 2026-01, 2026-03, 2026-03a

### Steps to reproduce:

1. Run a mailcow instance with active IMAP clients connected
2. Restart the nginx container — e.g. via `./update.sh`, `docker compose restart nginx-mailcow`,
   or any operation that cycles the stack
3. During the brief window where `nginx:9082` is unreachable, a Dovecot auth attempt from any
   reconnecting client hits the error path in `passwd-verify.lua`
4. Auth cache entries for affected users are wiped
5. Mail clients receive hard authentication failures; iOS clients in particular may display
   password prompts when the device next wakes, even hours later

### Logs:

```plain text
dovecot-mailcow-1 | auth-worker(139): HTTP request failed with host or service not provided, or not known for user user@example.com
dovecot-mailcow-1 | auth-worker(139): HTTP request failed with host or service not provided, or not known for user user@example.com
dovecot-mailcow-1 | auth-worker(139): lua(user@example.com,x.x.x.x,...): Password mismatch (SHA1 of given password: xxxxxx)
dovecot-mailcow-1 | imap-login: Disconnected: Connection closed (auth failed, 2 attempts in 8 secs): user=<user@example.com>, method=PLAIN, ...
```

Note that `"host or service not provided, or not known"` is a `getaddrinfo` error string — not an HTTP status code — being passed to the log concatenation and then triggering `PASSWORD_MISMATCH`.

### Which branch are you using?

master (stable)

### Which architecture are you using?

x86_64

### Operating System:

Debian GNU/Linux 13 (trixie)

### Server/VM specifications:

12GB, 4 cores

### Is Apparmor, SELinux or similar active?

no

### Virtualization technology:

kvm

### Docker version:

29.2.1

### docker-compose version or docker compose version:

5.1.0

### mailcow version:

2026-01

### Reverse proxy:

none

### Logs of git diff:

```plain text
n/a
```

### Logs of iptables -L -vn:

```plain text
n/a
```

### Logs of ip6tables -L -vn:

```plain text
n/a
```

### Logs of iptables -L -vn -t nat:

```plain text
n/a
```

### Logs of ip6tables -L -vn -t nat:

```plain text
n/a
```

### DNS check:

```plain text
198.252.206.1
```

wryfi:
## Contribution Guidelines

### Short Description

Fixes #7150.

`passwd-verify.lua` wiped the Dovecot auth cache for affected users whenever
a host reboot, a full-stack snapshot/backup, or any transient Docker network blip.

On a network-level failure, `https.request` returns `nil` plus an error *string*
(e.g. `"host or service not provided, or not known"`), so `c` is not a numeric HTTP
status code. The existing `if c ~= 200 and c ~= 401` check treated this identically
to a credential rejection and returned `PASSDB_RESULT_PASSWORD_MISMATCH`, which clears
the cache entry — causing hard auth failures for reconnecting clients (iOS Mail in
particular, which has no Keychain fallback).

This PR adds a guard ahead of the HTTP-status check: when `c` is not a number, the
request failed at the network level, so it returns `PASSDB_RESULT_INTERNAL_FAILURE`
("Upstream unreachable"), which **keeps** the existing cache entry instead of wiping
it. Genuine HTTP responses (200/401, and 4xx/5xx errors) are handled exactly as before.
The log concatenation is also hardened to `tostring(c)` so it can't error on a
non-string/nil value.

This restores the pre-2025-05 behavior where a brief nginx interruption was benign for
already-authenticated clients.

### Affected Containers

- dovecot-mailcow

## Did you run tests?

Yes — tested on a live mailcow instance by overriding `passwd-verify.lua` and exercising
the real-world scenario that originally surfaced the bug.

### What did you tested?

The Dovecot auth path (`dovecot-mailcow`) through a full-stack restart, using my backup
procedure as the trigger: stop the entire stack, snapshot the volumes, then start the
stack again. This is exactly the kind of operation that leaves `nginx:9082` unreachable
while Dovecot auth workers are still serving reconnecting IMAP clients.

I first noticed the failure on the stock file, then applied the patched
`passwd-verify.lua` as an override and repeated the same backup/restart cycle with active
IMAP clients connected.

- Stock file, full-stack restart → **awaited:** the documented bug. **got:** confirmed —
  auth cache entries for connected users were wiped and clients hit hard authentication
  failures during the nginx-unavailable window.
- Patched file, same full-stack restart → **awaited:** cache entries retained, clients
  unaffected. **got:** as awaited — no cache wipe, and connected IMAP clients stayed
  authenticated across the entire stop/snapshot/restart cycle, with no re-auth prompts.
- `luac -p data/conf/dovecot/auth/passwd-verify.lua` → **awaited:** no syntax errors.
  **got:** syntax OK.

### DIFF
--- a/data/conf/dovecot/auth/passwd-verify.lua
+++ b/data/conf/dovecot/auth/passwd-verify.lua
@@ -32,8 +32,17 @@ function auth_password_verify(request, password)
   -- Returning PASSDB_RESULT_PASSWORD_MISMATCH will reset the user's auth cache entry.
   -- Returning PASSDB_RESULT_INTERNAL_FAILURE keeps the existing cache entry,
   -- even if the TTL has expired. Useful to avoid cache eviction during backend issues.
+
+  -- On a network-level failure (nginx unreachable, DNS failure, timeout) https.request
+  -- returns nil plus an error string, so c is not a numeric HTTP status code. Treat this
+  -- as a backend outage and keep the cache entry, rather than wiping it as a mismatch.
+  if type(c) ~= "number" then
+    dovecot.i_info("HTTP request to auth backend failed with " .. tostring(c) .. " for user " .. request.user)
+    return dovecot.auth.PASSDB_RESULT_INTERNAL_FAILURE, "Upstream unreachable"
+  end
+
   if c ~= 200 and c ~= 401 then
-    dovecot.i_info("HTTP request failed with " .. c .. " for user " .. request.user)
+    dovecot.i_info("HTTP request failed with " .. tostring(c) .. " for user " .. request.user)
     return dovecot.auth.PASSDB_RESULT_PASSWORD_MISMATCH, "Upstream error"
   end
 


milkmaker:

FreddleSpl0it (approved):

Issue #5518, opened 2023-11-08
Fail2ban GUI reports incorrect ban time


gitterdoneplease:
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... I understand that not following the below instructions will result in immediate closure and/or deletion of my issue.
- [X] ... I have understood that this bug report is dedicated for bugs, and not for support-related inquiries.
- [X] ... I have understood that answers are voluntary and community-driven, and not commercial support.
- [X] ... I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

### Description

```plain text
I've set the Fail2ban parameters like this:

Ban time (s): 1800
Max. ban time (s): 100000
✓ Ban time is incremented with each ban
Max. attempts: 10
Retry window (s) for max. attempts: 600
IPv4 subnet size to apply ban on (8-32): /24

The GUI reports attackers like:

141.98.11.0/24 (8388580h 23m 06s)

1,000 years might be a little too much. :)
```

### Logs:

```plain text
No error occurs
```

### Steps to reproduce:

```plain text
Enable Fail2ban.

Set desired parameters.

Get attacked.

Check GUI.
```

### Which branch are you using?

master

### Operating System:

Ubuntu 22.04.3 LTS

### Server/VM specifications:

VM, 16 cores, 32GB of RAM

### Is Apparmor, SELinux or similar active?

no

### Virtualization technology:

Proxmox (KVM/Qemu)

### Docker version:

24.0.5

### docker-compose version or docker compose version:

v2.19.1

### mailcow version:

2023-10a

### Reverse proxy:

N/A

### Logs of git diff:

```plain text
diff --git a/data/conf/dovecot/dovecot.conf b/data/conf/dovecot/dovecot.conf
index 159e39f4..026a5fd5 100644
--- a/data/conf/dovecot/dovecot.conf
+++ b/data/conf/dovecot/dovecot.conf
@@ -164,7 +164,7 @@ service lmtp {
   }
   user = vmail
 }
-listen = *,[::]
+listen = *
 ssl_cert = </etc/ssl/mail/cert.pem
 ssl_key = </etc/ssl/mail/key.pem
 userdb {
[...]

Issue #5125, opened 2023-03-14
Netfilter: "Error reading log line from pubsub" causing bans to clear several times a day


pollock83:
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [x] ... I understand that not following the below instructions will result in immediate closure and/or deletion of my issue.
- [X] ... I have understood that this bug report is dedicated for bugs, and not for support-related inquiries.
- [X] ... I have understood that answers are voluntary and community-driven, and not commercial support.
- [X] ... I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

### Description

```plain text
Several times a day the netfilter container will log the following error: "Error reading log line from pubsub". Immediately after throwing this error, it will clear all bans from fail2ban. This is essentially making the netfilter container useless as within a few hours of an IP being banned, the container will almost certainly encounter this error and the IP will be unbanned anyway. Sometimes this happened every few hours, sometimes it can happen multiple times within a few minutes.
```

### Logs:

```plain text
mailcowdockerized-netfilter-mailcow-1  | 2023-03-14T11:51:59.798388367Z Error reading log line from pubsub
mailcowdockerized-netfilter-mailcow-1  | 2023-03-14T11:51:59.810342474Z Clearing all bans
mailcowdockerized-netfilter-mailcow-1  | 2023-03-14T11:51:59.810609023Z Unbanning 85.185.66.0/24
mailcowdockerized-netfilter-mailcow-1  | 2023-03-14T11:51:59.815155481Z Unbanning 111.70.18.0/24
mailcowdockerized-netfilter-mailcow-1  | 2023-03-14T11:51:59.816334090Z Unbanning 1.234.47.0/24
mailcowdockerized-netfilter-mailcow-1  | 2023-03-14T11:51:59.817276635Z Unbanning 124.195.191.0/24
mailcowdockerized-netfilter-mailcow-1  | 2023-03-14T11:51:59.818385944Z Unbanning 136.169.211.0/24
mailcowdockerized-netfilter-mailcow-1  | 2023-03-14T11:51:59.819767913Z Unbanning 80.227.147.0/24
[...]

PR #7330, opened 2026-07-13
netfilter: prevent autopurge/blacklistUpdate/whitelistUpdate threads from dying silently on uncaught exceptions

### LINKED ISSUE #1162: Banned IP's are only unbanned at netfilter container restart
I'm experiencing an issue with the netfilter container.
The IP's that are banned, are never unbanned again.
Only after a netfilter container restart, all banned IP's are unbanned.

The only change I have made to the default setup is changing the ban size from /24 to /32.
From what I can see in the log, everything seems to work as expected, but I only see unbans when restarting.

The following log snippet shows the issue:
fail2ban_mailcow_issue
And from the logfile (reverse order):
```
2018-03-12T17:15:49.173041891Z 80.71.xxx.xxx matched rule id 3
2018-03-12T17:15:49.175335618Z Banning 80.71.xxx.xxx/32 for 30 minutes
2018-03-12T18:12:28.696823816Z Clearing all bans
2018-03-12T18:12:28.700002630Z Unbanning 94.102.xxx.xxx/32
2018-03-12T18:12:28.717804329Z Unbanning 107.170.xxx.xxx/32
2018-03-12T18:12:28.724197916Z Unbanning 80.71.xxx.xxx/32
2018-03-12T18:12:29.950214862Z Cleaning previously cached bans
2018-03-12T18:12:29.952490285Z Subscribing to Redis channel F2B_CHANNEL
2018-03-12T18:13:23.274285283Z 94.102.xxx.xxx matched rule id 1
2018-03-12T18:13:23.278280614Z 9 more attempts in the next 600 seconds until 94.102.xxx.xxx/32 is banned
```
The logs aren't very verbose, so any suggestions on how to debug further are greatly appreciated.

aleksa-dobre:
## Problem

Several long-standing issues report that fail2ban/netfilter bans never expire automatically and only a full `netfilter-mailcow` container restart clears them:
- #1162 (2018, closed with no documented fix)
- #5879 (2024, still open)
- #5125 ("Error reading log line from pubsub" causing bans to clear)
- #5518 (GUI shows absurd/negative ban times)

Symptoms reported by multiple users across these threads: negative ban times in the admin UI, the "Unban" button not doing anything, banned IPs staying blocked for days or weeks past their configured `ban_time`, restarting the netfilter container being the only known workaround (which wipes *all* current bans, not just the stuck one).

## Root cause

`autopurge()`, `blacklistUpdate()`, and `whitelistUpdate()` each run in their own daemon thread with a `while not quit_now:` loop, but — unlike `watch()` (and `mailcowChainOrder()`), which already wrap their loop body in `try/except` and recover by setting `quit_now`/`exit_code` — none of these three catch exceptions. Python threads do not restart themselves after an uncaught exception: once one of these loops raises anything, that thread silently disappears for the remaining lifetime of the container, while the rest of the process (admin UI, and the ability to *create* new bans via `watch()`) keeps working fine — which is exactly why this is so easy to miss and so hard to reproduce on demand.

We captured two real occurrences of this in our own production logs:

```
Exception in thread Thread-2 (autopurge):
Traceback (most recent call last):
  File "/usr/lib/python3.12/site-packages/redis/connection.py", line 2965, in get_connection
    connection.can_read()
  ...
redis.exceptions.ConnectionError: Error 113 connecting to <redis-ip>:6379. Host is unreachable.
```
(a transient Redis connectivity blip)

and, about two weeks later, presumably after the container had been restarted for an unrelated reason (which would explain why `autopurge` was alive again to hit a *different* exception):

```
Exception in thread Thread-2 (autopurge):
Traceback (most recent call last):
  File "/usr/lib/python3.12/threading.py", line 1075, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.12/threading.py", line 1012, in run
    self._target(*self._args, **self._kwargs)
  File "/app/main.py", line 308, in autopurge
    MAX_ATTEMPTS = int(f2boptions['max_attempts'])
                       ~~~~~~~~~~^^^^^^^^^^^^^^^^
KeyError: 'max_attempts'
```

This second traceback matches what @evilMouse independently reported in #5879 ("...it hanged with the last line visible: `Error reading log line from pubsub: 'max_attempts'`"). The likely mechanism: `refreshF2boptions()` does `f2boptions = {}` and then repopulates the dict key by key, with no locking, while `f2boptions` is a module-level global that's read concurrently from multiple threads (`autopurge`, and `ban()` which is called from `watch()`). If another thread reads `f2boptions['max_attempts']` in the brief window after it's been reset to `{}` but before it's repopulated, it raises `KeyError`. This is a real data race independent of (and in addition to) the missing exception handling — but the missing exception handling is what turns a transient, one-off hiccup into a permanent outage that requires manually restarting the container to notice and fix.

Once `autopurge()` is dead, entries in Redis `F2B_ACTIVE_BANS` are never checked for expiry again, so their corresponding nftables/firewall rules stay in place indefinitely, well past the ban duration shown in the UI — including for legitimate users who happen to share a NAT/IP with unrelated malicious traffic, or who get banned for a self-inflicted/transient reason (e.g. a device somewhere on their network retrying a stale cached password).

## Fix

Apply the same recovery pattern already used by `watch()`/`mailcowChainOrder()` to the other three background threads: wrap the loop body in `try/except Exception`, log it via the existing logger, and set `quit_now = True; exit_code = 2` so the process exits cleanly — `restart: always` in `docker-compose.yml` then brings up a fresh instance automatically. This is a minimal, low-risk change: no new dependencies, no behavior change on the happy path, it just stops a single transient error in one loop iteration from permanently disabling that thread for the rest of the container's life.

## Testing

- `python3 -m py_compile data/Dockerfiles/netfilter/main.py` passes.
- The recovery pattern (log + `quit_now`/`exit_code` + rely on `restart: always`) is not new — it's the existing convention this file already uses in `watch()` and `mailcowChainOrder()` for unrecoverable thread state, so this change makes the other three background threads consistent with that established convention rather than introducing a new one.
- While developing this fix we ran an external workaround in production (a cron job directly cleaning expired `F2B_ACTIVE_BANS` Redis entries and their nftables rules, bypassing the dead thread) for several days without incident, which is what let us capture the tracebacks quoted above from real traffic.

Relates to / likely fixes #1162, #5879, #5125, #5518.


milkmaker:
Thanks for contributing!

aleksa-dobre:
Closing this in favor of #7120, which addresses the same root problem more thoroughly (it also fixes the `unban()` early-return-when-not-in-local-`bans`-dict bug, removes some dead code that was causing one of the crash modes, and adds locking around unban to avoid a race with `ban()`). Left a comment there with two real tracebacks we captured in production that corroborate the diagnosis, in case they're useful while that PR is still being tested.

Thanks for looking either way!

PR #7331, opened 2026-07-13
feat: distinguish login portals visually

### LINKED ISSUE #7327: Make the different login portals MUCH more obvious
### Summary

The three different login portals need to be much more obvious.

### Motivation

For occasional users / an admin covering for a colleague just the once and such, we find that the three "different, but all alike" login pages of mailcow are a perpetual source of confusion and not inconsiderable frustration. The "switch to a different login portal" links at the bottom are extremely low-key unless you're very actively looking for them.

At least KeePassXC-Browser in Firefox by default does not tell the different portals apart, offering all logins on all portals, adding to this confusion.

### Additional context

I'd suggest to set the background color of the different portals to different values, so people acquire "back brain" understanding of what's going on there.

My suggestion would be:
- none/white/or a green hue for the regular user login
- purple or blue for the domain admin login
- red for the system admin login

The "switch to a different login portal" links should be much larger and more obvious, more like buttons, and reflect those colors.

SahilGawade07:
## Contribution Guidelines

### Short Description

This PR addresses #7327 by making the user, domain administrator, and system administrator login portals more visually distinct.

Each login portal now has its own visual identity:
- Green for the user portal
- Blue for the domain administrator portal
- Red for the system administrator portal

The links under "Other logins" are also displayed as more prominent, color-coded buttons based on the destination portal.

Portal-specific styling is provided for both light and dark mode.

### Affected Containers

- php-fpm-mailcow
- nginx-mailcow

### What did you tested?

Tested all three login portals on a local mailcow instance:

- User login portal
- Domain administrator login portal
- System administrator login portal
- Portal switching through the "Other logins" buttons
- Light mode
- Dark mode

Expected the three login portals to be visually distinguishable and the portal-switch links to be more prominent while preserving the existing navigation and visibility conditions.

The result matched the expected behavior. Each portal has a distinct visual identity, the "Other logins" actions are displayed as color-coded buttons, and navigation between the login portals works as expected in both light and dark mode.

#### User portal

User login portal with green visual identity

#### Domain administrator portal

Domain administrator login portal with blue visual identity

#### System administrator portal

System administrator login portal with red visual identity

Closes #7327

### DIFF
--- a/data/web/css/build/014-mailcow.css
+++ b/data/web/css/build/014-mailcow.css
@@ -433,3 +433,24 @@ button[aria-expanded='true'] > .caret {
 .hr-title:not(:empty)::after {
   margin-left: 10px;
 }
+/* Visual differentiation of login portals */
+body.login-portal-user {
+  background-color: #f2faf5 !important;
+}
+body.login-portal-user .card {
+  border-top: 4px solid #198754 !important;
+}
+
+body.login-portal-domainadmin {
+  background-color: #f3f4fc !important;
+}
+body.login-portal-domainadmin .card {
+  border-top: 4px solid #0d6efd !important;
+}
+
+body.login-portal-admin {
+  background-color: #fdf3f3 !important;
+}
+body.login-portal-admin .card {
+  border-top: 4px solid #dc3545 !important;
+}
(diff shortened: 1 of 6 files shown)


apio-sys:
I don't think this is a good idea. If someone have personalized logo's there you might have a very strange/noisy color mix. I personally would not want to implement this so maybe make it a toggle switch in admin for those who really want to (maybe with fields to specify each bg color). Thanks for the work and sorry to be negative, but I don't feel you can impose this on all (that use the product across multiple servers) without giving a choice.

Issue #7332, opened 2026-07-13
CORS "Allowed Origins" only accepts bare hostnames, so it can never match a real browser Origin header


fallmo:
### Contribution guidelines

- [x] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### Checklist prior issue creation

- [x] I understand that failure to follow below instructions may cause this issue to be closed.
- [x] I understand that vague, incomplete or inaccurate information may cause this issue to be closed.
- [x] I understand that this form is intended solely for reporting software bugs and not for support-related inquiries.
- [x] I understand that all responses are voluntary and community-driven, and do not constitute commercial support.
- [x] I confirm that I have reviewed previous [issues](https://github.com/mailcow/mailcow-dockerized/issues) to ensure this matter has not already been addressed.
- [x] I confirm that my environment meets all [prerequisite requirements](https://docs.mailcow.email/getstarted/prerequisite-system/) as specified in the official documentation.

### Description

The API CORS settings (Configuration → Administrative Settings → CORS) let an admin configure an "Allowed Origins" list, intended to permit browser JavaScript on other origins to call the mailcow API cross-origin — as this project's own [2023-07 release announcement](https://mailcow.email/posts/2023/release-2023-07/) describes it:

> "Other origins" ... may include having different domain, port or scheme (HTTP or HTTPS).

That's the correct definition of an origin. However, the actual validation in `functions.inc.php`, `cors()`, `case "edit"` rejects anything that isn't a bare hostname:

```php
// functions.inc.php:2289
if (!filter_var($origin, FILTER_VALIDATE_DOMAIN, FILTER_FLAG_HOSTNAME) && $origin != '*') {
```

`FILTER_VALIDATE_DOMAIN` / `FILTER_FLAG_HOSTNAME` only accepts strings like `example.com` — a scheme (`http://`) or port (`:3000`) fails validation outright.

This breaks the feature at a fundamental level: per the Fetch/CORS spec, a browser's `Origin` header is always the full `scheme://host[:port]`, never a bare hostname. The matching code later on does an exact string comparison against that header:

```php
// functions.inc.php:2354-2359
$allowed_origins = explode(', ', $cors_settings['allowed_origins']);
$cors_settings['allowed_origins'] = $allowed_origins[0];
if (in_array('*', $allowed_origins)){
  $cors_settings['allowed_origins'] = '*';
[...]

PR #7346, opened 2026-07-14
[Web] make CORS allowed origins accept and match real origins

smpaz7467:
## What

The CORS "Allowed Origins" list cannot allow any origin. It validates entries as
bare hostnames, but matches them against a header that is never a bare hostname.

**Saving.** `cors()` `case "edit"` validated each entry with
`FILTER_VALIDATE_DOMAIN` / `FILTER_FLAG_HOSTNAME`, which accepts `example.com` and
rejects anything carrying a scheme or a port. A real origin could not be saved at
all — it failed with `cors_invalid_origin`.

**Matching.** `set_headers` then compared that hostname list against
`$_SERVER['HTTP_ORIGIN']`, which per Fetch is always `scheme://host[:port]`. The
`in_array` could never match, so the code fell through to reflecting
`$allowed_origins[0]` — the first configured entry — regardless of who was asking.

Net effect: the allowlist only ever worked via the `*` wildcard, which defeats the
point of an allowlist.

## Fix

Validate and store origins in the form a browser actually sends, and match on that.
One shared `normalize_cors_origin()` is used by both the save path and the match
path, so the two can't drift apart again.

A bare hostname is still accepted and read as `https://<host>`. That is deliberate:
it is what the shipped default has always been (`$_SERVER['SERVER_NAME']`), and the
settings form is pre-filled with it — so rejecting it would mean an admin who opens
the CORS page and clicks Save gets an error on a value they never chose, and every
existing install would have to re-enter its origins by hand. https is the only
scheme mailcow serves, so the inference is safe. The defaults now carry the scheme
too.

An origin that is *not* allowed now gets **no** `Access-Control-Allow-Origin` header
at all. Reflecting some other configured origin told the browser nothing and only
obscured the mismatch (this is what produced the confusing error in the issue's
screenshot). `Vary: Origin` is sent when the header depends on the request origin.

## Verification

`php -l` clean. Exercised the real functions from the patched file.

Saving — what an admin can actually store:

| admin types | before | after |
|---|---|---|
| `http://localhost:3000` | **rejected** | `http://localhost:3000` |
| `https://app.example.com` | **rejected** | `https://app.example.com` |
| `https://a.com:8443` | **rejected** | `https://a.com:8443` |
| `mail.example.org` | `mail.example.org` | `https://mail.example.org` |
| `*` | `*` | `*` |

Matching — header emitted for a given `Origin`:

| stored | Origin | before | after |
|---|---|---|---|
| `http://localhost:3000` | `http://localhost:3000` | *unreachable, cannot be saved* | reflected |
| `mail.example.org` (legacy) | `https://mail.example.org` | `ACAO: mail.example.org` (never matches) | reflected |
| `https://a.com, https://b.com` | `https://b.com` | *unreachable* | reflected |
| `https://a.com, https://b.com` | `https://evil.com` | `ACAO: https://a.com` | **no header** |
| `https://a.com:8443` | `https://a.com:9000` | *unreachable* | **no header** (port is part of the origin) |
| `*` | anything | `*` | `*` |
| `https://a.com` | *(none sent)* | `ACAO: https://a.com` | **no header** |

Normalizer accepts `http`/`https` with optional port, IPv4 and IPv6 literals, `*`,
lowercases the host, tolerates a trailing slash; rejects paths, query strings,
fragments, embedded credentials, non-http(s) schemes (`ftp:`, `javascript:`,
`file:`), empty input and out-of-range ports.

Refs #7332

### DIFF
--- a/data/web/inc/functions.inc.php
+++ b/data/web/inc/functions.inc.php
@@ -2269,6 +2269,56 @@ function rspamd_ui($action, $data = null) {
     break;
   }
 }
+// Bring a configured origin into the form a browser sends in its Origin header,
+// scheme://host[:port]. Returns false if it cannot be one.
+// A bare hostname is legacy configuration from when only hostnames validated here,
+// it is read as https since that is the only scheme mailcow serves.
+function normalize_cors_origin($origin) {
+  $origin = trim($origin);
+  if ($origin === '') {
+    return false;
+  }
+  if ($origin === '*') {
+    return '*';
+  }
+  if (!preg_match('~^[a-z][a-z0-9+.-]*://~i', $origin)) {
+    $origin = 'https://' . $origin;
+  }
+
+  $parts = parse_url($origin);
+  if ($parts === false || empty($parts['scheme']) || empty($parts['host'])) {
+    return false;
+  }
+  // an origin is only a scheme, a host and an optional port
+  if (isset($parts['query']) || isset($parts['fragment']) || isset($parts['user']) || isset($parts['pass'])) {
+    return false;
+  }
+  if (isset($parts['path']) && $parts['path'] !== '' && $parts['path'] !== '/') {
+    return false;
+  }
+
+  $scheme = strtolower($parts['scheme']);
+  if ($scheme !== 'http' && $scheme !== 'https') {
+    return false;
+  }
+
+  $host = strtolower($parts['host']);
+  // parse_url keeps the brackets of an IPv6 literal, they are not part of the address
+  $address = trim($host, '[]');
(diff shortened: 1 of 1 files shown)


smpaz7467:
Closing this as superseded by #7333, which was merged for #7332 while this was open.

That PR takes the opposite design decision to this one: it requires `allowed_origins` to be entered as full origins and validates them strictly (`valid_origin()`), where this PR normalized a bare host name into `https://<host>` so existing configurations kept working. Since the strict form is now what mailcow stores and documents, normalizing bare hosts would only reintroduce ambiguity, so this branch is not worth rebasing.

One gap does remain after #7333: the fallback on the read path in `cors('get')` still substitutes a bare `$_SERVER['SERVER_NAME']`, so any installation that never saved CORS settings explicitly still answers `Access-Control-Allow-Origin` with a host name rather than an origin. I have opened #7381 for just that, using the same `getBaseURL()` default the save path already uses.

PR #7384, opened 2026-07-30
[DockerAPI] Limit container discovery to mailcow project

### LINKED ISSUE #7374: Mailbox and domain deletion consistently take ~12.3 seconds on 2026-07
## Description

After upgrading a development Mailcow stack from `2026-03a` to `2026-07`, deletion through the REST API consistently takes approximately 12.3 seconds. Mailbox creation and retrieval remain fast.

This reproduces for both mailbox and domain deletion. The API returns HTTP 200 and the requested item is deleted successfully; the issue is the fixed server-side latency.

I searched existing issues before filing and did not find a matching report.

## Steps to reproduce

1. Run Mailcow `2026-07`.
2. Create a disposable domain through the REST API.
3. Create a mailbox under that domain.
4. Measure mailbox deletion:

```bash
curl -ksS \
  -H "X-API-Key: $API_KEY" \
  -H "Content-Type: application/json" \
  -X POST \
  "$MAILCOW_URL/api/v1/delete/mailbox" \
  --data '["probe@example.test"]' \
  -o /dev/null \
  -w 'code=%{http_code} starttransfer=%{time_starttransfer} total=%{time_total}\n'
```

5. Repeat with fresh mailboxes.
6. Repeat against `/api/v1/delete/domain` with a disposable domain.

## Observed result

Five successful mailbox deletions:

```text
12.395799s
12.362496s
12.325501s
12.328252s
12.339893s
```

The variation across these runs was approximately 71ms. In each request, essentially the entire duration was spent waiting for the first response byte.

Domain deletion on the same stack:

```text
12.296195s
```

Control measurements from the same environment:

```text
mailbox creation:                    0.062491s
mailbox retrieval:                   0.011028s
direct dockerapi maildir cleanup:    0.110933s
```

The direct cleanup control invoked the same Docker API `maildir` / `cleanup` operation used by the mailbox delete path against the Dovecot container. It completed successfully in about 111ms.

No long-running MariaDB query was visible while the REST request was waiting.

The same deletion operation was sub-second before upgrading from `2026-03a`.

## Expected result

Mailbox and domain deletions should complete without an unexplained fixed delay of approximately 12 seconds.

## Logs

Relevant Docker API timing around a mailbox deletion:

```text
2026-07-27T18:29:08.342719719Z GET /containers/json 200
2026-07-27T18:29:08.353490067Z api call: container_post__exec__maildir__cleanup
2026-07-27T18:29:08.445728628Z POST /containers/<redacted>/exec 200
```

The Dovecot maildir-cleanup portion completed in roughly 100ms and did not emit a corresponding error. Nginx logged the delete API request with HTTP 200 only after the approximately 12-second wait.

## Possible regression boundary

Commit [`277a307f`](https://github.com/mailcow/mailcow-dockerized/commit/277a307fb91f701f2f2619debd50ec183ca34192) changed mailbox deletion flow while fixing SOGo view refresh by removing an early `return true`. This is only a possible regression boundary, not a diagnosed cause. The reproducing environment has `SKIP_SOGO=y`, and domain deletion shows the same latency.

## System information

```text
Branch/tag:             2026-07 (detached at release tag)
Architecture:           x86_64
Operating system:       Ubuntu 24.04.4 LTS
CPU:                    32 logical CPUs
Memory:                 31.3 GiB
Docker client/server:   29.5.2 / 29.5.2
Docker Compose:         v5.1.4
Storage driver:         overlay2
Container host:         Docker Desktop
Reverse proxy:          none beyond Mailcow nginx
SKIP_SOGO:              y
```

This was first reproduced in a reduced development stack containing the normal Mailcow PHP-FPM, nginx, MariaDB, Redis, Docker API, Dovecot, Postfix, Rspamd, and Unbound services. SOGo, ClamAV, and services unnecessary for API development are omitted. I can provide additional measurements from a stock full installation if required.

Local configuration differences are limited to the reduced-stack configuration and custom Dovecot push-notification integration. The delay occurs before the Docker API maildir cleanup request, so that cleanup and its Dovecot customization do not appear to explain the latency.

entomb:
Fixes #7374.

## Contribution Guidelines

### Short Description

Mailbox and domain deletion call the Docker API to discover the Dovecot container before cleaning up mail storage. The `/containers/json` endpoint previously inspected every running container on the Docker host before its callers filtered the response to the current mailcow project.

On the reproducing host, four unrelated containers each took three seconds to inspect. Those sequential timeouts caused the consistent ~12.3-second deletion latency reported in #7374.

This change filters the Docker Engine container list by the current Compose project label before inspecting container metadata. It also passes `COMPOSE_PROJECT_NAME` to `dockerapi-mailcow` and bumps the image patch tag to `2.12a`.

### Affected Containers

- dockerapi-mailcow

## Did you run tests?

Yes.

### What did you test?

- Reproduced mailbox and domain deletion latency through the REST API.
- Timed Docker API container discovery independently from maildir cleanup.
- Built and ran the patched Docker API image against the same Docker host.
- Checked both running-container discovery and `all=true` discovery.
- Repeated mailbox and domain deletion through the normal REST API with the stack routed through the patched Docker API image.

Expected: Docker API discovery should inspect only containers in the current mailcow Compose project and should not inherit delays from unrelated containers.

Before: discovery returned 21 cross-project containers and took approximately 12.2 seconds.

After: discovery returned only the 9 containers in the mailcow project and completed in approximately 0.10 seconds. The `all=true` path remained project-scoped and completed in approximately 0.11 seconds.

With the patched image, mailbox deletion completed in 0.211 seconds and domain deletion completed in 0.215 seconds. Both requests returned HTTP 200 with success responses, and the mailbox and domain were absent from the database afterward.

### DIFF
--- a/data/Dockerfiles/dockerapi/main.py
+++ b/data/Dockerfiles/dockerapi/main.py
@@ -115,7 +115,12 @@ async def get_containers(all: bool = False):
 
   containers = {}
   try:
-    for container in (await dockerapi.async_docker_client.containers.list(all=all)):
+    # Filter before inspecting containers so unrelated projects cannot delay
+    # mailcow operations when their container metadata is unavailable.
+    container_filters = json.dumps({
+      "label": [f"com.docker.compose.project={os.environ['COMPOSE_PROJECT_NAME']}"]
+    })
+    for container in (await dockerapi.async_docker_client.containers.list(all=all, filters=container_filters)):
       container_info = await container.show()
       containers.update({container_info['Id']: container_info})
     return Response(content=json.dumps(containers, indent=4), media_type="application/json")
--- a/data/Dockerfiles/dockerapi/main.py
+++ b/data/Dockerfiles/dockerapi/main.py
@@ -115,12 +115,15 @@ async def get_containers(all: bool = False):
 
   containers = {}
   try:
-    # Filter before inspecting containers so unrelated projects cannot delay
-    # mailcow operations when their container metadata is unavailable.
-    container_filters = json.dumps({
-      "label": [f"com.docker.compose.project={os.environ['COMPOSE_PROJECT_NAME']}"]
-    })
-    for container in (await dockerapi.async_docker_client.containers.list(all=all, filters=container_filters)):
+    container_list_options = {"all": all}
+    compose_project_name = os.environ.get('COMPOSE_PROJECT_NAME')
+    if compose_project_name:
+      # Filter before inspecting containers so unrelated projects cannot delay
+      # mailcow operations when their container metadata is unavailable.
+      container_list_options["filters"] = json.dumps({
+        "label": [f"com.docker.compose.project={compose_project_name}"]
+      })
+    for container in (await dockerapi.async_docker_client.containers.list(**container_list_options)):
       container_info = await container.show()
       containers.update({container_info['Id']: container_info})
     return Response(content=json.dumps(containers, indent=4), media_type="application/json")
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -601,7 +601,7 @@ services:
             - watchdog
 
     dockerapi-mailcow:
-      image: ghcr.io/mailcow/dockerapi:2.12
+      image: ghcr.io/mailcow/dockerapi:2.12a
       security_opt:
         - label=disable
       restart: always
@@ -610,6 +610,7 @@ services:
       environment:
         - DBROOT=${DBROOT}
         - TZ=${TZ}
+        - COMPOSE_PROJECT_NAME=${COMPOSE_PROJECT_NAME:-mailcow-dockerized}
         - REDIS_SLAVEOF_IP=${REDIS_SLAVEOF_IP:-}
         - REDIS_SLAVEOF_PORT=${REDIS_SLAVEOF_PORT:-}
         - REDISPASS=${REDISPASS}


smpaz7467:
Took a look. The approach is right: the cost is the per-container `container.show()` in the inspect loop, not the listing itself, so filtering before that loop is what removes the stall. Your 37→9 numbers are consistent with that.

I verified the mechanism against the stock `dockerapi:2.12` image, running aiodocker against a live Docker socket over three runs. A matching label filter returns only the project's containers; a non-matching one returns an empty list without raising. The `json.dumps` filter string works as written.

Worth noting for review that this follows what mailcow already does rather than introducing a new convention: php-fpm is already given `COMPOSE_PROJECT_NAME=${COMPOSE_PROJECT_NAME:-mailcow-dockerized}` (docker-compose.yml:176, in its environment block — the same expression this PR adds for dockerapi), `functions.docker.inc.php` already filters on `com.docker.compose.project` client-side in both `get_id` and `containers`, and ofelia filters on that same label. So this moves an existing filter server-side.

One optional nit: `os.environ['COMPOSE_PROJECT_NAME']` raises KeyError if the variable is ever absent, and since nothing in main.py sets a status_code that surfaces as HTTP 200 with an error body. `os.environ.get(...)`, skipping the filter when unset, would degrade to current behaviour instead. Compose always sets it, so this only matters under image/compose skew.

On the DNS theory from my earlier comment — your probe disproved it.

entomb:
Thank you very much for taking a look and for pointing that out. I’ve addressed the compatibility case in ce54d838.

When `COMPOSE_PROJECT_NAME` is available, container discovery continues to use the project label filter. If it is absent, the `filters` argument is now omitted entirely, preserving the previous unfiltered behavior instead of raising a `KeyError`.

Thanks again for the helpful review!

Issue #7398, opened 2026-08-04
Domain-wide footer silently skipped for multipart mail after Rspamd 4.x upgrade (+ CTE mismatch corrupts embedded images)


53c70r:
### Contribution guidelines

- [x] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### Checklist prior issue creation

- [x] I understand that failure to follow below instructions may cause this issue to be closed.
- [x] I understand that vague, incomplete or inaccurate information may cause this issue to be closed.
- [x] I understand that this form is intended solely for reporting software bugs and not for support-related inquiries.
- [x] I understand that all responses are voluntary and community-driven, and do not constitute commercial support.
- [x] I confirm that I have reviewed previous [issues](https://github.com/mailcow/mailcow-dockerized/issues) to ensure this matter has not already been addressed.
- [x] I confirm that my environment meets all [prerequisite requirements](https://docs.mailcow.email/getstarted/prerequisite-system/) as specified in the official documentation.

### Description

The "Domain-wide footer" feature is broken since the Rspamd `3.14.3` → `4.1.4` upgrade (2026-07-30, `feat/rspamd-4.1.4`, via an intermediate `4.1.0` step in the same update). Two independent, silent failures:

1. **The footer is not applied to any multipart e-mail** — only flat single-part `text/plain` or `text/html` messages (no MIME boundary at all) still receive it. Since virtually every modern mail client (Outlook, Thunderbird HTML mode, Apple Mail, webmailers) sends multipart messages, this disables the feature for the vast majority of outbound mail.
2. **When the footer is applied, embedded images in the footer HTML can be corrupted** for messages whose original `Content-Transfer-Encoding` was `7bit` (observed with Thunderbird composing plain HTML mail).

Both fail silently — no error, no log entry, no bounce, message just goes out unmodified (1) or with a broken image (2). This is made harder to notice because mailcow's default Rspamd console logging is set to `level = "silent"` (`data/conf/rspamd/override.d/logging.inc`).

### Bug 1 — footer silently skipped for multipart messages

**Root cause:** `lua_mime.add_text_footer()` in Rspamd core (`/usr/share/rspamd/lualib/lua_mime.lua`) has this early-exit guard:

```lua
local text_parts = task:get_text_parts()

if not (html_footer or text_footer) or not (text_parts and #text_parts > 0) then
  return false
end
```

[...]

PR #7404, opened 2026-08-05
[Rspamd] announce the encoding the footer rewrite actually used

smpaz7467:
## Description

This is bug 2 of #7398 only — the `Content-Transfer-Encoding` mismatch. Bug 1 of that issue is not addressed here; see the note at the end.

`lua_mime.add_text_footer()` does not always quoted-printable encode the body. In `do_append_footer()` the encoder is chosen per text part:

```lua
local cte = 'quoted-printable'
local encode_func = function(input) return rspamd_util.encode_qp(input, 80, ...) end
if part:get_cte() == '7bit' then
  cte = '7bit'
  encode_func = function(input) ... end   -- passthrough, no QP applied
end
...
state.new_cte = cte
```

and it reports what it settled on back to the caller as `new_cte` (documented at `lua_mime.lua:440`, set at `:413`).

The footer rewrite in `rspamd.local.lua` ignored that and wrote the literal `quoted-printable` in all three places it emits the header. For a message that arrives as `7bit` the body is therefore passed through unencoded while the header claims quoted-printable.

That matters because the footer template typically carries an embedded base64 image as one very long unbroken line. Declared as quoted-printable, that line has no soft breaks protecting it. A relay that wraps over-long lines inserts a hard break into it, and a receiving client — decoding as quoted-printable, and seeing a break not preceded by `=` — keeps it as a real line break, corrupting the base64 and breaking the image.

The three literals become `rewrite.new_cte`, falling back to `quoted-printable` when it is absent.

Worth noting the fallback is only defensive: `need_rewrite_ct` is set solely for single-part `text/*` messages (`lua_mime.lua:461-475`), which is exactly the branch where `new_cte` is populated, so the hardcoded literal was wrong precisely where it was reachable.

## Verification

Live mailcow rspamd (`4.1.4`, mailcow's own image), domain-wide footer configured through the API containing an embedded base64 image on one line. Messages fed with `rspamc -m -u sender@footer.test -F sender@footer.test`, so `MOO_FOOTER` runs. Before/after captured with the same script, swapping only this file and restarting rspamd between runs; the loaded file was grepped inside the container each time to confirm which version was live.

Consistency is measured as: if the message declares quoted-printable, is the body actually encoded that way?

**Before**

| incoming CTE | declared CTE out | longest body line | QP soft breaks | verdict |
|---|---|---|---|---|
| `7bit` | `quoted-printable` | 1262 | 0 | **mismatch — says QP, body is raw** |
| `quoted-printable` | `quoted-printable` | 80 | 15 | consistent |
| `8bit` | `quoted-printable` | 14 | 0 | consistent |

**After**

| incoming CTE | declared CTE out | longest body line | QP soft breaks | verdict |
|---|---|---|---|---|
| `7bit` | `7bit` | 1262 | 0 | consistent — no QP claimed |
| `quoted-printable` | `quoted-printable` | 80 | 15 | consistent |
| `8bit` | `quoted-printable` | 14 | 0 | consistent |

Only the `7bit` path changes; the quoted-printable and 8bit paths emit exactly what they did before.

The footer itself still applies unchanged across structures — flat `text/plain`, flat `text/html`, `multipart/alternative`, `multipart/related`, and `multipart/mixed > multipart/related` + attachment all still receive it, with sub-part encodings (`base64` attachments) untouched.

## Note on bug 1 of #7398

I could not reproduce bug 1 ("footer silently skipped for any multipart message") on this setup. With mailcow's rspamd 4.1.4 image the footer was applied to every structure I tried, including the two marked ❌ in the issue:

| structure | footer applied |
|---|---|
| flat `text/plain` | yes |
| flat `text/html` | yes |
| `multipart/alternative` | yes |
| `multipart/related` (html + inline image) | yes |
| `multipart/mixed` > `multipart/related` + attachment | yes |

`task:get_text_parts()` returning empty for multipart would also disable a large amount of unrelated rspamd functionality, which argues against it being generally true. That is only my synthetic reproduction against `rspamc` though — @53c70r saw it on a real deployment, so there is likely a condition I did not recreate (footer html/plain combination, `mbox_exclude`, or the milter path rather than `rspamc`). Flagging it so bug 1 is not assumed fixed by this PR; it is untouched here.

Refs #7398


FreddleSpl0it:
This PR is not based on staging branch. I will close this.

PR #7405, opened 2026-08-05
fix: less disk writes redis/unbound

ValdikSS:
## Contribution Guidelines

## What does this PR include?

* unbound: mount /tmp to tmpfs to avoid disk writes
* redis: flush to disk every 12 hours, avoid frequent writes

### Short Description

Reduces unnecessary writes to SSD/HDD.

###  Affected Containers

- redis
- unbound

## Did you run tests?

I've tested manually

### What did you tested?

I've tested that unbound healthcheck script still works, and that redis does not write every 5 minutes and has applied the configuration, checked with redis cli.

Less disk writes, about 5-6 GB/day less.

### DIFF
--- a/data/conf/redis/redis-conf.sh
+++ b/data/conf/redis/redis-conf.sh
@@ -3,6 +3,7 @@
 cat <<EOF > /redis.conf
 requirepass $REDISPASS
 user quota_notify on nopass ~QW_* -@all +get +hget +ping
+save 43200 1000
 EOF
 
 if [ -n "$REDISMASTERPASS" ]; then
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,6 +5,8 @@ services:
       environment:
         - TZ=${TZ}
         - SKIP_UNBOUND_HEALTHCHECK=${SKIP_UNBOUND_HEALTHCHECK:-n}
+      tmpfs:
+        - /tmp
       volumes:
         - ./data/hooks/unbound:/hooks:Z
         - ./data/conf/unbound/unbound.conf:/etc/unbound/unbound.conf:ro,Z


milkmaker:
Thanks for contributing!

patschi:
For unbound, can you please provide more information on what this change brings? Which disk writes exactly? Do you know how much? As this can increase memory consumption noticeably.

For redis, I think 12 hours is way too much. In redis we also have, for example, the DKIM keys saved and we shouldn't loose them when only saving every 12 hours or every 1000 OPS...

patschi:
After additional research, I think we won't merge the change to 12 hours. As per redis documentation:
>You can configure Redis to have it save the dataset every N seconds if there are at least M changes in the dataset
https://redis.io/docs/latest/operate/oss_and_stack/management/persistence/#snapshotting

So should any crash occur, the user can loose up to 12 hours (!!) of ALL changed DKIM keys, BAYES training, logs, etc and ONLY when there were at least 1000 write operations in that time. If there were, let's say, 990 writes within 12 hours - in my understanding - redis will only save during the next cycle. 

Which seems to be rather drastic, and not worth the few GB writes it would do within that timespan.

Altough, I'd agree every 5 minutes is probably a bit too much too.

ValdikSS:
>For unbound, can you please provide more information on what this change brings? Which disk writes exactly?

This is for unbound container, not for unbound software itself.

A /healthcheck.sh script writes its status to /tmp, this patch makes these writes go to tmpfs.
It's just two byte write, but for a disk write this means at least 4KiB of data write + atime+mtime update + metadata update + for CoW FS (btrfs) this means changing the pointers, etc, at least 50+ KB write.

It runs every 30 seconds.

>For redis, I think 12 hours is way too much. In redis we also have, for example, the DKIM keys

What's the redis key name for these?
My redis database consists of 20864 (`RS_*`) + 205 (`RR:*`) + 147 (`rn_*`) + 1412 (`fuzzy_*`) rspamd keys, and there's about 100 other keys only (and some strange binary keys).

I just found forwarded hosts which seem to be present only in redis and not in mysql (I haven't found it in mysqldump), but otherwise I though everything is stored in mysql and redis is used for cache mainly?
I've asked [Too frequent disk writes, does Redis need persistance?](https://community.mailcow.email/d/6131-too-frequent-disk-writes-does-redis-need-persistance) on forum, have not received the answer.

Do you have a list of keys which are stored only in redis and needs to be persistent?

>Which seems to be rather drastic, and not worth the few GB writes it would do within that timespan.

Redis cache is one of the main write offender on my system.
I've installed mailcow on a dedicated machine with fresh SSD in 2021. No software besides mailcow, and mailcow is used only by me, single user, very idle server.
Yet my SSD life right now is at 40% after 5 years with almost 60 TB written. Every NAND cell is rewritten 1000+ times according to SMART data.
I've checked `docker compose stats` of the current container state and redis had many tens of GBs written.
After my patch, it has written 338 MB only during 14 days since last mailcow restart.

Unlike conventional databases, redis saves its data by just dumping all the current state into the file, regardless of its current contents. The write is always of the size of the database.

Screenshot_20260819_145921-fs8

ValdikSS:
The commit messages have comments if you haven't checked them.

PR #7428, opened 2026-08-18
[Web] Add ACL + global switch to disable external alias goto

FreddleSpl0it:
## Contribution Guidelines

### Short Description

Prevents aliases from targeting external (non local) domains, which commonly breaks SPF.

There are two levels of control:

* Global switch `$ALIAS_DISABLE_EXTERNAL_DOMAINS` in `vars.inc.php` (default `false`). When enabled it applies to all roles including admins and overrides the domain admin ACL.
* New domain admin ACL `alias_external_goto` (default allowed) that lets a super admin restrict individual domain admins. It is rendered automatically in the domain admin edit UI through the existing `da_acl` system.

A domain counts as local only if it exists as an active `domain` or `alias_domain`. The check runs in both alias `add` and `edit` paths, and goto domains are normalised via IDN before the comparison.

### Affected Containers

* php-fpm-mailcow

### DIFF
--- a/data/web/inc/functions.inc.php
+++ b/data/web/inc/functions.inc.php
@@ -633,6 +633,18 @@ function logger($_data = false) {
     return true;
   }
 }
+function is_local_mailcow_domain($domain) {
+  // True if domain is a locally managed, active primary or alias domain
+  global $pdo;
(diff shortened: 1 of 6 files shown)


marcojarjour:
That looks great, thank you very much for all your hard work!
I’m really looking forward to testing the changes as soon as they’re released, but it already looks exactly as I’d imagined.

Issue #5445, opened 2023-10-07
"Login Failed" with Generic OIDC SSO with Authentik


derfabianpeter:
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... I understand that not following the below instructions will result in immediate closure and/or deletion of my issue.
- [X] ... I have understood that this bug report is dedicated for bugs, and not for support-related inquiries.
- [X] ... I have understood that answers are voluntary and community-driven, and not commercial support.
- [X] ... I have verified that my issue has not been already answered in the past. I also checked previous [issues](https://github.com/mailcow/mailcow-dockerized/issues).

### Description

```plain text
I've set up mailcow on "Nightly: 45717ccd" since I wanted to test the SSO setup with Authentik as per [this blog post](https://mailcow.email/posts/2023/mailcow-idp/).

Configuration in mailcow:

- Authorization Endpoint: https://id.example.de/application/o/authorize
- Token Endpoint: https://id.example.de/application/o/token
- Userinfo Endpoint: https://id.example.de/application/o/userinfo
- Redirect URL: https://mail.example.de
- Attribute Mapping: "default" maps to the Default Template
- Client ID & Secret as shown in Authentik

Configuration in Authentik:

- Added a property mapping / scope named "maildow-dedicated" in Authentik
- Added the following expression as value for attribute mapping:

#####
return { "mailcow_template": "default" }
#####

- Added the scope to the OpenID provider 
- Validated that the claim "mailcow_template" with value "default" is also added to the ID token

Everything looks configured correctly, yet when logging in via SSO I get "Login failed/Anmeldung fehlgeschlagen".
```

### Logs:

```plain text
# mailcow UI logs:

["login_failed","Invalid response received from Authorization Server. Expected JSON."]


# nginx logs:

mailcowdockerized-php-fpm-mailcow-1    | 172.22.1.10 -  07/Oct/2023:15:18:03 +0200 "GET /index.php" 200
[...]

PR #7430, opened 2026-08-19
Fix/generic OIDC connection test

torrycrass:
Fix generic-oidc connection test: use OIDC discovery endpoint instead of OPTIONS to token URL

## Contribution Guidelines

### Short Description

Fixes the Test Connection button for Generic-OIDC identity providers, which always returns "Connection failed" regardless of whether the configuration is correct.

The existing code sends an HTTP `OPTIONS` request to the token endpoint and requires a `200` response. The OIDC specification does not require token endpoints to support OPTIONS — most providers (DEX, Authentik, Zitadel, etc.) return `404` or `405`, causing the test to always fail even when SSO login works correctly.

This PR replaces the OPTIONS request with a GET to the OIDC discovery endpoint (`/.well-known/openid-configuration`), which every spec-compliant OIDC provider must serve. The response is validated by checking for the presence of the `issuer` field, providing a meaningful correctness check rather than a meaningless method probe.

A minimal fallback alternative (accepting `405` in addition to `200`) is noted in the PR if the discovery URL derivation approach is not preferred.

Confirmed present in `2026-05c` and `2026-07a`. Related issue: #5445
 
**References:**
- Related issue: https://github.com/mailcow/mailcow-dockerized/issues/5445
- OIDC Discovery spec: https://openid.net/specs/openid-connect-discovery-1_0.html

### Affected Containers

- php-fpm-mailcow


FreddleSpl0it:
This PR is not based on staging branch. Please read https://github.com/mailcow/mailcow-dockerized/blob/master/CONTRIBUTING.md

